PR #8, opened 2022-09-23
develop

YaroslavKolbaiev:
[DEMO LINK](https://YaroslavKolbaiev.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,115 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext, useEffect, useMemo, useRef, useState,
+} from 'react';
+import classNames from 'classnames';
+import {
+  createTodo, getTodos, deleteTodo, updateTodo,
+} from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { Error } from './components/Error';
+
+enum FilterType {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,223 +1,178 @@
...
+  const [filter, setFilter] = useState<string>(FilterType.ALL);
+  const [changeAllTodos, setChangeAllTodos] = useState(true);
+  const [errorType, setError] = useState('');
+
+  const fetchTodo = async () => {
+    if (query.trim().length === 0) {
+      setError('Type something');
+      setTimeout(() => {
+        setError('');
+      }, 3000);
```
qoa11a, line 37: I see that you clear error after 3 seconds in 3 places. How about writing helper function to show some error and clear it after x seconds? Than you call it in these 3 places instead of repetitive code.

qoa11a: `const showError = (text: string, hideTimeout: number) => { implementation here... }`

src/App.tsx
```diff
@@ -1,223 +1,178 @@
...
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState<string>(FilterType.ALL);
+  const [changeAllTodos, setChangeAllTodos] = useState(true);
+  const [errorType, setError] = useState('');
```
qoa11a, line 30: [errorType, setError] => either [error, setError], or [errorType, setErrorType]
As I see you put error text here, so it doesn't sound right to call it "type"

src/utils/fetchClient.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
AlisaVlasova, line 1: remove this disable

AlisaVlasova (changes requested):

PR #26, opened 2022-09-29
Draft

VolodymyrChuyko:
[DEMO LINK](https://VolodymyrChuyko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,70 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import classNames from 'classnames';
+import {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Filter } from './components/Filter';
+import { TodoList } from './components/TodoList';
+
+import { FilterStatus } from './types/Filter';
+import { Todo } from './types/Todo';
+
+const getTodoById = (todos: Todo[], todoId: number) => {
+  return todos.find(({ id }) => id === todoId) || null;
+};
+
+const filterTodos = (todos: Todo[], filterStatus: FilterStatus) => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,11 +1,70 @@
...
+
+  const handleStatusChange = (todoId: number) => {
+    const copyTodos = [...visibleTodos];
+    const changedTodo = getTodoById(copyTodos, todoId);
+
+    if (changedTodo) {
+      changedTodo.completed = !changedTodo.completed;
+      setVisibleTodos(copyTodos);
+    }
+  };
```
NataliyaAn, line 47: Todo changes should be in the next task. 
I also added and changed todo manually without sending it to the server to check the task. But this method would be necessary to remove for review.

src/App.tsx
```diff
@@ -38,178 +99,43 @@ export const App: React.FC = () => {
...
+        )}
       >
         <button
           data-cy="HideErrorButton"
           type="button"
           className="delete"
+          onClick={handleCloseError}
         />
 
         Unable to add a todo
```
NataliyaAn, line 141: In this case, the user must see 'Unable to GET todos from server'. Not all three messages together

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        data-cy="FilterLinkAll"
+        href="#/"
+        className={classNames(
+          'filter__link',
+          { selected: filterStatus === 'all' },
+        )}
+        onClick={() => onFilter('all')}
```
NataliyaAn, line 19: Will be better 
```
onClick={() => (
      filterStatus !== 'all'
        && onFilter('all')
   )}
```
In this case `onFilter` won't work if `filterStatus` didn't change

NataliyaAn (approved):
Nice job! 
You could made some fixes

AndriiRut (approved):
good job! 🚀

PR #30, opened 2022-09-29
no additional components and problems with error output

DimaKobzar7:
[DEMO LINK](https://DimaKobzar7.github.io/react_todo-app-loading-todos/)

It's not over yet. There is no division into components, it is not clear what is wrong with the error message and a bunch of constant re-renders. And MANY comments

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,288 @@
+// надо прать функцию что ищет по айди и выводить объекты эти как тудушки
+// сами тудушки сделать компонентом отдельным как и сообщение про ошибку
+// сделать механизм фильтрации ол актив и комплит по тудушкам
+// файли authcontext fetchclient и auto form пока не трогать
+// ошибка должна показыватся при ломаном урл и на крестик убираться или через 3 сек исчезать
+// или исчезать если мы выбираем новую тудушку а на крест не нажали а 3 сек не прошло
+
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
+      setError(false);
+    }, 3000);
+  }
+
+  const closeError = (boolean: boolean) => {
+    setError(boolean);
+  };
+
+  useEffect(() => {
+    getTodos(5).then(response => {
```
AndriiRut, line 34: I remember you rewrote this. but don't forget to not hardcode users ID )

src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const user = useContext(AuthContext);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortType, setSortType] = useState<string>(SortTypes.All);
+  const [error, setError] = useState(false);
   const newTodoField = useRef<HTMLInputElement>(null);
 
+  if (!error) {
```
AndriiRut, line 23: if this condition is for error, it must be in the error component.

AndriiRut (approved):
Good job! 🚀

PR #32, opened 2022-09-30
loading and sort todo

artem-kodesnikov:
[DEMO LINK](https://artem-kodesnikov.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,23 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotification } from './components/Error/ErrorNotification';
+import { TodoList } from './components/TodoList/TodoList';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoCount } from './components/TodoCount/TodoCount';
+import { TodoFilter } from './components/Filter/TodoFilter';
+// import { TodoFooter } from './components/Footer/TodoFooter';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortBy, setSortBy] = useState('all');
+  const [errorMessage, setErrorMessage] = useState('');
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
@@ -13,6 +25,29 @@ export const App: React.FC = () => {
       newTodoField.current.focus();
     }
   }, []);
+
+  useEffect(() => {
+    getTodos(4)
+      .then(todosFromServer => setTodos(todosFromServer))
+      .catch(() => setErrorMessage('Unable to load data'));
+  }, []);
+
+  const activeTodo = todos.filter((todo) => {
+    return !todo.completed;
+  });
+
+  const visibleTodo = todos.filter((todo) => {
+    switch (sortBy) {
+      case 'Active':
+        return !todo.completed;
+
+      case 'Completed':
+        return todo.completed;
+
+      default:
+        return todo;
+    }
+  });
 
   return (
     <div className="todoapp">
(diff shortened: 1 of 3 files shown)


vladgusakovskiy (approved):
Cool 👍

AndriiRut (changes requested):

src/App.tsx
```diff
@@ -1,26 +1,58 @@
...
+import { Todo } from './types/Todo';
+import { TodoCount } from './components/TodoCount/TodoCount';
+import { TodoFilter } from './components/Filter/TodoFilter';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortBy, setSortBy] = useState('all');
```
dellachaise, line 16: Better crate enum for filter options. When you would write test tasks, using `enum` in simular case would be a ➕

src/App.tsx
```diff
@@ -1,26 +1,58 @@
...
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
   }, []);
 
+  useEffect(() => {
+    getTodos(user?.id || 0)
+      .then(todosFromServer => setTodos(todosFromServer))
```
dellachaise, line 27: Let’s use `async await`. It gives you a simpler way to work with asynchronous promise-based code

src/App.tsx
```diff
@@ -1,26 +1,58 @@
...
+      .catch(() => setErrorMessage('Unable to load data'));
+  }, [user]);
+
+  const activeTodo = todos.filter((todo) => {
+    return !todo.completed;
+  });
```
dellachaise, line 33: useMemo

src/components/Error/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+type Props = {
+  errorMessage: string
+};
+
+export const ErrorNotification: React.FC<Props> = ({ errorMessage }) => {
+  const [isError, setIsError] = useState(false);
+
+  setTimeout(() => {
+    setIsError(true);
+  }, 3000);
```
dellachaise, line 13: you need to create setTimeout on the first render and clear timeout on unmount

src/components/Filter/TodoFilter.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+        <a
+          data-cy="FilterLinkActive"
+          href="#/active"
+          className={classNames('filter__link',
+            { 'filter__link selected': selectedTab === 'Active' })}
+          onClick={() => {
+            setSortBy('Active');
+            setSelectedTab('Active');
+          }}
```
dellachaise, line 35: Create a separate method instead of creating an inline event handler.

src/components/Filter/TodoFilter.tsx
```diff
@@ -0,0 +1,61 @@
...
+        </a>
+        <a
+          data-cy="FilterLinkCompleted"
+          href="#/completed"
+          className={classNames('filter__link',
+            { 'filter__link selected': selectedTab === 'Completed' })}
+          onClick={() => {
+            setSortBy('Completed');
+            setSelectedTab('Completed');
+          }}
```
dellachaise, line 47: Here the same.

src/components/TodoCount/TodoCount.tsx
```diff
@@ -0,0 +1,13 @@
+import { Todo } from '../../types/Todo';
```
dellachaise, line 1: I think no need to crate a separate component for this purpose, you could move it to `TodoFilter`.
What do you think about it?

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,41 @@
+import classNames from 'classnames';
```
dellachaise, line 1: This component name is not so good.
Reread [Task checklist point 3](https://github.com/mate-academy/react_todo-app-loading-todos/blob/master/checklist.md). Component names examples there more meaningful.

Be always attentive with tasks descriptions and checklists. There **no redundant** words. The same would be in test tasks. You need to read it attentively and write the code according to it. It is important.

dellachaise (approved):
Good job 👍 

Need to add changes to the next parts of the task according to the comments.

Be always attentive with tasks descriptions and checklists. There **no redundant** words. The same would be in test tasks. You need to read it attentively and write the code according to it. It is important.
And using `enum` in appropriate way(as in this task) would be always ➕  in your test tasks.

PR #34, opened 2022-09-30
Develop

ibondrk:
[DEMO LINK](https://ibondrk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,87 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import {
+  FC,
+  useContext,
+  useState,
+  useEffect,
+  useReducer,
+  useMemo,
+  useRef,
+} from 'react';
+// import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
 
-export const App: React.FC = () => {
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { SortType } from './types/Filter';
+
+function filtTodos(
+  todos: Todo[],
+  sortType: SortType,
+) {
+  const visibleTodos = [...todos];
+
+  switch (sortType) {
+    case SortType.Active:
+      return visibleTodos.filter(todo => !todo.completed);
+
+    case SortType.Completed:
+      return visibleTodos.filter(todo => todo.completed);
+    default:
+      return visibleTodos;
+  }
+}
+
+const reducer = (count: number, action: string) => {
+  switch (action) {
+    case 'increase':
+      return count + 1;
+    case 'decrease':
+      return count - 1;
+    case 'clear':
+      return 0;
+    default:
+      return count;
+  }
+};
+
+export const App: FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [sortType, setSortType] = useState<SortType>(SortType.All);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completeItem, dispatch] = useReducer(reducer, 0);
+
+  const increase = () => dispatch('increase');
+
+  const getUserFromServer = (userId: number) => {
+    getTodos(userId)
+      .then(userTodosFromServer => setTodos(userTodosFromServer))
+      .catch(() => setErrorMessage('Unable to update todos'));
+  };
+
+  useEffect(() => {
+    if (!user) {
+      return;
+    }
+
+    getUserFromServer(user.id);
+  }, [user]);
+
+  useEffect(() => {
+    todos.map(todo => {
+      if (!todo.completed) {
+        increase();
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+
+      {todos.map((todo) => (
```
Andrii-Stiebalov, line 13: use destructuring

Andrii-Stiebalov (changes requested):

Andrii-Stiebalov:
image

Andrii-Stiebalov (approved):
Good job!!!👏👏👏

AndriiRut (changes requested):
hide filters when you don't have any todos
image

src/App.tsx
```diff
@@ -1,19 +1,96 @@
...
+      return 0;
+    default:
+      return count;
+  }
+};
 
-export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const App: FC = () => {
   const user = useContext(AuthContext);
```
ArturErmolenko, line 51: you have auth user so you could take id

src/App.tsx
```diff
@@ -1,19 +1,96 @@
...
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [sortType, setSortType] = useState<SortType>(SortType.All);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completeItem, dispatch] = useReducer(reducer, 0);
+
+  const increase = () => dispatch('increase');
+
+  const getUserFromServer = (userId: number) => {
```
ArturErmolenko, line 60:
```suggestion
  const getUserTodosFromServer = (userId: number) => {
```

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+  errorMessage: string;
+};
+
+export const ErrorNotification: FC<Props> = ({ errorMessage }) => {
+  const [isClosed, setIsClosed] = useState(false);
+
+  useMemo(() => setTimeout(() => {
+    setIsClosed(true);
+  }, 3000), [isClosed]);
```
ArturErmolenko, line 14: you need to create setTimeout on the first render and clear timeout on unmount, use useEffect

src/App.tsx
```diff
@@ -1,19 +1,96 @@
...
+  switch (sortType) {
+    case SortType.Active:
+      return visibleTodos.filter(todo => !todo.completed);
+
+    case SortType.Completed:
+      return visibleTodos.filter(todo => todo.completed);
+    default:
+      return visibleTodos;
+  }
+}
```
ArturErmolenko, line 35: create helper folder , and import its function where you need

ArturErmolenko (changes requested):
Good job 👍 
just a small improvements

src/App.tsx
ibondrk: But i used this function only in one place and just changing filtType inside the code. Can u describe how to do it better? (Probably one example)

dellachaise: Inside `components` folder you could create folder `helpers` with file inside for example `filterTodos.helper.ts` and move this code there. And in App.tsx just import and use it.

In a real project we do this way to make components more readable and simple to understand.

ibondrk: @dellachaise thanks, for help i'll do this

src/types/Filter.ts
```diff
@@ -0,0 +1,5 @@
+export enum SortType {
```
dellachaise, line 1: 🔥

src/App.tsx
```diff
@@ -1,19 +1,96 @@
...
+      return count;
+  }
+};
 
-export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const App: FC = () => {
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [errorMessage, setErrorMessage] = useState<string>('');
```
dellachaise, line 53: No need to write styles for primitives.
```suggestion
  const [errorMessage, setErrorMessage] = useState('');
```

dellachaise (approved):
Great job 👍 

Pay attention to comments and add fixes according to it to other parts of a project.

Also use this knowledge in the future with test projects(advices and keep doing things that you did great 👍 )

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+import classNames from 'classnames';
+import { SortType } from '../../types/Filter';
+
+type Props = {
+  sortType: SortType;
+  completeItem: number;
+  onSortChange: (sortType: SortType) => void;
+};
+
+export const Footer: FC<Props> = ({
```
dellachaise, line 11: Better rename to `Filter`.

Reread [Task checklist point 3](https://github.com/mate-academy/react_todo-app-loading-todos/blob/master/checklist.md). Component names examples there more meaningful.

Be always attentive with tasks descriptions and checklists. There **no redundant** words. The same would be in test tasks. You need to read it attentively and write the code according to it. It is important to pass all employment stages and get the job 😉

PR #35, opened 2022-09-30
Develop

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,91 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState, useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { AuthContext } from './components/Auth/AuthContext';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
+
+function preperTodos(todos:Todo[], sortFilter:string, user: User) {
+  return todos.filter(todo => todo.userId === user.id)
+    .filter(todo => {
+      switch (sortFilter) {
+        case 'active':
+          return !todo.completed;
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,19 +1,64 @@
...
+  const [visibelTodos, setVisibelTodos] = useState<Todo[]>([...todos]);
+
+  const [isError, setError] = useState(false);
+  const [messageError, setMessageError] = useState('');
+
+
+  const deleteCompletedTodos = () => (
+    setTodos(state => (
+      state.filter(todo => !todo.completed) || []))
+  );
```
VolodymyrChuyko, line 26:
```suggestion
  const deleteCompletedTodos = () => (
    setTodos(todos => (
      todos.filter(todo => !todo.completed)))
  );
```
A variable name should describe what is stored in it.

src/components/Error.tsx
```diff
@@ -0,0 +1,43 @@
...
+        'has-text-weight-normal',
+        { hidden: !isError },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        aria-label="close message error"
+        type="button"
+        className="delete"
+        onClick={() => setError(false)}
```
VolodymyrChuyko, line 37:
```suggestion
        onClick={() => setError(false)}
```
you can make an onClich handler instead of using an inline function

VolodymyrChuyko (approved):
Great job! Just small issues

AndriiRut (changes requested):
don't load todos from server
image

try use console.log to user/result of getTodos/ state with todos

NataliyaAn:
> don't load todos from server image
> 
> try use console.log to user/result of getTodos/ state with todos

Data is loaded from the server if there are data on server. If I try get data for ID 1. I got data
image

AndriiRut (approved):
Good job! 🚀

PR #58, opened 2022-10-03
Full task solution

adrqq:
I implemented a solution of 3 parts in this file, so consider this a full task. 
Haven't checked yet, whether I missed something

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+
+export const Footer: React.FC = () => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="todosCounter">
+        4 items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          data-cy="FilterLinkAll"
+          href="#/"
+          className="filter__link selected"
+        >
+          All
+        </a>
+
+        <a
+          data-cy="FilterLinkActive"
+          href="#/active"
+          className="filter__link"
+        >
+          Active
+        </a>
+        <a
+          data-cy="FilterLinkCompleted"
+          href="#/completed"
+          className="filter__link"
+        >
+          Completed
+        </a>
+      </nav>
+
+      <button
+        data-cy="ClearCompletedButton"
+        type="button"
+        className="todoapp__clear-completed"
+      >
+        Clear completed
+      </button>
+    </footer>
+  );
+};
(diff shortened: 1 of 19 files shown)


src/components/Header/Header.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const handleNewTodoTitleChange
+      = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setNewTodoTitle(event.target.value);
+      };
+
+    const handleNewTodoTitleSubmit = (event: React.FormEvent) => {
+      event.preventDefault();
+
+      if (newTodoTitle === '') {
```
anna-kolmakova, line 42: You need to add .trim() check - to disable adding of empty title (spaces only). Now I can add title name with spaces only.

anna-kolmakova (changes requested):
Good job! Just need to fix some options 👍

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,251 @@
...
+          <button
+            type="button"
+            className="todo__remove"
+            data-cy="TodoDeleteButton"
+            onClick={handleDeleteTodo}
+          >
+            ×
+          </button>
+
+          <div data-cy="TodoLoader" className="modal overlay">
```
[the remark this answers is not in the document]
adrqq: Yes, but I optimized it so it will be instant because it looks much better

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  const handleFilterType = (typeOfFilter: string) => {
+    setFilterType(typeOfFilter);
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="todosCounter">
+        {`${todos.length} ${todos.length === 1 ? 'item' : 'items'} left`}
```
anna-kolmakova, line 38: Here you need calculate only active (not completed) todos

anna-kolmakova (approved):

adrqq:
For some reason, I can not deploy my changes it deploys but nothing changes

AndriiRut (changes requested):
you have to hide filters if user don't have any todos `todos.length === 0`

AndriiRut (approved):
Add loader on other action. such as adding, deleting and etc.
on the clear all button add attribute disable instead of conditional render of it.

https://user-images.githubusercontent.com/99652389/199781994-c4491372-9db7-4f91-bed2-3a1a43559222.mov

PR #81, opened 2022-10-19
solution

maryna-mazur:


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+      <ul>
+        {todos?.map(todo => (
+          <li key={todo.id}>
+            <TodoItem todo={todo} />
+          </li>
+        ))}
+      </ul>
+
+      {/* <div data-cy="Todo" className="todo">
```
TarasHoliuk, line 19: don't leave comments
if you need them for the next steps you can save them in another branch (it is also always available in the master branch).

src/App.tsx
```diff
@@ -1,16 +1,46 @@
...
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
-  const newTodoField = useRef<HTMLInputElement>(null);
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [errorText] = useState('');
```
TarasHoliuk, line 21: You don't need 2 states
Keep only `[error, setError]` (with string), and set some error message if loading todos is unsuccessful (when you run `getTodos`)
In `ErrorNotification` you can just check if `error` exists: `''` will give you false

TarasHoliuk (changes requested):
Well done! But some improvements are needed

nazarmatsevych (approved):

PR #85, opened 2022-10-31
first try

Vovahunter:
[DEMO LINK](https://vovahunter.github.io/react_todo-app-loading-todos/)


vlad-demchuk (changes requested):
the first try failed)
fix the demo link (404)

AlisaVlasova (changes requested):
I have one uncompleted todo on server but it wasn`t sow up(
image

src/App.tsx
```diff
@@ -1,223 +1,171 @@
...
+
+  useEffect(() => {
+    if (user) {
+      getTodos(user.id)
+        .then(result => setTodos(result))
+        .catch(() => {
+          setErrorMessage(ErrorMessage.CannotLoadTodos);
+        });
+    }
+  }, [todos]);
```
qoa11a, line 82: You have todos as dependecy of useEffect which set new todos. It results in recursion call.
image

qoa11a (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,171 @@
...
+      ...temporaryTodo,
+      title,
+    });
+
+    setIsAdding(true);
+
+    addTodo(user.id, title)
+      .catch(() => setErrorMessage(ErrorMessage.UnableToAdd))
+      .finally(() => setIsAdding(false));
+  };
```
AlisaVlasova, line 109: тудушка не відображається, тільки temporaryTodo, потім зникає, з'являються лише після перезагрузки

AlisaVlasova: те саме з видаленням, список оновлюється лише при перезавантаженні сторінки

AlisaVlasova (changes requested):

src/App.tsx
```diff
@@ -1,223 +1,112 @@
...
+import { TodosFilter } from './components/TodoFilter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { TodoStatus } from './types/TodoStatus';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>(todos);
```
TarasHoliuk, line 24: no need to create a new state variable if you can get it from another state variable
just create variable a common `visibleTodos`

src/App.tsx
```diff
@@ -1,223 +1,112 @@
...
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
+    let todosCopy = [...todos];
```
TarasHoliuk, line 60: The `filter` method doesn't mutate original array, so no need to create copy before

src/App.tsx
```diff
@@ -1,223 +1,112 @@
...
   useEffect(() => {
-    // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+
+    loadTodos();
+  }, []);
+
+  useEffect(() => {
```
TarasHoliuk, line 59: useEffect is redundant here
if you'll create a common variable `visibleTodos` it will update on every state change (if `todos` or `todoStatus` changed)

if your goal was to avoid unnecessary defining of visible todos when `hasError` state changed, use `useMemo` instead

TarasHoliuk (changes requested):
GJ! Your app works as expected (load todos correctly) :+1:  But I've added a few suggestions related to code

PR #87, opened 2022-11-02
Implemented loading todos

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/utils/getVisibleTodos.ts
+++ b/src/utils/getVisibleTodos.ts
@@ -1,0 +1,13 @@
+import { Todo } from '../types/Todo';
+import { FilterBy } from '../types/FilterBy';
+
+export const getVisibletodos = (todos: Todo[], filterBy: FilterBy): Todo[] => {
+  switch (filterBy) {
+    case FilterBy.ACTIVE:
+      return todos.filter(todo => !todo.completed);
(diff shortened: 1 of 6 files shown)


src/utils/getVisibleTodos.ts
```diff
@@ -0,0 +1,13 @@
...
+import { FilterBy } from '../types/FilterBy';
+
+export const getVisibletodos = (todos: Todo[], filterBy: FilterBy): Todo[] => {
+  switch (filterBy) {
+    case FilterBy.ACTIVE:
+      return todos.filter(todo => !todo.completed);
+    case FilterBy.COMPLETED:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
```
AntonVaida, line 11: Maybe you can use switch-case inside filter()

AntonVaida (approved):
GJ🔥

src/utils/getVisibleTodos.ts
anastasiia-nahaiska: Good idea, but in this case, I should save a boolean in filterType. I need strings because I use its when creating the filter

Denchepornyuk (approved):

PR #94, opened 2022-11-04
React Todo App Load Todos

Vladosik1337:
[DEMO LINK](https://Vladosik1337.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, {
+  useCallback,
   useContext,
   useEffect,
   useRef,
@@ -17,12 +18,13 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
   const [errorStatus, setErrorStatus] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
-  const getTodosFromApi = async () => {
+  const getTodosFromApi = useCallback(async () => {
     if (user) {
       try {
         const todosApi = await getTodos(user?.id);
@@ -36,7 +38,7 @@ export const App: React.FC = () => {
         }, 3000);
       }
     }
-  };
+  }, []);
 
   useEffect(() => {
     getTodosFromApi();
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -37,187 +88,83 @@ export const App: React.FC = () => {
...
-              Completed
-            </a>
-          </nav>
+        {todos.length > 0 && (
+          <>
+            <TodoList todos={visibleTodos} />
+
+            <footer className="todoapp__footer" data-cy="Footer">
+              <span className="todo-count" data-cy="todosCounter">
+                {`${todos.filter(todo => !todo.completed).length} items left`}
```
vmamchur, line 97: maybe it'd be better to create a separate variable for this filter

bugsworld38 (approved):
Everything else looks good to me.
Good job!

src/App.tsx
```diff
@@ -1,19 +1,70 @@
...
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { FilterBy } from './types/FilterBy';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [errorStatus, setErrorStatus] = useState(false);
```
Denchepornyuk, line 19: would be great to move logic from App, component
It isn't required but it's better to keep App as simple as possible

src/App.tsx
```diff
@@ -1,19 +1,70 @@
...
+  const visibleTodos = todos.filter(todo => {
+    switch (filterBy) {
+      case FilterBy.ACTIVE:
+        return !todo.completed;
+      case FilterBy.COMPLETED:
+        return todo.completed;
+      default:
+        return todo;
+    }
+  });
```
Denchepornyuk, line 58: It's better to keep visibleTodos in the state then a separate variable in the component

src/App.tsx
```diff
@@ -1,19 +1,70 @@
...
+        setTodos(todosApi);
+      } catch {
+        setErrorStatus(true);
+
+        setTimeout(() => {
+          setErrorStatus(false);
+        }, 3000);
+      }
+    }
+  };
```
Denchepornyuk, line 39: you should use `useCallback` and `useMemo` hooks

src/App.tsx
Denchepornyuk: agree, move it to a separate variable

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #99, opened 2022-11-04
Develop

grafit-off:
[DEMO LINK](https://grafit-off.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,105 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  ErrorNotification,
+} from './components/ErrorNotification/ErrorNotification';
+import { Footer } from './components/Footer/Footer';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+
+import { Todo } from './types/Todo';
+import { TodosFilter } from './types/TodosFilter';
+import { getTodos } from './api/todos';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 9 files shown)


nikitaovch99 (approved):
GJ

src/App.tsx
```diff
@@ -1,223 +1,110 @@
...
+          return completed;
+
+        default:
+          return true;
+      }
+    });
+
+    setStatusToFilter(status);
+    setFiltredTodos(filtredByStatusTodos);
+  }, [todos]);
```
OleksiiBlezniuk, line 51: State with filtered todos is not required here, we could save result of filtering to constant using useMemo to avoid unnecessary filtration. Let's do it this way

src/App.tsx
```diff
@@ -1,223 +1,110 @@
...
+
+  const uncompletedTodosLength = useMemo(() => {
+    return todos.filter((todo) => !todo.completed).length;
+  }, [todos]);
```
OleksiiBlezniuk, line 55: No need to use return keyword if you can return value after arrow

OleksiiBlezniuk, line 55:
```suggestion
  const uncompletedTodosLength = useMemo(() => todos
    .filter((todo) => !todo.completed).length, [todos]);
```

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+      aria-label="Close notification"
+      className="delete"
+      onClick={closeNotification}
+    />
+
+    Unable to add a todo
+    <br />
+    Unable to delete a todo
+    <br />
+    Unable to update a todo
```
OleksiiBlezniuk, line 32: It's not clear why I see all this errors if I just got an error while todos loading (and I don't see message about loading)

OleksiiBlezniuk (changes requested):
GJ, left few small comments to fix

src/components/ErrorNotification/ErrorNotification.tsx
grafit-off: @OleksiiBlezniuk I implement errors message in second task, because it is not clear to me how to imitate such errors,  in this task I have not such methods as add, delete, update. So this one it is only like a dummy

grafit-off: But I will transfer this one from second, so it's okay

src/App.tsx
grafit-off: Thx, it's really good point, and it's looks easier!

OleksiiBlezniuk (approved):

PR #104, opened 2022-11-05
started components for todo app was created

iPanda177:
[UNBELIEVABLE DEMO LINK](https://iPanda177.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,59 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+
+import { getTodos } from './api/todos';
+
+import { SortType } from './types/SortType';
+import { Todo } from './types/Todo';
+
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './components/ErrorMessage';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,17 +1,59 @@
...
+  const filteredTodo = [...todos].filter(todo => {
+    switch (sortType) {
+      case SortType.Active:
+        return !todo.completed;
+      case SortType.Completed:
+        return todo.completed;
+      default:
+        return todo;
+    }
+  });
```
iam-datsenko, line 34: I think it's better to save todos to the state, and just update the state after filtering

src/App.tsx
```diff
@@ -37,187 +79,18 @@ export const App: React.FC = () => {
           </form>
         </header>
```
iam-datsenko, line 80: Move it to a separate function too

iam-datsenko (approved):
GJ!

src/App.tsx
iPanda177: this filtred array goes as a props to another component :)

Anna-Tkachenko (approved):

src/App.tsx
Anna-Tkachenko: will be better to cover it with useMemo

src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,43 @@
...
+        onClick={() => setError(false)}
+      >
+        <></>
+      </button>
+
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo
```
Anna-Tkachenko, line 40: it should be managed with error message variable

Anna-Tkachenko (approved):

PR #111, opened 2022-11-06
react_todo-app-loading-todos

xiimiik:
[DEMO LINK](https://xiimiik.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,100 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import {
+  ErrorNotification,
+} from './components/ErrorNotification/ErrorNotification';
+import { Footer } from './components/Footer/Footer';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { FilterType } from './types/FilterType';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.All);
+  const [hasError, setHasError] = useState(false);
+
+  const getTodosFromServer = async () => {
+    try {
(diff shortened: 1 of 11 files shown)


vladyslav-harkusha (approved):
You can add optimizations to functions and components using React.memo() and UseCallback()

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
+    try {
+      if (user) {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      }
+    } catch (error) {
+      setHasError(true);
+    }
+  };
```
OleksiiBlezniuk, line 38: Use useCallback hook here, or move function creation to useEffect (this function will be used only on first render, but it will be created on every render)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <a
+          data-cy="FilterLinkActive"
+          href="#/active"
+          className={classNames('filter__link', {
+            selected: filterType === FilterType.Active,
+          })}
+          onClick={() => setFilterType(FilterType.Active)}
+        >
+          Active
+        </a>
```
OleksiiBlezniuk, line 48: Or you can create array from FilterType enum and map it here 🙂

OleksiiBlezniuk: But it's not critical

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoListItem } from '../TodoListItem/TodoListItem';
+
+type Props = {
+  todos: Todo[]
+};
+
+export const TodoList: React.FC<Props> = React.memo(({ todos }) => {
+  return (
```
OleksiiBlezniuk, line 11: you can return value after arrow, without return keyword

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(({ completed, id, title }) => (
+        <div
+          data-cy="Todo"
+          className={classNames('todo', {
+            completed,
+          })}
+          key={id}
+        >
```
OleksiiBlezniuk, line 20: Looks like this div should be moved to TodoListItem component

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,37 @@
...
+        className="delete"
+        aria-label="Error button"
+        onClick={onClose}
+      />
+
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo
```
OleksiiBlezniuk, line 34: Add error message about error while todos loading
But also you can create an object with errorType and errorMessage properties for future

OleksiiBlezniuk (approved):

PR #117, opened 2022-11-07
solution

pelahiia:
[DEMO LINK](https://pelahiia.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,55 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useRef, useState,
+} from 'react';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,223 +1,85 @@
...
+      case FilterTypes.active:
+        return !todo.completed;
+
+      case FilterTypes.completed:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
```
qoa11a, line 35: It's just a recommendation, not a change request. I advice you to read about useMemo.
https://beta.reactjs.org/apis/react/useMemo

In a nutshell, when either props or state of React component changes it makes the component rerender. And new value for filteredTodos is calculated on every render as for now. It may be bad for slow devices as UX may be laggy. This hook helps to optimize such things.

qoa11a (approved):

PR #132, opened 2022-11-08
first solution

MAKSOti:
[DEMO LINK](https://MAKSOTi.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,57 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useMemo, useRef, useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/Errors/Errors';
+import { FilterType } from './types/FilterType';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState(FilterType.All);
+  const visibleTodos = useMemo<Todo[]>(() => {
+    if (todos) {
+      return [...todos].filter(todo => {
+        switch (filter) {
+          case FilterType.Active:
+            return !todo.completed;
+          case FilterType.Completed:
+            return todo.completed;
+          case FilterType.All:
+          default:
+            return true;
+        }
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,18 +1,57 @@
...
       newTodoField.current.focus();
     }
-  }, []);
+
+    if (user) {
+      getTodos(user.id)
+        .then(result => setTodos(result))
+        .catch(() => {
+          setErrorMessage('Cannot load todos');
+        });
```
anton-iskryk, line 48: I think it's better to use here try/catch construction and async/await syntax.

src/App.tsx
```diff
@@ -37,187 +76,24 @@ export const App: React.FC = () => {
...
-          >
-            Clear completed
-          </button>
-        </footer>
+        {todos.length > 0 && (
+          <>
+            <TodoList todos={visibleTodos} />
+            <Footer
+              filterBy={(filterType: FilterType) => setFilter(filterType)}
+              todosQuantity={todos.filter((todo) => !todo.completed).length}
```
anton-iskryk, line 84: I think it's better to make variables in the app component and here write the names of those variables. It's harder to read this code.

anton-iskryk (approved):
GJ! I wrote a few comments for you.

src/App.tsx
```diff
@@ -1,18 +1,57 @@
...
+import { ErrorNotification } from './components/Errors/Errors';
+import { FilterType } from './types/FilterType';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState(FilterType.All);
```
Denchepornyuk, line 18: add type for the state - `filter`

src/App.tsx
Denchepornyuk: agree, use async await and try catch

src/App.tsx
Denchepornyuk: agree, move it to the separate variables and don't forget to use useCallback and useMemo

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  todosQuantity: number;
+  selectedFilter: FilterType;
+}
+
+export const Footer: React.FC<Props> = ({
+  filterBy,
+  todosQuantity,
+  selectedFilter,
+}) => {
+  return (
```
Denchepornyuk, line 16: you don't need return here

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
```
Denchepornyuk, line 9: the same here, avoid return if you can

Denchepornyuk (changes requested):

PR #135, opened 2022-11-08
create first part of todo task

AndriyZhabynets:
[DEMO LINK](https://AndriyZhabynets.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,89 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext, useEffect, useMemo, useRef, useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,223 +1,89 @@
...
-            Clear completed
-          </button>
-        </footer>
+          <Footer
+            getFilterTodo={setFilterBy}
+            filteredTodos={filteredTodos}
+            selectedTab={filterBy}
+          />
+        </>
       </div>
```
glugovsky, line 79: TodoList and Footer should be visible only if there are todo items

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+      {todos.map(({ title, id, completed }) => {
+        return (
+          <div
+            key={id}
+            data-cy="Todo"
+            className={classnames(
+              'todo',
+              {
+                completed: completed === true,
```
glugovsky, line 23: Here you can avoid explicit equality check, `completed` is already a boolean, it is either true of false on its own

glugovsky (approved):

PR #136, opened 2022-11-09
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,79 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
+import { getTodos } from './api/todos';
+
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorMessages } from './components/ErrorMessages';
+import { TodosFilter } from './components/TodosFilter/TodosFilter';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodosList } from './components/TodosList/TodosList';
+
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,19 +1,80 @@
...
 
+  const loadUserTodos = useCallback(async () => {
+    if (user) {
+      try {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      } catch {
+        setHasError(true);
+      }
```
grafit-off, line 38: You should add some message that describes the error that can occur when we load todos

src/components/ErrorMessages/ErrorMessages.tsx
```diff
@@ -0,0 +1,45 @@
...
+        type="button"
+        className="delete"
+        onClick={onCloseError}
+      />
+
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo
```
grafit-off, line 42: You can remove these messages and add only message appropriate to the error. In the first task, you should implement appears error message in case loadTodos throw an Error

grafit-off (approved):
Looks ok, I left some comments below

src/App.tsx
```diff
@@ -1,19 +1,80 @@
...
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
+import { getTodos } from './api/todos';
+
 import { AuthContext } from './components/Auth/AuthContext';
```
Andr1yK, line 12:
```suggestion
import { getTodos } from './api/todos';
import { AuthContext } from './components/Auth/AuthContext';
```

PR #145, opened 2022-11-13
TodoApp part1

zhenija:
[DEMO LINK](https://zhenija.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,97 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useCallback,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Errors } from './components/Errors';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterMethod, setFilterMethod] = useState('all');
+
+  const getTodosFromsServer = useCallback(async () => {
+    try {
+      if (user) {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      }
+    } catch (Error) {
+      setHasError(true);
+      setErrorMessage('Loading error!');
+
+      setTimeout(() => {
+        setHasError(false);
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, []);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
+
+    getTodosFromsServer();
   }, []);
+
+  const filteredTodos = () => {
+    const toFilter = todos.filter(todo => {
+      switch (filterMethod) {
+        case 'active':
+          return !todo.completed;
+        case 'completed':
+          return todo.completed;
+        default:
+          return todos;
+      }
+    });
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
 import { AuthContext } from './components/Auth/AuthContext';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Errors } from './components/Errors';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Anastasiia-Svintsova, line 18: remove this

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterMethod, setFilterMethod] = useState('all');
+
+  const getTodosFromsServer = useCallback(async () => {
+    try {
+      if (user) {
+        const todosFromServer = await getTodos(user.id);
+
+        setTodos(todosFromServer);
+      }
```
Anastasiia-Svintsova, line 32: to avoid nesting you also can write smth like this
```suggestion
    if (!user) return;
    
    try {
      const todosFromServer = await getTodos(user.id);

      setTodos(todosFromServer);
```

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
+
+      setTimeout(() => {
+        setHasError(false);
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, []);
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
```
Anastasiia-Svintsova, line 45: remove comments all over the project

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
 
+  const filteredTodos = () => {
+    const toFilter = todos.filter(todo => {
+      switch (filterMethod) {
+        case 'active':
+          return !todo.completed;
+        case 'completed':
+          return todo.completed;
+        default:
+          return todos;
```
Anastasiia-Svintsova, line 61: create an enum for filters

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
-            <a
-              data-cy="FilterLinkAll"
-              href="#/"
-              className="filter__link selected"
-            >
-              All
-            </a>
+        <Header
+          newTodoField={newTodoField}
+          todos={filteredTodos()}
```
Anastasiia-Svintsova, line 75: it is better to write the function call in a separate constant and use the value here

src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
-            </a>
-            <a
-              data-cy="FilterLinkCompleted"
-              href="#/completed"
-              className="filter__link"
-            >
-              Completed
-            </a>
-          </nav>
+        {todos.length > 0 && (
```
Anastasiia-Svintsova, line 78:
```suggestion
        {!!todos.length && (
```

src/components/Errors.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const Errors:React.FC<Props> = ({
+  hasError,
+  setHasError,
+  errorMessage,
+}) => (
+  <div
+    data-cy="ErrorNotification"
+    className="notification is-danger is-light has-text-weight-normal"
+    hidden={!hasError}
```
Anastasiia-Svintsova, line 18: It's better to use render condition instead of hidden attribute. Just don't render this component at all if you don't have an error

Anastasiia-Svintsova (changes requested):
- I should see 2 here now, not 4
image

src/App.tsx
```diff
@@ -1,223 +1,108 @@
...
+        />
+
+        {!!todos.length && (
+          <>
+            <TodoList todos={filteredTodos} />
+
+            <Footer
+              filterMethod={filterMethod}
+              setFilterMethod={setFilterMethod}
+              todosLeft={todosLeft()}
```
polosanya, line 93: Try to save this value as you did with filtered todos

polosanya (changes requested):
Almost perfect! Please check my comments

src/components/TodoList.tsx
```diff
@@ -0,0 +1,18 @@
+// import classNames from 'classnames';
```
Anastasiia-Svintsova, line 1: this comment line is redundant

Anastasiia-Svintsova (approved):

PR #149, opened 2022-11-28
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,71 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  useCallback,
+} from 'react';
+import { getTodos } from './api/todos';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,223 +1,97 @@
...
+    } catch (error) {
+      setHasError(true);
+      setErrorMessage('Error!');
+
+      setTimeout(() => {
+        setHasError(false);
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, []);
```
qoa11a, line 43: Don't use useCallback without a reason. You should use useCallback only if you pass result to child component or into deps array.
https://beta.reactjs.org/apis/react/useCallback

qoa11a (changes requested):

Virass (changes requested):
I can't checked any tasks

image

anastasia-babko:
> I can't checked any tasks
> 
> image

But ability to toggle a todo status with checkbox is not required here. I've already added this toggling in react_todo-app-with-api task.

AlisaVlasova (approved):

PR #162, opened 2022-12-12
Develop

andriy-shymkiv:
[DEMO LINK](https://andriy-shymkiv.github.io/react_todo-app-loading-todos/)

I will handle the rest of Errors later, in the next two tasks


Illia-Skladnik (approved):
1. Your Toggle all button in header doesn't work. It should be inactive of you got as minimum 1 active todo (as I understand)
(If you manually toggle all todos to completed state, “Toggle all” should stay active.)
2. status-lable also doesn't work. There should be check symbol in completed todos.
3. you shouldn't use user as a props I guess. You arleady have it in useContext. 
4. I'm not sure it is necessary, but you can create separate component for your errors block.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const handleFormSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    if (!query.length) {
+      onErrorChange(true);
+    }
+
+    setTimeout(() => {
+      onErrorChange(false);
+    }, 3000);
```
mykola-koval, line 27: let's clear this timeout when we close the error

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+        {`${activeTodos.length} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        <a
+          data-cy="FilterLinkAll"
+          href="#/"
+          className={cn(
+            'filter__link',
+            { selected: clickedValue === 0 },
```
mykola-koval, line 70: this totally should be refactored

mykola-koval (approved):
you can implement fixes in the next task

PR #168, opened 2022-12-12
filter by status, decompose, rendered list were created

mitattt:
[DEMO LINK](https://mitattt.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,81 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Errors } from './components/Errors';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { Loader } from './components/Loader';
+import { TodoList } from './components/TodoList';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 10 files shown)


iryna-lelyk (approved):
Great solution 💪

src/App.tsx
```diff
@@ -1,223 +1,81 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState<boolean>(false);
+  const [currError, setCurrError] = useState<string>('');
+  const [isLoading, setIsLoading] = useState(true);
+  const [status, setStatus] = useState<Status>(Status.All);
+
+  const getTodosFromServer = useCallback(async () => {
+    try {
+      setIsLoading(true);
+      const todosFromServer = await getTodos(user?.id || 0);
```
Olexandr6, line 30:
```suggestion
      const todosFromServer = await getTodos(user?.id || 1);
```

src/App.tsx
```diff
@@ -1,223 +1,81 @@
...
+import { Header } from './components/Header';
+import { Loader } from './components/Loader';
+import { TodoList } from './components/TodoList';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
```
Olexandr6, line 20: Let's move this useRef with the corresponding `useEffect` to the `Header` component. Looks like you need this logic only in `Header` component, not in the `App.tsx`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+        case Status.Completed:
+          return todo.completed;
+
+        case Status.All:
+        default:
+          return todos;
+      }
+    });
+  }, [todos, status]);
```
Olexandr6, line 26: Move this logic directly to `useMemo`, you don't need `useCallback` here

G-YaSha (changes requested):
Almost done🔥🔥🔥🔥🔥
Now, let's work on improvements:

G-YaSha (approved):

PR #215, opened 2023-01-10
Develop

Oleh-Khashchevskyi:
[DEMO LINK](https://Oleh-Khashchevskyi.github.io/react_todo-app-loading-todos/)


src/utilities/Validation.ts
```diff
@@ -0,0 +1,17 @@
...
+  value: true,
+  message: 'This field is required',
+};
+
+export const isEmailValid = {
+  value: /\S+@\S+\.\S+/,
+  message: 'E-mail is not valid',
+};
+
+export const isMinValue = (value:string | undefined, min:number) => {
```
TarasHoliuk, line 11:
```suggestion
export const isMinValue = (value?: string, min:number) => {
```

src/utilities/HttpClient.ts
```diff
@@ -0,0 +1,29 @@
+import axios from 'axios';
+
+const instance = axios.create({
+  baseURL: 'https://mate.academy/students-api',
+});
+
+export const HttpClient = {
```
TarasHoliuk, line 7:
```suggestion
export const httpClient = {
```
use camelCase if it's not React component/type/class etc.

src/services/StorageService.ts
```diff
@@ -0,0 +1,17 @@
+import User from 'models/User';
+
+const StorageService = {
```
TarasHoliuk, line 3: camelCase

src/store/users/usersSlice.ts
```diff
@@ -0,0 +1,43 @@
...
+  name: 'users',
+  initialState,
+  reducers: {
+    setCurrentUser: (state:State, action:PayloadAction<User | null>) => {
+      state.currentUser = action.payload;
+    },
+  },
+  extraReducers: (builder) => {
+    builder
+      // Fetch Users
```
TarasHoliuk, line 24: I guess this comment is redundant - it is clear from the code

src/components/UserMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+          >
+            {currentUser.name}
+            <ArrowDropDownIcon />
+          </Box>
+        )}
+        onClick={openMenu}
+      />
+      <Menu>
+        <MenuItem onClick={handleClick}>
+          Sign Out
```
TarasHoliuk, line 42: :+1:

TarasHoliuk:
probably, you have to install smth like eslint-plugin-import to fix it (or just use relative paths):
image

TarasHoliuk (changes requested):
Nicely done :+1: 
You've done all 3 parts in the first task with additional futures :fire: 

I've left a few small comments

src/utilities/Validation.ts
Oleh-Khashchevskyi: cannot be set this way

unless so (value: string = '', min:number)

src/utilities/Validation.ts
```diff
@@ -0,0 +1,17 @@
...
+  value: true,
+  message: 'This field is required',
+};
+
+export const isEmailValid = {
+  value: /\S+@\S+\.\S+/,
+  message: 'E-mail is not valid',
+};
+
+export const isMinValue = (value:string, min:number) => {
```
tetiana-ivanchenko, line 11:
```suggestion
export const isMinValue = (value: string, min: number) => {
```

tetiana-ivanchenko (approved):

PR #218, opened 2023-01-15
Solution

vadiimvooo:
[DEMO LINK](https://vadiimvooo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,79 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React,
+{
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { ErrorMessages } from './components/ErrorMessages';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import TodoList from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/Filter';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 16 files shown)


daryna-zinchenko (approved):
looks good to me!

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,36 @@
+import { FilterStatus } from '../../types/Filter';
+import { Todo } from '../../types/Todo';
+import { Filter } from '../Filter';
+
+type Props = {
+  remainingTodos: Todo[],
```
ivan-kadykalo, line 6: You are using only length from this array, so no needs to drill whole list

src/App.tsx
```diff
@@ -1,223 +1,79 @@
...
+          setError('Can\'t load todos');
+        });
+    }
   }, []);
 
+  const remainingTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+  const filteredTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case 'Completed':
```
ivan-kadykalo, line 45: Use enum

ivan-kadykalo, line 45: If you will do a switch before filtering, the filter type will be checked only once, if you make it inside the filter method, the filter type will be checked on every element during the iteration.

```suggestion
  const filteredTodos = useMemo(() => {
    switch (filterType) {
      case FilterType.Active:
        return todos.filter(todo => (
          !todo.completed
        )
        
      ...........
        
      case FilterType.All:
      case default:
        return todos;
    }
  });
```

src/App.tsx
```diff
@@ -1,223 +1,79 @@
...
+    if (user) {
+      getTodos(user.id)
+        .then(setTodos)
+        .catch(() => {
+          setError('Can\'t load todos');
+        });
+    }
   }, []);
 
+  const remainingTodos = todos.filter(todo => !todo.completed);
```
ivan-kadykalo, line 41: Do you need all todos or just their quantity?

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #219, opened 2023-01-16
Initial commit

artem-hirzhev:
[DEMO LINK](https://artem-hirzhev.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,23 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  KeyboardEvent, useContext, useEffect, useRef, useState,
+} from 'react';
+import { getTodos, addTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Footer } from './components/Footer';
+import { TodosList } from './components/TodosList';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Header } from './components/Header';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [title, setTitle] = useState<string>('');
+  const [filter, setFilter] = useState(Filter.all);
+  const [onError, setOnError] = useState<string>('');
+  const [isHidden, setIsHidden] = useState(true);
+
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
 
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,9 +1,23 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  KeyboardEvent, useContext, useEffect, useRef, useState,
+} from 'react';
+import { getTodos, addTodo } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Footer } from './components/Footer';
+import { TodosList } from './components/TodosList';
```
SanyaBratashchuk, line 8: Create index file for imports

src/App.tsx
```diff
@@ -1,9 +1,23 @@
...
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Header } from './components/Header';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[] | []>([]);
```
SanyaBratashchuk, line 15:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/App.tsx
```diff
@@ -14,210 +28,86 @@ export const App: React.FC = () => {
...
+      title,
+      userId: user.id,
+      completed: false,
+    };
+
+    addTodo(todo)
+      .then(result => setTodos(
+        prevTodos => (prevTodos
+          ? [...prevTodos, result]
+          : [result]),
```
SanyaBratashchuk, line 53:
```suggestion
        prevTodos => (
          [...prevTodos, result]
          ),
```

src/App.tsx
```diff
@@ -14,210 +28,86 @@ export const App: React.FC = () => {
...
+      case Filter.completed:
+        return arr.filter(element => element.completed);
+
+      default:
+        return arr;
+    }
+  };
+
+  const visibleTodos = !todos ? [] : handleFilter(todos, filter);
+  const activeCount = !todos ? 0 : todos.filter(todo => !todo.completed).length;
```
SanyaBratashchuk, line 79: In which case todo can be falthy? If initial value is empty array which is truthy value

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setIsHidden(true)}
+      />
+      {onError === 'add' && 'Unable to add a todo'}
+      {/* <br /> */}
+      {onError === 'delete' && 'Unable to delete a todo'}
+      {/* <br /> */}
```
SanyaBratashchuk, line 34: Remove comments

SanyaBratashchuk (changes requested):

src/App.tsx
DarkMistyRoom: create an index file with all imports so that all imports in the app are from one this file

src/App.tsx
```diff
@@ -14,210 +28,84 @@ export const App: React.FC = () => {
     }
   }, []);
 
+  useEffect(() => {
+    if (user) {
+      getTodos(user.id)
+        .then(setTodos);
```
DarkMistyRoom, line 34: let's catch some error

src/types/Todo.ts
```diff
@@ -1,5 +1,5 @@
 export interface Todo {
-  id: number;
+  id?: number;
```
DarkMistyRoom, line 2: id should be required

DarkMistyRoom (changes requested):

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import classNames from 'classnames';
+
+type Props = {
+  isHidden: boolean;
+  setIsHidden: (arg: boolean) => void;
```
Volodymyr-Melnyk-Fron10der, line 6: Correct, but it's better to type `setState` function returned from `useState` as `Dispatch<SetStateAction<boolean>>`, so you will be able to use `setIsHidden `like `setIsHidden((previousValue) => someReturn)`

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onError: string;
```
Volodymyr-Melnyk-Fron10der, line 7: 1. `onError` sounds like a callback function name, I would recommend to name it as 'errorType'
2. If you know all posiible values of prop - create new type or enum of it. Example:

```
// for example in some your '...types.ts' file 
export type ErrorType = 'add' | 'load' | 'delete' | 'update'
```

and then 

```
import { ErrorType  } from '...types.ts';

type Props = {
  isHidden: boolean;
  setIsHidden: (arg: boolean) => void;
  onError: ErrorType | ''
}  

```

src/App.tsx
```diff
@@ -1,223 +1,117 @@
...
+  ErrorNotification,
+  Header,
+} from './components';
 
 export const App: React.FC = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState<string>('');
+  const [filter, setFilter] = useState(Filter.all);
+  const [onError, setOnError] = useState<string>('');
```
Volodymyr-Melnyk-Fron10der, line 20: use your ErrorType (see comment below)
`const [errorType, setErrorType] = useState<ErrorType | ''>('');`

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal',
+        { hidden: isHidden },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
```
Volodymyr-Melnyk-Fron10der, line 28: Not critical, but I would name it like `className="button-hide"` or `button-close`

Volodymyr-Melnyk-Fron10der (changes requested):
Great job! I see a few moments can be improved, make it please )

src/components/ErrorNotification.tsx
artem-hirzhev: Didn't know about this possibility (Dispatch<SetStateAction<boolean>>). It will be useful. Thanks!

SanyaBratashchuk (approved):

PR #225, opened 2023-01-20
react_todo-app-loading-todos

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,17 +1,72 @@
...
+  };
+
+  const matchTheError = (errorType: string) => {
+    if (!errorType) {
+      return '';
+    }
+
+    switch (errorType) {
+      case 'ServerFail':
+        return 'Unable to load data from the Server';
```
AlisaVlasova, line 42: use enum instead of switch case

src/App.tsx
```diff
@@ -1,17 +1,72 @@
...
+    setFilter('all');
+  };
+
+  const handleActiveTodos = () => {
+    setFilter('active');
+  };
+
+  const handleCompletedTodos = () => {
+    setFilter('completed');
+  };
```
AlisaVlasova, line 33: it will be enough to use one handler and enum for filter values

AlisaVlasova (changes requested):

nazarmatsevych (approved):
👍

PR #232, opened 2023-01-22
Solution

4ebupel:
[DEMO LINK](https://4ebupel.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,28 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  useCallback,
+  useMemo,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
 
+import { getTodos } from './api/todos';
+
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterType';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TododList';
+import { Footer } from './components/Footer/Footer';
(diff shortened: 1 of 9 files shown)


anastasiia-roiek (approved):
nice!

src/App.tsx
```diff
@@ -14,210 +34,76 @@ export const App: React.FC = () => {
...
+        case FilterType.Active:
+          return !todo.completed;
+
+        case FilterType.Completed:
+          return todo.completed;
+
+        default:
+          throw new Error('Invalid type');
+      }
+    });
```
ivan-kadykalo, line 71: If you will do a switch before filtering, the filter type will be checked only once, if you make it inside the filter method, the filter type will be checked on every element during the iteration.
```suggestion
  const filteredTodos = useMemo(() => {
    switch (filterType) {
      case FilterType.Active:
        return todos.filter(todo => (
          !todo.completed
        )
      ............
    }
  });
```

src/App.tsx
```diff
@@ -14,210 +34,76 @@ export const App: React.FC = () => {
...
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterType) {
+        case FilterType.All:
+          return todo;
```
ivan-kadykalo, line 60: AFAIS filter todo for FilterType.All just make a copy or todos. 
In addition to my comment about the switch case, If you need a copy make `[...todos]`, if you don't need just return original `todos`

ivan-kadykalo (changes requested):
There shouldn't be infinity loading
image

ivan-kadykalo (approved):
Good job 🔥

PR #238, opened 2023-01-23
Step1: Added components: TodoItem, TodoList, Geader and Footer

RedkoSerhii:
did not finished ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,4 +40,4 @@ Filter todos by status `All` / `Active` / `Completed`:
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_todo-app-loading-todos/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://RedkoSerhii.github.io/react_todo-app-loading-todos/) and add it to the PR description.
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,43 @@
+import React, { memo } from 'react';
+
+import { FilterType } from '../../types/FilterType';
+import { Filter } from '../Filter/Filter';
+
+type Props = {
+  activeTodosAmount: number;
+  hasCompletedTodos: boolean;
+  filterType: FilterType;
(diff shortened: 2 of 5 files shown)


Sergii-Osedchuk (approved):
Looks good. Just can`t see where do you use TodoLoader component.

src/App.tsx
```diff
@@ -1,9 +1,27 @@
...
+import { Todo } from './types/Todo';
+import { ErrorNotification } from
+  './components/ErrorNotification/ErrorNotification';
+import { StatusFilter } from './types/StatusFilter';
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [statusFilter, setStatusFilter] = useState(StatusFilter.All);
```
ivan-kadykalo, line 24: `FilterStatus` sounds better

src/App.tsx
```diff
@@ -14,210 +32,76 @@ export const App: React.FC = () => {
...
+  const filteredTodos = useMemo(() => {
+    return todos.filter((todo) => {
+      switch (statusFilter) {
+        case StatusFilter.All:
+          return todo;
+
+        case StatusFilter.Active:
+          return !todo.completed;
+
+        case StatusFilter.Completed:
```
ivan-kadykalo, line 63: If you will do a switch before filtering, the filter type will be checked only once, if you make it inside the filter method, the filter type will be checked on every element during the iteration.
```suggestion
  const filteredTodos = useMemo(() => {
    switch (filterType) {
      case FilterType.Active:
        return todos.filter(todo => (
          !todo.completed
        )
      ............
    }
```

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
+        data-cy="FilterLinkAll"
+        href="#/"
+        className={cn('filter__link', {
+          selected: statusFilter === StatusFilter.All,
+        })}
+        onClick={() => onChangeStatusFilter(StatusFilter.All)}
```
ivan-kadykalo, line 22:
```suggestion
        onClick={() => changeFilterStatus(FilterStatus.All)}
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/App.tsx
RedkoSerhii: ok, fixed it

src/App.tsx
RedkoSerhii: ok, fixed it

src/components/Filter/Filter.tsx
RedkoSerhii: ok, fixed it

ivan-kadykalo (approved):
Good job 🔥

PR #248, opened 2023-01-24
add new folders for components, replaced Errors block by new componen…

kreisberg:
…t in App

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+type FooterProps = {
+  //
+};
+
+export const Footer: React.FC<FooterProps> = () => {};
--- a/README.md
+++ b/README.md
@@ -40,4 +40,4 @@ Filter todos by status `All` / `Active` / `Completed`:
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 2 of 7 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { memo } from 'react';
+import cn from 'classnames';
+
+export enum Filters {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
```
DiDemkiv, line 7: Enum options should be in upper case

DiDemkiv (changes requested):
Couple small mistakes, but otherwise looks good to me 🚀

kreisberg:
> Couple small mistakes, but otherwise looks good to me 🚀

Thank's for all of them, so useful! Fix everything ✔️

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { memo, useEffect } from 'react';
+import cn from 'classnames';
+
+type ErrorProps = {
```
solaryasha, line 4:
```suggestion
type Props = {
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { memo } from 'react';
+import cn from 'classnames';
+
+export enum Filters {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
+
+type FooterProps = {
```
solaryasha, line 10:
```suggestion
type Props = {
```

solaryasha: It's just pattern, all these types called `Props`

solaryasha (approved):

PR #249, opened 2023-01-24
tasks uploaded & filter created

OlggaJul:
[DEMO LINK](https://olggajul.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,68 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import cn from 'classnames';
+import { getTodos } from './api/todos';
 import { AuthContext } from './components/Auth/AuthContext';
+import { Todo } from './types/Todo';
+
+enum FilterCondition {
+  ALL = 'all',
+  COMPLETED = 'completed',
+  ACTIVE = 'active',
+}
 
 export const App: React.FC = () => {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const user = useContext(AuthContext);
   const newTodoField = useRef<HTMLInputElement>(null);
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [errorText, setErrorText] = useState('');
+  const [filterCondition, setFilterCondition] = useState(FilterCondition.ALL);
+
+  if (isError) {
+    setTimeout(() => setIsError(false), 3000);
+  }
+
+  const getSelected = (allTodos: Todo[]): Todo[] => {
+    return allTodos.filter(item => {
+      switch (filterCondition) {
+        case FilterCondition.ALL:
+          return true;
+
+        case FilterCondition.COMPLETED:
+          return item.completed === true;
+
+        case FilterCondition.ACTIVE:
+          return item.completed === false;
+
+        default: return true;
+      }
+    });
+  };
 
   useEffect(() => {
     // focus the element with `ref={newTodoField}`
     if (newTodoField.current) {
       newTodoField.current.focus();
     }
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,222 +1,107 @@
...
 
-          <button
-            data-cy="ClearCompletedButton"
-            type="button"
-            className="todoapp__clear-completed"
-          >
-            Clear completed
-          </button>
-        </footer>
+        {(todosList.length > 0 || filterCondition !== FilterCondition.ALL) && (
```
official-artem, line 77: Create a constant with these conditions.

src/components/Main/Header.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const Header: React.FC<Props> = memo(({
+  todosList,
+  newTodoField,
+  setIsError,
+}) => {
+  return (
+    <header className="todoapp__header">
+      {todosList.length > 0 && (
```
official-artem, line 18:
```suggestion
      {todosList.length && (
```

src/components/Main/TodoList.tsx
```diff
@@ -0,0 +1,51 @@
...
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todosList: Todo[],
+};
+
+export const TodoList: React.FC<Props> = memo(({
+  todosList,
+}) => {
```
official-artem, line 11:
```suggestion
export const TodoList: React.FC<Props> = memo(
({ todosList }) => {
```

official-artem (approved):

src/App.tsx
```diff
@@ -1,222 +1,107 @@
...
-  }, []);
+
+    if (user) {
+      getTodos(user?.id)
+        .then(getSelected)
+        .then(setTodosList)
+        .catch(() => {
+          setTodosList([]);
+          setIsError(true);
+          setErrorText('Unable to upload a todo-list');
```
ivan-kadykalo, line 60: Good practice to use ENUM for error messages
```suggestion
          setErrorText(TodoErrors.UnableUpload);
```

src/App.tsx
ivan-kadykalo: > Create a constant with these conditions.

Good point. Fix it

src/App.tsx
```diff
@@ -1,222 +1,107 @@
...
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
+        {errorText}
+        {/* Unable to add a todo
+          <br />
+          Unable to delete a todo
+          <br />
+          Unable to update a todo */}
```
ivan-kadykalo, line 104: Create a enum for this errors and delete commented code

src/App.tsx
```diff
@@ -1,222 +1,107 @@
...
+
+        case FilterCondition.COMPLETED:
+          return item.completed === true;
+
+        case FilterCondition.ACTIVE:
+          return item.completed === false;
+
+        default: return true;
+      }
+    });
```
ivan-kadykalo, line 44: Better to check filter type before filtering. In this case you can avoid iteration for `FilterCondition.ALL`

```suggestion
    switch (filterCondition) {
        case FilterCondition.COMPLETED:
          return allTodos.filter(item => (
            item.completed
          )
          
        case FilterCondition.ACTIVE:
          return allTodos.filter(item => (
            !item.completed
          )
        
        case FilterCondition.ALL:
        default:
          return allTodos;
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/components/Main/Header.tsx
OlggaJul: otherwise I see 0 on page refresh😬

official-artem: > otherwise I see 0 on page refresh😬

Yeah, my bad, you can use !!todosList.length or leave your condition.

ivan-kadykalo (approved):
Good job 🔥

PR #256, opened 2023-01-25
solution

KrisPlanka:
[DEMO LINK](https://KrisPlanka.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,223 +1,88 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useContext, useEffect, useRef } from 'react';
+import React, {
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import { AuthContext } from './components/Auth/AuthContext';
+import { getTodos } from './api/todos';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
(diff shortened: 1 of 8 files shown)


olena-yanovska (changes requested):
Check please your filter, it doesn't work

src/App.tsx
```diff
@@ -1,223 +1,90 @@
...
-              className="todo__remove"
-              data-cy="TodoDeleteButton"
-            >
-              ×
-            </button>
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (status) {
+        case Filter.All:
+          return todo;
```
ivan-kadykalo, line 52: If you will do a switch before filtering, the filter type will be checked only once, if you make it inside the filter method, the filter type will be checked on every element during the iteration.

```suggestion
const filteredTodos = useMemo(() => {
    switch (filterType) {
      case FilterType.Active:
        return todos.filter(todo => (
          !todo.completed
        )
      ............
    }
  });
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/App.tsx
KrisPlanka: > If you will do a switch before filtering, the filter type will be checked only once, if you make it inside the filter method, the filter type will be checked on every element during the iteration.

Fixed, thank you!

ivan-kadykalo (approved):
Good job 🔥

PR #272, opened 2023-01-27
first implementation

Blenemy:


agniya-i (changes requested):
an active filter link should be hightlighted 


https://user-images.githubusercontent.com/32853034/215266028-5cdffb5b-c5de-43b7-a76f-0e1885275e9d.mov

qoa11a (approved):

PR #274, opened 2023-01-29
Solution

vitaliidox:
[DEMO LINK](https://vitaliidox.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,223 +1,76 @@
...
+    if (!user) {
+      return;
     }
+
+    getTodos(user.id)
+      .then(data => setTodos(data))
+      .catch(() => setIsError(Error.Update));
+  };
+
+  const filterdTodos = () => {
```
tetiana-ivanchenko, line 33:
```suggestion
  const filteredTodos = () => {
```

src/App.tsx
```diff
@@ -1,223 +1,76 @@
...
+        case Filter.All:
+        default:
+          return true;
+      }
+    });
+  };
+
+  const getFilterStatus = (arg: Filter) => {
+    setFilter(arg);
+  };
```
tetiana-ivanchenko, line 49: if your function name starts with "get" - it is expected to return something that we are trying to get, so use anothe name

src/App.tsx
```diff
@@ -1,223 +1,76 @@
...
 
-      <div
-        data-cy="ErrorNotification"
-        className="notification is-danger is-light has-text-weight-normal"
-      >
-        <button
-          data-cy="HideErrorButton"
-          type="button"
-          className="delete"
+      {isError !== null && (
```
tetiana-ivanchenko, line 68:
```suggestion
      {isError && (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { TodosItem } from '../TodosItem';
+
+type Props = {
+  todos: Todo[] | null,
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos && todos.map((todo: Todo) => (
```
tetiana-ivanchenko, line 11: if you haven't todos you shouldn't render TodoList component at all

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
+import { Todo } from '../../types/Todo';
+import { TodosItem } from '../TodosItem';
+
+type Props = {
+  todos: Todo[] | null,
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
```
tetiana-ivanchenko, line 10: use semantic tags for the list

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setIsError(null)}
+      />
+      Unable to add a todo
+      <br />
+      Unable to delete a todo
+      <br />
+      Unable to update a todo
```
tetiana-ivanchenko, line 37: you should render relevant error for each case, not all together

tetiana-ivanchenko (changes requested):

nazarmatsevych (changes requested):
Try to fix your demo or ask for help in fe_chat

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -15,7 +14,7 @@ export const ErrorNotification: React.FC<Props> = ({ error, setIsError }) => {
     }, 3000);
   }, []);
 
-  const getErrorsNotification = useMemo(() => {
+  const getErrorNotification = useMemo(() => {
```
TarasHoliuk, line 17: for such a case (when you have some logic in your `memo` function) better use `useCallback` hook, and call the resulting function at the place

and when you use `useMemo` function you should just return some value you want to memorize (and the variable name should be without a verb, i.e `errorNotification`)

This comment is just FYI (no need to re-request review)

TarasHoliuk (approved):

PR #284, opened 2023-02-05
first part implemented

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,96 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
 
-const USER_ID = 0;
+import { Header } from './components/Header';
+import { Main } from './components/Main';
+import { Footer } from './components/Footer';
+import { ErrorPanel } from './components/ErrorPanel';
+
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+
+const USER_ID = 6160;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [isEditing] = useState(false);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [selectedFilter, setSelectedFilter] = useState<
+    'all' | 'active' | 'completed'
+  >('all');
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  const onFilterAll = () => {
+    setSelectedFilter('all');
+    setFilteredTodos(todos);
+  };
+
+  const onFilterActive = () => {
+    setSelectedFilter('active');
+    setFilteredTodos(activeTodos);
+  };
+
+  const onFilterCompleted = () => {
+    setSelectedFilter('completed');
+    setFilteredTodos(completedTodos);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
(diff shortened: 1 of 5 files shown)


UmizDemud (approved):
I will approve, but next time use an approach where you are not storing multiple arrays in the memory. I think you need 
- `todos` : Because that is the base array
- `todosLength`, 'activeTodosLength', 'completedTodosLength (or even boolean if there is one or not.) for the Footer // Sorry for the names :D just explaining the idea
- 'visibletodos':  Because after the filter applies you'll still need one base to revert and one to print for the user

I think the rest is pretty good.

src/App.tsx
```diff
@@ -1,156 +1,96 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [isEditing] = useState(false);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [selectedFilter, setSelectedFilter] = useState<
+    'all' | 'active' | 'completed'
+  >('all');
```
naziamoff, line 22: create enum for the type. You should use enums when you have a bunch of string values - it is much easier to maintain this way

src/App.tsx
```diff
@@ -1,156 +1,96 @@
...
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
```
naziamoff, line 25: now you calculate it on each re-render. Let's improve our performance and think of a way of how we can do filtering only the way it is needed

src/components/ErrorPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const ErrorPanel: React.FC<Props> = ({
+  errorMessage,
+  showError,
+  clearError,
+}) => {
+  if (!showError) {
+    return null;
+  }
```
naziamoff, line 17: should you really return null instead of component here? Probably you can just render it conditionally from parent component

src/components/ErrorPanel.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    showError && (
```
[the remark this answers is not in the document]
naziamoff: yeah, seems redundant to me

src/components/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const Footer: React.FC<Props> = ({
+  activeTodos,
+  completedTodos,
+  onFilterAll,
+  onFilterActive,
+  onFilterCompleted,
+  selectedFilter,
+}) => {
+  return (
+    <footer className="todoapp__footer">
```
naziamoff, line 23: as you don't do any calculations/fetches here you can return right away. But still I'd recommend to create some variables here and check whether they can improve readability of the component itself

src/components/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+            'filter__link',
+            { selected: selectedFilter === 'completed' },
+          )}
+          onClick={onFilterCompleted}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
```
naziamoff, line 63:
```suggestion
```

src/components/Main.tsx
```diff
@@ -0,0 +1,89 @@
...
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  filteredTodos: Todo[],
+  isEditing: boolean,
+};
+
+export const Main: React.FC<Props> = ({ filteredTodos, isEditing }) => {
+  return (
```
naziamoff, line 11: why not just returning right away?
```suggestion
export const Main: React.FC<Props> = ({ filteredTodos, isEditing }) => (
```

naziamoff (approved):
not bad, still there are some things to fix. 
I approve this one, but you should fix those comments in your next ToDo App task. Let me know as soon as done - I'll take a look

PR #287, opened 2023-02-06
Solution ts_react_todo_app_load_todos

anton-martyniuk1:
[DEMO LINK](https://anton-martyniuk1.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -5,4 +5,14 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const saveTodo = (userId: number, data: any) => {
+  return client.post<Todo[]>(`/todos?userId=${userId}`, data);
+};
+
+export const updateTodo = (userId: number, data: any) => {
+  return client.patch<Todo[]>(`/todos?userId=${userId}`, data);
+};
+
+export const removeTodos = (userId: number) => {
+  return client.delete(`/todos?userId=${userId}`);
+};
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -1,0 +1,33 @@
+import React, { useState } from 'react';
+
+import { Todo } from '../types/Todo';
+import { Header } from './Header';
+import { Main } from './Main';
+import { Footer } from './Footer';
+import { Filter } from '../types/Filter';
+
+type Props = {
+  todos: Todo[] | null,
+};
+
+export const Content: React.FC<Props> = ({ todos }) => {
+  const [filter, setFilter] = useState<Filter>(Filter.all);
+
+  const isAllActive = todos?.every(todo => !todo.completed);
+
+  return (
+    <div className="todoapp__content">
+      <Header active={isAllActive} />
+
+      <Main visibleTodos={todos} />
+
+      {!todos?.length || (
+        <Footer
+          visibleTodos={todos}
+          filter={filter}
+          onChange={setFilter}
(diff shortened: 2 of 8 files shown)


qoa11a (changes requested):
1. Filtering should work.
2. Error messages should appear only when there's an error, and there should be proper error text while at the moment there's a text for 3 different errors.

src/types/Error.ts
```diff
@@ -0,0 +1,6 @@
+export enum Error {
+  succes = '',
```
polosanya, line 2:
```suggestion
  success= '',
```

src/api/todos.ts
```diff
@@ -5,4 +5,14 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const saveTodo = (userId: number, data: any) => {
```
polosanya, line 8: Try to avoid using any. Specify the data

src/components/Content.tsx
```diff
@@ -0,0 +1,53 @@
...
+        case (Filter.completed):
+          return todo.completed;
+
+        default:
+          return false;
+      }
+    }));
+  }, [visibleTodos, filter]);
+
+  const isAllActive = todos?.every(todo => !todo.completed);
```
polosanya, line 36: Would be great to memoize this variable

src/components/Footer.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const onActiveClick = () => {
+    onChange(Filter.active);
+  };
+
+  const onCompletedClick = () => {
+    onChange(Filter.completed);
+  };
+
+  const isSomeCompleted = visibleTodos?.some(todo => todo.completed);
```
polosanya, line 30: Same here

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments

src/components/Content.tsx
```diff
@@ -0,0 +1,56 @@
...
+          return !todo.completed;
+
+        case (Filter.completed):
+          return todo.completed;
+
+        default:
+          return false;
+      }
+    }));
+  }, [visibleTodos, filter]);
```
witflash, line 34: You don't need `visibleTodos` in dependencies. Just `filter`.

src/components/Footer.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
+
+import { Filter } from '../types/Filter';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  visibleTodos: Todo[] | null
```
witflash, line 8: I think you will never have `null` visibleTodos. You set an empty array by default, so here you will have an array always.
Check this case, please.

src/components/Footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const onActiveClick = () => {
+    onChange(Filter.active);
+  };
+
+  const onCompletedClick = () => {
+    onChange(Filter.completed);
+  };
+
+  const isSomeCompleted = useMemo(
+    () => visibleTodos?.some(todo => todo.completed),
```
witflash, line 31: And if you always have an array there is no need to use optional chaining.

witflash: This case is valid for Main component too.

witflash (changes requested):
Good work!
Check my comments, please. You can do some improvements.

PR #292, opened 2023-02-07
Develop

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,111 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodo, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { NewTodoForm } from './components/NewTodoForm';
+import { FormMain } from './components/FormMain';
+import { FormFooter } from './components/FormFooter';
+import { FormErrors } from './components/FormErrors';
 
-const USER_ID = 0;
+const USER_ID = 6146;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [error, setError] = useState(false);
+  const [errorMsg, setErrorMsg] = useState('Unable to load todos');
+  const [filter, setFilter] = useState('All');
+  const [completedTodo, setCompletedTodo] = useState(false);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setError(true);
+        setErrorMsg(errorMsg);
+
+        setTimeout(() => {
+          setError(false);
+        }, 3000);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (todo) {
+      getTodo(todo.id).then(setTodo);
+    }
+  }, [todo]);
+
+  const addTodo = (todoData: string) => {
+    const newTodo = {
+      id: Date.now(),
+      title: todoData,
+      completed: false,
+      userId: USER_ID,
+    };
+
+    setTodos([...todos, newTodo]);
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const selectedTodo = (id: number) => {
+    const ClickedTodo = todos.find(td => td.id === id);
+
+    if (ClickedTodo) {
+      if (completedTodo) {
+        setCompletedTodo(false);
+      }
+
+      if (!completedTodo) {
+        setCompletedTodo(true);
+      }
+
+      ClickedTodo.completed = completedTodo;
+    }
+  };
+
+  const visibleTodos = todos.filter((td) => {
+    if (filter === 'active') {
+      return !td.completed;
+    }
+
+    if (filter === 'completed') {
+      return td.completed;
+    }
+
+    return true;
+  });
+
+  const removeTodo = (todoId: number) => {
+    return todos.filter(td => td.id !== todoId);
+  };
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
-          <button type="button" className="todoapp__toggle-all active" />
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,156 +1,111 @@
...
+import { FormErrors } from './components/FormErrors';
 
-const USER_ID = 0;
+const USER_ID = 6146;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [error, setError] = useState(false);
+  const [errorMsg, setErrorMsg] = useState('Unable to load todos');
```
naziamoff, line 17: create a enum for the errorType

src/App.tsx
```diff
@@ -1,156 +1,111 @@
...
+
+    setTodos([...todos, newTodo]);
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const selectedTodo = (id: number) => {
+    const ClickedTodo = todos.find(td => td.id === id);
```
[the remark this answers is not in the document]
naziamoff: and probabaly `selected` will do better here?

src/App.tsx
```diff
@@ -1,156 +1,111 @@
...
-        </footer>
+        <NewTodoForm todos={visibleTodos} onAdd={addTodo} />
+        {todos.length > 0
+        && (
+          <FormMain
+            todos={visibleTodos}
+            selectedTodo={selectedTodo}
+            onRemove={removeTodo}
+          />
+        )}
```
naziamoff, line 100:
```suggestion
        {todos.length > 0 && (
          <FormMain
            todos={visibleTodos}
            selectedTodo={selectedTodo}
            onRemove={removeTodo}
          />
        )}
```

src/components/FormFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+          onClick={() => {
+            setFilter('completed');
+          }}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
+      {(todos.filter(todo => todo.completed)).length > 0
```
naziamoff, line 60: make a separate variables for the conditions like that. It is not easy to read this way

src/components/FormFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+            Clear completed
+          </button>
+        )
+        : (
+          <button
+            type="button"
+            className="todoapp__clear-completed__no-completed"
+          >
+            Clear completed
+          </button>
```
naziamoff, line 75: what's the difference between those two buttons? To me it seems the difference can be easily handled with `classnames` library

naziamoff: what's more, seems you just don't need to render it if there are no completed todos. So to me it looks like we can just do it like `areSomeTodosCompleted && (...yourCode)`

naziamoff (changes requested):
not bad already, but let's polish

mmahfuzi:
> not bad already, but let's polish

Hello, I corrected most of the parts as you suggested, some of them I did differently according to what Misha did today :)

src/App.tsx
```diff
@@ -1,156 +1,111 @@
...
+import { FormFooter } from './components/FormFooter';
+import { FormErrors } from './components/FormErrors';
 
-const USER_ID = 0;
+const USER_ID = 6146;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [error, setError] = useState(false);
```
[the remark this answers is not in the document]
naziamoff: not fixed

src/App.tsx
```diff
@@ -1,156 +1,111 @@
...
+  const [errorMsg, setErrorMsg] = useState('Unable to load todos');
+  const [filter, setFilter] = useState('All');
```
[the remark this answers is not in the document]
naziamoff: not fixed

src/App.tsx
```diff
@@ -1,156 +1,140 @@
...
+
+    setNewTodoTitle('');
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter((td) => {
+    switch (filter) {
```
naziamoff, line 64: use enum here

src/App.tsx
```diff
@@ -1,156 +1,140 @@
...
+        setTodos(todos.filter(td => td.id !== todoId));
+      })
+      .catch(() => setErrorMsg('Unable to delete todo'));
+  };
+
+  const todoUpdate = (todoToUpdate: Todo) => {
+    updateTodo(todoToUpdate)
+      .then(() => {
+        setTodos(
+          todos.map(td => {
```
naziamoff, line 86: whts td?

src/App.tsx
```diff
@@ -1,156 +1,140 @@
...
+            if (td.id === todoToUpdate.id) {
+              return todoToUpdate;
+            }
+
+            return td;
+          }),
+        );
+      })
+      .catch(() => setErrorMsg('Unable to update todo'));
```
naziamoff, line 95: enum for error message

src/components/FormFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const FormFooter: React.FC<Props> = ({
+  todos, filter, setFilter, onClearCompleted,
+}) => {
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {`${(todos.filter(todo => !todo.completed)).length} items left`}
```
naziamoff, line 20: must be moved to a separate variable, you should not do such calculations on render

naziamoff (changes requested):

src/App.tsx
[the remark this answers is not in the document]
mmahfuzi: I added <boolean>, is that what you mean?

src/App.tsx
mmahfuzi: Short for todo, since I have a state with name todo I used td for these. In fact I can't use todo I get error

meforoma (approved):
well, it's very decent @mmahfuzi 👏🏼

PR #293, opened 2023-02-07
First Part

grslckr:
[DEMO LINK](https://grslckr.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,75 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { ShowError } from './components/ShowError/ShowError';
+import { Todo } from './types/Todo';
+import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
+
+const USER_ID = 6156;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToRender, setTodosToRender] = useState<Todo[]>([]);
+  const [todoFilter, setTodoFilter] = useState('All');
+  const [error] = useState('');
+
+  const renderClearCompleted
+    = !!(todos.filter(todo => todo.completed).length);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(r => {
+        setTodos(r);
+        setTodosToRender(r);
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const handleFilterClick = (filter: string) => {
+    if (todoFilter !== filter) {
+      setTodoFilter(filter);
+
+      switch (filter) {
+        case 'All':
+        default:
+          setTodosToRender(todos);
+          break;
+        case 'Active':
+          setTodosToRender(todos.filter(todo => !todo.completed));
+          break;
+        case 'Completed':
+          setTodosToRender(todos.filter(todo => todo.completed));
+          break;
+      }
+    }
+  };
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
-          <button type="button" className="todoapp__toggle-all active" />
+        <Header userId={USER_ID} />
 
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
+        <TodoList todos={todosToRender} />
 
-        <section className="todoapp__main">
-          {/* This is a completed todo */}
-          <div className="todo completed">
-            <label className="todo__status-label">
-              <input
-                type="checkbox"
-                className="todo__status"
-                checked
-              />
-            </label>
-
-            <span className="todo__title">Completed Todo</span>
-
-            {/* Remove button appears only on hover */}
-            <button type="button" className="todo__remove">×</button>
-
-            {/* overlay will cover the todo while it is being updated */}
-            <div className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is not completed */}
-          <div className="todo">
-            <label className="todo__status-label">
-              <input
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span className="todo__title">Not Completed Todo</span>
-            <button type="button" className="todo__remove">×</button>
-
-            <div className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is being edited */}
-          <div className="todo">
-            <label className="todo__status-label">
-              <input
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            {/* This form is shown instead of the title and remove button */}
-            <form>
-              <input
-                type="text"
-                className="todo__title-field"
-                placeholder="Empty todo will be deleted"
-                value="Todo is being edited now"
-              />
-            </form>
-
-            <div className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is in loadind state */}
-          <div className="todo">
-            <label className="todo__status-label">
-              <input type="checkbox" className="todo__status" />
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
+        default:
+          setTodosToRender(todos);
+          break;
+        case 'Active':
+          setTodosToRender(todos.filter(todo => !todo.completed));
+          break;
+        case 'Completed':
+          setTodosToRender(todos.filter(todo => todo.completed));
+          break;
+      }
```
naziamoff, line 50: create a enum for filterBy, string values are not easily maintainable

src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
-              Completed
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        {todos.length !== 0 && (
```
naziamoff, line 63:
```suggestion
        {todos.length > 0 && (
```

src/components/ShowError/ShowError.tsx
```diff
@@ -0,0 +1,37 @@
...
+      setIsHidden(false);
+      setTimeout(() => {
+        setIsHidden(true);
+      }, 3000);
+    }
+  }, [error]);
+
+  return (
+    <div className={cn(
+      'notification', 'is-danger', 'is-light', 'has-text-weight-normal',
```
naziamoff, line 24:
```suggestion
      'notification is-danger is-light has-text-weight-normal',
```

src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
 import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header/Header';
 
-const USER_ID = 0;
+const USER_ID = 6156;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToRender, setTodosToRender] = useState<Todo[]>([]);
```
naziamoff, line 16:
```suggestion
  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```

src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
+  const [error] = useState('');
+
+  const renderClearCompleted
+    = !!(todos.filter(todo => todo.completed).length);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(r => {
+        setTodos(r);
+        setTodosToRender(r);
```
naziamoff, line 27: never use one-letter names for your variables, I don't understand what you're getting there because of the name`r` . Let's give descriptive names for our variables

src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
+const USER_ID = 6156;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToRender, setTodosToRender] = useState<Todo[]>([]);
+  const [todoFilter, setTodoFilter] = useState('All');
+  const [error] = useState('');
+
+  const renderClearCompleted
+    = !!(todos.filter(todo => todo.completed).length);
```
naziamoff, line 21: totally not clear what's happening here from the first glance. Let's simplify and give it a better name

naziamoff (changes requested):

src/App.tsx
```diff
@@ -1,156 +1,76 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+
```
naziamoff, line 2:
```suggestion
```

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+import { Header } from './components/Header/Header';
+import { FilterBy } from './types/Filter';
 
-const USER_ID = 0;
+const USER_ID = 6156;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [todoFilter, setTodoFilter] = useState('All');
```
naziamoff, line 18: you already have enum for filtering, why not just use it here?

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+  const [error] = useState('');
```
naziamoff, line 19: enum for error type

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+
+  const isAnyTodoCompleted
+    = Boolean(todos.filter(todo => todo.completed).length);
```
naziamoff, line 22: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+        setTodos(response);
+        setVisibleTodos(response);
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const handleFilterClick = (filter: FilterBy) => {
```
naziamoff, line 36: you don't need the word `handle` here. The names should be meaningful, I have to understand what's happening in the function body from its name.
```suggestion
  const changeTodosFilter = (filter: FilterBy) => {
```

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+    if (todoFilter !== filter) {
```
naziamoff, line 37: instead of writing the whole logic inside the `if` statement you can just do it like

```
if (todoFilter === filter) {
  return;
}

...function logic goes here...
```

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
+      setTodoFilter(filter);
+
+      switch (filter) {
+        case FilterBy.all:
+        default:
```
naziamoff, line 42: you can use switch inside the filter method

src/App.tsx
```diff
@@ -1,156 +1,76 @@
...
-        <button type="button" className="delete" />
-
-        {/* show only one message at a time */}
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <ShowError error={error} />
```
naziamoff, line 73: this is a component, but is named like a function. I'd say it is better to call it just `Error` or sth

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,37 @@
...
+type Props = {
+  userId: number,
+};
+
+export const Header: React.FC<Props> = ({ userId }) => {
+  const [todoTitle, setTodoTitle] = useState('');
+
+  const handlePostTodo = () => {
+    postTodo(userId, todoTitle);
+  };
```
naziamoff, line 15: no need in handle
```suggestion
  const createTodo = () => {
    postTodo(userId, todoTitle);
  };
```

src/components/ShowError/ShowError.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import cn from 'classnames';
+import { useEffect, useState } from 'react';
+
+type Props = {
+  error: string,
+};
+
+export const ShowError: React.FC<Props> = ({ error }) => {
+  const [isHidden, setIsHidden] = useState(true);
```
naziamoff, line 11: you don't need this state if you render the component conditionally from the parent one

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,67 @@
+import cn from 'classnames';
+import React from 'react';
+import { FilterBy } from '../../types/Filter';
+
+type Props = {
+  filter: string,
+  onFilterClick: (filter: FilterBy) => void,
```
naziamoff, line 7: https://mateacademyen.slack.com/archives/C0451UGKJLE/p1675787265139599

naziamoff (changes requested):
fix the remaining comments and you're good

src/api/todos.ts
```diff
@@ -5,4 +5,12 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const postTodo = (userId: number, title: string) => {
+  const todo = {
+    userId,
+    title,
+    id: new Date(),
```
meforoma, line 12: i find `Date.now()` more suitable for an id
compared to `new Date()`

Date now()

```suggestion
    id: Date.now(),
```

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+
+import cn from 'classnames';
+import { useEffect, useState } from 'react';
+
+type Props = {
+  error: string,
+};
+
+export const ShowError: React.FC<Props> = ({ error }) => {
```
meforoma, line 10: keep component Name & the file name synced :)
```suggestion
export const Error: React.FC<Props> = ({ error }) => {
```

meforoma (changes requested):
looks good @grslckr 
let's see few comment i left you.
and deploy your latest code - it seems not functional by the current DEMO LINK.

meforoma (approved):

PR #297, opened 2023-02-09
done

jackelashkar:
[DEMO LINK](https://jackelashkar.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,7 +65,7 @@ export const App: React.FC = () => {
         <Header
           onTodoTitle={todoTitle}
           onSetTodoTitle={setTodoTitile}
-          onHandleSubmit={() => handleSubmit}
+          onHandleSubmit={handleSubmit}
         />
 
         <TodoList
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -3,7 +3,7 @@
 type Props = {
   onTodoTitle: string,
   onSetTodoTitle: (value: string) => void,
-  onHandleSubmit: () => void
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,155 +1,91 @@
...
-                type="text"
-                className="todo__title-field"
-                placeholder="Empty todo will be deleted"
-                value="Todo is being edited now"
-              />
-            </form>
+        <Header
+          onTodoTitle={todoTitle}
+          onSetTodoTitle={setTodoTitile}
+          onHandleSubmit={() => handleSubmit}
```
meforoma, line 68: if you do it like this - you need to operate with args
otherwise `handleSubmit` will not execute
```suggestion
          onHandleSubmit={(event) => handleSubmit(event)}
          // or just
          onHandleSubmit={handleSubmit}
```

meforoma (changes requested):
- you can't submit a todo at the moment - try it yourself
make it work and we'll revisit )

src/App.tsx
```diff
@@ -65,7 +65,7 @@ export const App: React.FC = () => {
         <Header
           onTodoTitle={todoTitle}
           onSetTodoTitle={setTodoTitile}
-          onHandleSubmit={() => handleSubmit}
+          onHandleSubmit={handleSubmit}
```
meforoma, line 68: preposition 'on' just doesn't feel intuitively right
```suggestion
          todoTitle={todoTitle}
          setTodoTitle={setTodoTitile}
          handleSubmit={handleSubmit}
```

meforoma (approved):
✅

PR #301, opened 2023-02-13
todo-app 1/3

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,81 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useCallback, useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { Notification } from './components/Notification';
+import { TodoList } from './components/TodoList';
+import { ErrorMessages } from './types/ErrorMessages';
+import { FilterBy } from './types/FilterBy';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 6315;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterBy.ALL);
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(ErrorMessages.NONE);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(fetchTodos => {
+        setTodos(fetchTodos);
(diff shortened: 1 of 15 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const Notification: React.FC<Props> = ({
+  hasError,
+  errorMessage,
+  onClear,
+}) => {
+  useEffect(() => {
+    setInterval(onClear, 3000);
+  }, [errorMessage]);
```
GarryRocksman, line 17: so, where do you clear the interval?

GarryRocksman (approved):
GJ🔥

src/App.tsx
```diff
@@ -1,156 +1,81 @@
...
+import { Todo } from './types/Todo';
+
+const USER_ID = 6315;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterBy.ALL);
```
Anna-Tkachenko, line 16: let's use enum for this field instead of simple string

src/App.tsx
```diff
@@ -1,156 +1,81 @@
...
+  const [hasError, setHasError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(ErrorMessages.NONE);
```
Anna-Tkachenko, line 18: add type for this variable

src/App.tsx
```diff
@@ -1,156 +1,81 @@
...
+      .then(fetchTodos => {
+        setTodos(fetchTodos);
+      })
+      .catch(() => {
+        setHasError(true);
+        setErrorMessage(ErrorMessages.ONLOAD);
+      });
+  }, []);
+
+  const visibleTodos = todos.filter(todo => {
```
Anna-Tkachenko, line 31: all additional functionality that you used for your data will be better to move to separate utils function instead of writing this code directly in the component

src/components/Notification/Notification.tsx
Anna-Tkachenko: check @GarryRocksman comment
`setInterval` creates the process that will call your function each 3 seconds for now

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,36 @@
...
+    )}
+    >
+
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
+        type="button"
+        className="delete"
+        onClick={onClear}
+      />
+
```
Anna-Tkachenko, line 32: delete redundant empty line

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #302, opened 2023-02-13
Develop

sercansimsek:


src/components/Footer.tsx
```diff
@@ -0,0 +1,81 @@
...
+      </nav>
+
+      {completedTodos.length > 0 && (
+        <button
+          type="button"
+          className={cn(
+            'todoapp__clear-completed',
+            { 'todoapp__clear-completed__no': completedTodos.length > 0 },
+          )}
+          onClick={() => onRemoveCompleted()}
```
meforoma, line 73:
```suggestion
          onClick={onRemoveCompleted}
```

src/App.tsx
```diff
@@ -1,10 +1,114 @@
...
+      })
+      .catch(() => setErrorMessage(ErrorType.UpdatingError));
+  };
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const removeCompleted = () => {
+    completedTodos.forEach(todo => {
+      removeTodo(todo.id);
+    });
+  };
```
meforoma, line 86: try removing all completed items - it's not working as intended
removes a single todo at a time
make it to remove all completed

meforoma (changes requested):
good job 👍🏼 
few things to fix though.

meforoma (approved):
regarding `removeCompleted` - i think you can cover this in next iterations, so let's move on 😉

PR #303, opened 2023-02-13
solution

Avichez:
[DEMO LINK](https://Avichez.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,54 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useMemo, useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { FilterBy } from './types/FilterBy';
+import { warningTimer } from './utils/warningTimer';
+import { TodoInfo } from './components/TodoInfo';
+import { Footer } from './components/Footer/Footer';
+import { Header } from './components/Header';
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+const USER_ID = 6316;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [hasError, setHasError] = useState(false);
+  const hasActiveTodos = todos.some(todo => todo.completed === false);
+  const hasCompletedTodos = todos.some(todo => todo.completed === true);
+
+  useEffect(() => {
+    const onLoadGetTodos = async () => {
+      try {
+        const todosData = await getTodos(USER_ID);
+
+        setTodos(todosData);
+      } catch (error) {
+        setHasError(true);
+        warningTimer(setHasError, false, 3000);
+      }
+    };
+
+    onLoadGetTodos();
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    const filteredTodos = todos.filter(todo => {
+      switch (filterBy) {
+        case FilterBy.ACTIVE:
+          return !todo.completed;
+        case FilterBy.COMPLETED:
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+
+    return filteredTodos;
+  }, [todos, filterBy]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -14,143 +58,30 @@ export const App: React.FC = () => {
...
-            <span className="todo__title">Todo is being saved now</span>
-            <button type="button" className="todo__remove">×</button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+          {visibleTodos.map(todo => (
```
anna-inozemets, line 64: May be it is better to create a separate component, like TodoList, for this section

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+        setHasError(true);
+        warningTimer(setHasError, false, 3000);
+      }
+    };
+
+    onLoadGetTodos();
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    const filteredTodos = todos.filter(todo => {
```
anna-inozemets, line 38: I guess, it is possible to create a separate function for this part of code

anna-inozemets (approved):
Great solution🔥

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+import { Notification } from './components/Notification';
 
-const USER_ID = 0;
+const USER_ID = 6316;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [hasError, setHasError] = useState(false);
+  const hasActiveTodos = todos.filter(todo => !todo.completed);
```
alexandra-protyanova, line 19: if you call variable like that, use some, bad naming

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+  const hasCompletedTodos = todos.some(todo => todo.completed === true);
```
alexandra-protyanova, line 20: useMemo is highly appreciated

alexandra-protyanova, line 20: variable is boolean, no need to compare with true

src/App.tsx
alexandra-protyanova: let us do it

src/App.tsx
```diff
@@ -14,143 +58,30 @@ export const App: React.FC = () => {
...
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
+        <Header hasActiveTodos={hasActiveTodos.length} />
```
alexandra-protyanova, line 61: use some for this need

src/App.tsx
alexandra-protyanova: `TodoList` should be created

src/App.tsx
```diff
@@ -14,143 +58,30 @@ export const App: React.FC = () => {
...
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        {!!todos.length && (
+          <Footer
+            quantity={hasActiveTodos.length}
+            filterBy={filterBy}
+            setFilterBy={setFilterBy}
```
alexandra-protyanova, line 74: don't pass set function, create wrappers for them

src/App.tsx
```diff
@@ -14,143 +58,30 @@ export const App: React.FC = () => {
...
-        <button type="button" className="delete" />
-
-        {/* show only one message at a time */}
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <Notification
```
alexandra-protyanova, line 81: shouldn't be rendered if there is no error

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+  hasCompletedTodos: boolean,
+};
+
+export const Footer: React.FC<Props> = ({
+  quantity,
+  filterBy,
+  setFilterBy,
+  hasCompletedTodos,
+}) => {
+  return (
```
alexandra-protyanova, line 18: redundant return

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <nav className="filter">
+        <a
+          href="#/"
+          className={cn('filter__link', {
+            selected: filterBy === FilterBy.ALL,
+          })}
+          onClick={() => setFilterBy(FilterBy.ALL)}
+        >
+          All
+        </a>
```
alexandra-protyanova, line 34: you duplicate a lot of code, just map enum instead

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,34 @@
...
+  hasActiveTodos: number,
+};
+
+export const Header: React.FC<Props> = ({ hasActiveTodos }) => {
+  const [title, setTitle] = useState('');
+
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
alexandra-protyanova, line 14: fix linter, (first link in google)

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,32 @@
...
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setHasError(false)}
+      />
+
+      {/* show only one message at a time */}
+      Unable to add a todo
```
alexandra-protyanova, line 25: you should render error from enum here

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import cn from 'classnames';
+
+type Props = {
+  hasError: boolean,
+  setHasError: (value: boolean) => void,
+};
+
+export const Notification: React.FC<Props> = ({ hasError, setHasError }) => {
+  return (
```
alexandra-protyanova, line 10: remove return

alexandra-protyanova (changes requested):

src/App.tsx
Avichez: @alexandra-protyanova it's done by adding the corresponding class, otherwise, the animation doesn't work properly.

alexandra-protyanova (approved):

PR #304, opened 2023-02-13
Develop

KareemAchkar:
(https://KareemAchkar.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,76 @@
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  onSelected: string
+  onSetSelected: (value: string) => void
+  Todos: Todo[]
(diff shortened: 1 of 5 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </a>
+
+          <a
+            href="#/completed"
+            className={cn('filter__link', {
+              selected: onSelected === 'Completed',
+            })}
+            onClick={() => onSetSelected('Completed')}
+
+          >
```
meforoma, line 53: 'All', 'Active', 'Completed' should be enum-ed
do it and use this enum across your code

meforoma (approved):
good-good overall 👍🏼 
let's just involve an enum for filtering - you can apply this in the next task, but be sure to do it - otherwise reject :)

PR #311, opened 2023-02-17
Develop

svitjojo:
[DEMO LINK](https://svitjojo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Notification/Notification.tsx
+++ b/src/components/Notification/Notification.tsx
@@ -1,0 +1,36 @@
+import { useEffect, useState } from 'react';
+import cn from 'classnames';
+import { ErrorMessage } from '../../types/ErrorMessage';
+
+type Props = {
+  errorMessage: ErrorMessage | string,
+};
+
+export const Notification: React.FC<Props> = ({ errorMessage }) => {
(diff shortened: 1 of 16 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react';
+import cn from 'classnames';
+import { ErrorMessage } from '../../types/ErrorMessage';
+
+type Props = {
+  errorMessage: ErrorMessage | string,
+};
+
+export const Notification: React.FC<Props> = ({ errorMessage }) => {
+  const [isHidden, setIsHidden] = useState(false);
```
Avichez, line 10: You can use here state from App component `hasError, setHasError` and change its status to false after Timeout. But then you need to remove the condition ` {hasError && <Notification errorMessage={errorMessage} />}` in App component, otherwise, you won't see the animation of hiding the component.
Just as suggestion 🙂

Avichez (approved):
Great beginning of the solution 🔥

alexandra-protyanova (approved):
I'll leave comments in last part of the task

PR #336, opened 2023-02-20
Develop

dolnaro:
[DEMO LINK](https://dolnaro.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,8 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import classNames from 'classnames';
 import React, { useEffect, useState } from 'react';
-import { getTodos } from './api/todos';
+import { getTodos, createTodo } from './api/todos';
 import { Todo } from './types/Todo';
-import { createTodo } from './api/todos';
 
 type Status = 'all' | 'active' | 'completed';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,12 +1,72 @@
...
+      .then(newTodo => {
+        setTodos(todos.concat(newTodo));
+      });
+  }
+
+  let visibleTodos = todos;
+
+  if (status === 'active') {
+    visibleTodos = todos.filter(todo => !todo.completed);
+  }
```
naziamoff, line 66: let's use `enum + switch/case` for limited number of string values. String values are difficult to maintain in your code

src/App.tsx
```diff
@@ -1,12 +1,72 @@
...
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+type Status = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
```
naziamoff, line 11: create enum for errors

naziamoff (changes requested):
fix comments. Also, UI seems to be broken

PR #352, opened 2023-02-27
solution

miamarichka:
[DEMO LINK](https://miamarichka.github.io/react_todo-app-loading-todos/)

Next reviewer, please check this conversations before correct the footer appearance. Thank you
https://mate-parttime.slack.com/archives/C041JA5EJKV/p1677753192380139

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -100,7 +100,7 @@ export const App: React.FC = () => {
         {todos.length > 0
           && <Main todos={visibleTodos} addComplitedTodo={addComplitedTodo} />}
 
-        {todos && (
+        {todos.length && (
           <Footer
             filterTodos={filterTodosBySelectOptions}
             todosCount={todos.length}
--- a/src/components/errorNotification.tsx
+++ b/src/components/errorNotification.tsx
@@ -10,10 +10,6 @@ export const ErrorNotification: React.FC<Props> = ({
   onNotificationClose,
   isErrorOccuring,
 }) => {
-  const onCloseHandler = () => {
-    onNotificationClose();
-  };
-
   return (
     <div className={classnames('notification', 'is-danger',
       'is-light', 'has-text-weight-normal', { hidden: isErrorOccuring })}
@@ -21,7 +17,7 @@ export const ErrorNotification: React.FC<Props> = ({
       <button
         type="button"
         className="delete"
-        onClick={onCloseHandler}
+        onClick={onNotificationClose}
       />
       {/* show only one message at a time */}
       Unable to add a todo


qoa11a (changes requested):
image
image
please compare your solution to the example. It should look the same

src/App.tsx
```diff
@@ -1,10 +1,91 @@
...
+        setVisibleTodos(todos.filter((todo: Todo) => !todo.completed));
+        break;
+      case SelectOptions.COMPLETED:
+        setVisibleTodos(todos.filter((todo: Todo) => todo.completed));
+        break;
+      case SelectOptions.ALL:
+        setVisibleTodos(todos);
+        break;
+      default:
+        setVisibleTodos(todos);
```
Anastasiia-Svintsova, line 38: Combine these cases

src/App.tsx
```diff
@@ -1,10 +1,91 @@
...
+
+  useEffect(() => {
+    loadTodosFromServer();
+  }, []);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setOnError(false);
+    }, 3000);
+  }, []);
```
Anastasiia-Svintsova, line 69: Clear the timeout on the component unmount, please 
https://bobbyhadz.com/blog/react-cleartimeout

src/components/errorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+  onNotificationClose: () => void,
+  isErrorOccuring: boolean,
+};
+export const ErrorNotification: React.FC<Props> = ({
+  onNotificationClose,
+  isErrorOccuring,
+}) => {
+  const onCloseHandler = () => {
+    onNotificationClose();
+  };
```
Anastasiia-Svintsova, line 15: Why do you need to wrap this function into another function?

src/components/footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+          href="#/"
+          className={classNames(
+            'filter__link', {
+              selected: selectedFilter === SelectOptions.ALL,
+            },
+          )}
+          onClick={() => {
+            setSelectedFilter(SelectOptions.ALL);
+            filterTodos(SelectOptions.ALL);
+          }}
```
Anastasiia-Svintsova, line 37: Move it to a separate function

src/components/footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  completedTodosCount,
+}) => {
+  const [selectedFilter, setSelectedFilter] = useState(SelectOptions.ALL);
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {todosCount}
+        {' '}
+        items left
```
Anastasiia-Svintsova, line 23:
```suggestion
        {`${todosCount} items left`}
```

src/components/footer.tsx
```diff
@@ -0,0 +1,84 @@
...
+          href="#/completed"
+          className={classNames(
+            'filter__link', {
+              selected: selectedFilter === SelectOptions.COMPLETED,
+            },
+          )}
+          onClick={() => {
+            setSelectedFilter(SelectOptions.COMPLETED);
+            filterTodos(SelectOptions.COMPLETED);
+          }}
```
Anastasiia-Svintsova, line 67: So, you can create a method for it. You do almost the same for the active and all

src/components/header.tsx
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React from 'react';
+
+export const Header: React.FC = () => {
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
```
Anastasiia-Svintsova, line 7: Remove unnecessary comments, please

src/components/todoInfo.tsx
```diff
@@ -0,0 +1,67 @@
...
+              type="text"
+              className="todo__title-field"
+              placeholder="Empty todo will be deleted"
+              value={editedValue}
+              onChange={onInputChange}
+              onBlur={onInputBlur}
+            />
+          </form>
+        )
+        : (
```
Anastasiia-Svintsova, line 51:
```suggestion
      {isTodoEditing ? (
          <form>
            <input
              type="text"
              className="todo__title-field"
              placeholder="Empty todo will be deleted"
              value={editedValue}
              onChange={onInputChange}
              onBlur={onInputBlur}
            />
          </form>
        ) : (
```

src/components/todoInfo.tsx
```diff
@@ -0,0 +1,67 @@
...
+    title,
+    completed,
+  } = todoInfo;
+
+  const onInputChange = ({ target }:React.ChangeEvent<HTMLInputElement>) => {
+    setEditedValue(target.value);
+  };
+
+  const onInputBlur = () => {
+    setIsTodoEditing(false);
```
Anastasiia-Svintsova, line 25: You don't need to implement editing in this part of the task. And btw it doesn't work now.

Anastasiia-Svintsova (changes requested):
- The footer design doesn't match the working version
image
image

- Why here is `0` while todo is loading? Wait until todos will be loaded before show smth

https://user-images.githubusercontent.com/99403198/222122060-04abe5f6-3f10-410b-8eab-c300c500cb33.mov

src/components/errorNotification.tsx
miamarichka: Because I call it on click event, If I dont, I`ll get this 
image

src/components/todoInfo.tsx
miamarichka: I need to so set this  setIsTodoEditing, It works in this way, I dont takes edited values or etc, so there are no functional for this

src/components/footer.tsx
miamarichka: > * Why here is `0` while todo is loading? Wait until todos will be loaded before show smth

Because It was short condition before render and it worked in that way. Now condition longer it dosent show 0

src/App.tsx
```diff
@@ -1,10 +1,91 @@
...
+    loadTodosFromServer();
+  }, []);
+
+  useEffect(() => {
+    const timeoutID = setTimeout(() => {
+      setOnError(false);
+    }, 3000);
+
+    return () => clearTimeout(timeoutID);
+  }, []);
```
anstsot, line 69: small advice for the future. if you want to hide error after 3 seconds you have to add `onError` as a dependency and if it's true set timeout to hide.

src/App.tsx
```diff
@@ -14,143 +95,26 @@ export const App: React.FC = () => {
...
+        {todos.length > 0
+          && <Main todos={visibleTodos} addComplitedTodo={addComplitedTodo} />}
+
+        {todos && (
+          <Footer
+            filterTodos={filterTodosBySelectOptions}
+            todosCount={todos.length}
+            completedTodosCount={completedTodosId.length}
+          />
+        )}
```
anstsot, line 109: `todos` is `[]` so it would always return true. you can remove this check
```suggestion
        <Footer
          filterTodos={filterTodosBySelectOptions}
          todosCount={todos.length}
          completedTodosCount={completedTodosId.length}
        />
```

src/components/errorNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <div className={classnames('notification', 'is-danger',
+      'is-light', 'has-text-weight-normal', { hidden: isErrorOccuring })}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={onCloseHandler}
```
anstsot, line 24: you can use the `onNotificationClose` directly here without additional wrapper.
```suggestion
        onClick={onNotificationClose}
```

src/components/errorNotification.tsx
anstsot: I've added the comment below

anstsot (approved):
Great job, approved!
I've left several comments (mostly advice) for the future

PR #353, opened 2023-02-28
Solution

KrisMakarovska:
[DEMO LINK](https://KrisMakarovska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,57 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { FilteredBy } from './types/FilteredBy';
+import { getTodos } from './api/todos';
+import { Footer } from './Footer.tsx';
+import { TodoList } from './TodoList';
+import { Header } from './Header';
+import { Notifications } from './Notification';
+
+const USER_ID = 6459;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToShow, setTodosToShow] = useState<Todo[]>(todos);
+  const [query, setQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState<FilteredBy>(FilteredBy.ALL);
+  const [errorFromServer, setErrorFromServer] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    const filteredTodos = todos.filter(todo => {
+      switch (todoStatus) {
+        case FilteredBy.ALL:
+          return true;
+
+        case FilteredBy.ACTIVE:
+          return !todo.completed;
+
+        case FilteredBy.COMPLETED:
+          return todo.completed;
+
+        default:
+          return todo;
+      }
+    });
+
+    setTodosToShow(filteredTodos);
+  }, [todos, todoStatus]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setErrorFromServer(false);
+      })
+      .catch(() => {
+        setErrorFromServer(true);
+        setErrorMessage('upload');
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosToShow, setTodosToShow] = useState<Todo[]>(todos);
+  const [query, setQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState<FilteredBy>(FilteredBy.ALL);
+  const [errorFromServer, setErrorFromServer] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    const filteredTodos = todos.filter(todo => {
```
agniya-i, line 24: it would be better to use `useMemo` instead of making this in `useEffect`

src/Footer.tsx.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const Footer: React.FC<Props> = ({
+  todosToShow,
+  todoStatus,
+  setTodoStatus,
+}) => {
+  const activeTodosLeft = todosToShow.filter(todo => !todo.completed);
+
+  return (
+    // {/* Hide the footer if there are no todos */} ++
```
agniya-i, line 20: you can remove comments

src/Header.tsx
```diff
@@ -0,0 +1,40 @@
...
+  setQuery: (query: string) => void,
+  todos: Todo[],
+};
+
+export const Header: React.FC<Props> = ({
+  query,
+  setQuery,
+  todos,
+}) => {
+  const isActicve = todos.filter(todo => !todo.completed);
```
agniya-i, line 15:
```suggestion
  const activeTodos = todos.filter(todo => !todo.completed);
```
use `useMemo` here as well

src/Notification.tsx
```diff
@@ -0,0 +1,51 @@
...
+  setErrorFromServer: (errorFromServer: boolean) => void,
+  errorMessage: string,
+};
+
+export const Notifications: React.FC<Props> = ({
+  errorFromServer,
+  setErrorFromServer,
+  errorMessage,
+}) => {
+  const [hidden, setHidden] = useState(true);
```
agniya-i, line 15:
```suggestion
  const [isHidden, setIsHidden] = useState(true);
```

agniya-i (changes requested):

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+import { Notifications } from './Notification';
+
+const USER_ID = 6459;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState<FilteredBy>(FilteredBy.ALL);
+  const [errorFromServer, setErrorFromServer] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
```
anstsot, line 20: you can use one error state instead of 2. just check always if there is something in `errorMessage`
```suggestion
  const [errorMessage, setErrorMessage] = useState('');
```

src/Notification.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const notificationCloser = () => {
+      const timerId = window.setTimeout(() => {
+        setisHidden(true);
+        setErrorFromServer(false);
+      }, 3000);
+
+      window.clearTimeout(timerId);
+    };
+
+    if (errorFromServer) {
```
anstsot, line 27: you can use the `errorMessage`
```suggestion
    if (errorMessage.length > 0) {
```

src/Notification.tsx
```diff
@@ -0,0 +1,51 @@
...
+      setisHidden(false);
+      notificationCloser();
+    } else {
+      setisHidden(true);
+    }
+  });
```
anstsot, line 33: please add the missing dependency here
```suggestion
  }, [errorMessage]);
```

anstsot (changes requested):
Great job with hooks!
Let's just reduce error state ti use one only

src/Notification.tsx
witflash: @KrisMakarovska
What about this comment? Maybe you forgot about it?

witflash (approved):
Well done 👍 

But don't leave unresolved comments from prev reviewers.

src/Notification.tsx
KrisMakarovska: > @KrisMakarovska What about this comment? Maybe you forgot about it?

Yes, my mistake. Already fixed it

PR #358, opened 2023-03-01
solution

Deniskakaka:
https://deniskakaka.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,13 +1,13 @@
-import { useEffect } from 'react';
+import { useEffect, FC } from 'react';
 
 /* eslint-disable jsx-a11y/control-has-associated-label */
 type Props = {
   errorText: string,
   errorClear: () => void
 };
-export const Error: React.FC<Props> = ({ errorClear, errorText }) => {
+export const Error: FC<Props> = ({ errorClear, errorText }) => {
   useEffect(() => {
-    const idTimer = setTimeout(() => errorClear(), 3000);
+    const idTimer = setTimeout(errorClear, 3000);
 
     if (errorText) {
       // eslint-disable-next-line @typescript-eslint/no-unused-expressions
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -5,10 +5,16 @@ type Props = {
   setFilter: (filter?: boolean) => void
 };
 
-export const Filter: React.FC<Props> = ({ setFilter }) => {
-  const [typeFilter, setTypeFilter] = useState('all');
+enum TypeFilter {
+  'all',
+  'active',
+  'completed',
+}
 
-  const hcooseFilter = (type: string, filter?: boolean) => {
+export const Filter: React.FC<Props> = ({ setFilter }) => {
+  const [typeFilter, setTypeFilter] = useState(TypeFilter[0]);
+
+  const chooseFilter = (type: string, filter?: boolean) => {
     setFilter(filter);
     setTypeFilter(type);
   };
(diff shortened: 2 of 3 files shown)


src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
...
+  errorClear: () => void
+};
+export const Error: React.FC<Props> = ({ errorClear, errorText }) => {
+  useEffect(() => {
+    const idTimer = setTimeout(() => errorClear(), 3000);
+
+    if (errorText) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      idTimer;
+    }
```
OlEzhA0, line 15: why do we need this block?

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect } from 'react';
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
+type Props = {
+  errorText: string,
+  errorClear: () => void
+};
+export const Error: React.FC<Props> = ({ errorClear, errorText }) => {
+  useEffect(() => {
+    const idTimer = setTimeout(() => errorClear(), 3000);
```
OlEzhA0, line 10:
```suggestion
    const idTimer = setTimeout(errorClear, 3000);
```

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect } from 'react';
```
OlEzhA0, line 1:
```suggestion
import { useEffect, FC } from 'react';
```

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
+type Props = {
+  errorText: string,
+  errorClear: () => void
+};
+export const Error: React.FC<Props> = ({ errorClear, errorText }) => {
```
OlEzhA0, line 8:
```suggestion
export const Error: FC<Props> = ({ errorClear, errorText }) => {
```

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useState } from 'react';
+
+type Props = {
+  setFilter: (filter?: boolean) => void
+};
+
+export const Filter: React.FC<Props> = ({ setFilter }) => {
+  const [typeFilter, setTypeFilter] = useState('all');
+
+  const hcooseFilter = (type: string, filter?: boolean) => {
```
OlEzhA0, line 11:
```suggestion
  const chooseFilter = (type: string, filter?: boolean) => {
```

OlEzhA0: would be better to also add useCallback

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,50 @@
+import classNames from 'classnames';
+import { useState } from 'react';
+
+type Props = {
+  setFilter: (filter?: boolean) => void
+};
+
+export const Filter: React.FC<Props> = ({ setFilter }) => {
+  const [typeFilter, setTypeFilter] = useState('all');
```
OlEzhA0, line 9: try to use Enum

OlEzhA0 (changes requested):

src/components/Error/Error.tsx
Deniskakaka: The error is thrown because FC writes that they are not used

VitaliyBondarenko1982 (approved):

PR #359, opened 2023-03-02
initial

kostasburnazaki:
[DEMO LINK](https://kostasburnazaki.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,43 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import classNames from 'classnames';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import { USER_ID } from './utils/fetchClient';
+import { getTodos } from './api/todos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos]
+    = useState<Todo[]>([]);
+  const [activeTodos, setActiveTodos]
+    = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [customError, setError] = useState({ active: false, text: '' });
+
+  const hideError = () => {
+    setError({ ...customError, active: false });
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setCompletedTodos(todosFromServer
+          .filter(todo => todo.completed));
+        setActiveTodos(todosFromServer.filter(todo => !todo.completed));
+        setVisibleTodos(todosFromServer);
+      })
+      .catch(() => {
+        setError({ active: true, text: 'Unable to update a todo' });
+        setTimeout(hideError, 3000);
+      });
+  }, []);
+
+  useEffect(() => {
+    hideError();
+  }, [visibleTodos]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
@@ -15,10 +48,14 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
-          <button type="button" className="todoapp__toggle-all active" />
+          <button
+            type="button"
+            className={classNames(
+              'todoapp__toggle-all',
+              { active: Boolean(activeTodos.length) },
+            )}
+          />
 
-          {/* Add a todo on form submit */}
           <form>
             <input
               type="text"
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,10 +1,43 @@
...
+import { USER_ID } from './utils/fetchClient';
+import { getTodos } from './api/todos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedTodos, setCompletedTodos]
+    = useState<Todo[]>([]);
+  const [activeTodos, setActiveTodos]
+    = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```
slavanosatskyi, line 15: try to reimplement it in such way that you have only todos and filter state. And based on this data you will always be able to filter out needed todos. Just imagine, that you hundreds of todos, then you will need to store almost three times more todos in memory for each filter type. Also in future it might happen that new filters will be introduces and you will need to add more filtered copies of initial todos

src/App.tsx
```diff
@@ -15,10 +48,14 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
-          <button type="button" className="todoapp__toggle-all active" />
+          <button
```
slavanosatskyi, line 51: jsx looks fine, but lets try to split it into separate components - for example header, todoList and footer (feel free to come up with your own component names)

slavanosatskyi (changes requested):

src/App.tsx
kostasburnazaki: I had this idea but I didn't know what is more heavy: filtering every render or keeping more copies...

src/App.tsx
```diff
@@ -1,10 +1,70 @@
...
+
+import { USER_ID } from './utils/fetchClient';
+import { initData } from './constants/initData';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(initData.todos);
+
+  const [filter, setFilter] = useState<Filter>(initData.filter);
+
+  const [customError, setError] = useState<CustomError>(initData.customError);
```
DarkMistyRoom, line 25:
```suggestion
  const [todos, setTodos] = useState<Todo[]>(initData.todos);
  const [filter, setFilter] = useState<Filter>(initData.filter);
  const [customError, setError] = useState<CustomError>(initData.customError);
```

src/types/CustomError.ts
```diff
@@ -0,0 +1,4 @@
+export type CustomError = {
+  active: boolean,
+  text: string,
+};
```
DarkMistyRoom, line 4: would be better to use `enum `and list all possible errors with appropriate messages

src/App.tsx
```diff
@@ -14,143 +74,26 @@ export const App: React.FC = () => {
...
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <ErrorNotification
+        customError={customError}
+        hideError={hideError}
+      />
```
DarkMistyRoom, line 96: don't render this component if there is no errors

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { FC } from 'react';
+import { CustomError } from '../types/CustomError';
+
+type Props = {
+  customError: CustomError,
+  hideError: () => void,
+};
+
+export const ErrorNotification: FC<Props> = ({ customError, hideError }) => {
+  return (
```
DarkMistyRoom, line 11:
```suggestion
export const ErrorNotification: FC<Props> = ({ customError, hideError }) => (
```

DarkMistyRoom: use short syntax

src/components/ListOfTodos.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+            <button type="button" className="todo__remove">×</button>
+
+            <div className="modal overlay">
+              <div className="
+                        modal-background has-background-white-ter"
+              />
+              <div className="loader" />
+            </div>
+          </div>
```
DarkMistyRoom, line 43: move the Todo to a separate component

src/components/ListOfTodos.tsx
```diff
@@ -0,0 +1,47 @@
...
+            </label>
+
+            <span className="todo__title">{title}</span>
+
+            <button type="button" className="todo__remove">×</button>
+
+            <div className="modal overlay">
+              <div className="
+                        modal-background has-background-white-ter"
+              />
```
DarkMistyRoom, line 40:
```suggestion
              <div 
                className="
                  modal-background 
                  has-background-white-ter
                "
              />
```

DarkMistyRoom (changes requested):
Good job, just a few suggestions

src/components/TodoComponent.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    <button type="button" className="todo__remove">×</button>
+
+    <div className="modal overlay">
+      <div
+        className="
+                  modal-background
+                  has-background-white-ter
+                "
+      />
```
anstsot, line 37: it's a little hard to read so you can change it to:
```suggestion
      <div className="modal-background has-background-white-ter" />
```

src/components/ListOfTodos.tsx
```diff
@@ -0,0 +1,27 @@
...
+  <section
+    className="todoapp__main"
+  >
+    {todos.filter(filterCallback)
+      .map(({ id, title, completed }: Todo) => (
+        <TodoComponent
+          id={id}
+          completed={completed}
+          title={title}
+        />
```
anstsot, line 24: don't forget to add `key` by rendering a list of components
```suggestion
        <TodoComponent
          key={id}
          id={id}
          completed={completed}
          title={title}
        />
```

src/App.tsx
```diff
@@ -14,143 +68,28 @@ export const App: React.FC = () => {
...
-          </button>
-        </footer>
+        <Header activeTodoData={activeTodoData} />
+
+        <ListOfTodos
+          todos={todos}
+          filterCallback={filterCallback}
+        />
+
+        {Boolean(todos.length) && (
```
anstsot, line 78: you can simplify it to:
```suggestion
        {!!todos.length && (
```

src/components/ListOfTodos.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const ListOfTodos: FC<Props> = ({
+  todos,
+  filterCallback,
+}) => (
+  <section
+    className="todoapp__main"
+  >
+    {todos.filter(filterCallback)
```
anstsot, line 18: I would not recommend using `filter` function while rendering. it's better to propagate in `ListOfTodos` component filtered data

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+        return true;
+    }
+  });
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setActiveTodo({
+          hasActiveTodo: todosFromServer.some(todo => !todo.completed),
```
anstsot, line 44: `hasActiveTodo` should be boolean. so it would be better to store it properly
```suggestion
          hasActiveTodo: !!todosFromServer.some(todo => !todo.completed),
```

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

src/components/ListOfTodos.tsx
anstsot: in `App` component you have `todos` and `filter`. so you can create new variable `filteredTodos` and by using `useMemo` filter data.
it should look like that:
```
const filteredTodos = useMemo(
  () => todos.filter(...),
  [todos, filter],
);
```

anstsot: `filteredTodos` could be used in `ListOfTodos` component

kostasburnazaki: Understood. Thx!

src/App.tsx
```diff
@@ -1,10 +1,69 @@
...
+
+  const filterCallback = ((todo: Todo) => {
+    switch (filter) {
+      case Filter.Completed:
+        return todo.completed;
+      case Filter.Active:
+        return !todo.completed;
+      default:
+        return true;
+    }
```
loralevitska, line 36:
```suggestion
    switch (filter) {
      case Filter.Completed:
        return todo.completed;
        
      case Filter.Active:
        return !todo.completed;
        
      default:
        return true;
    }
```

loralevitska (approved):
Remove outline

PR #379, opened 2023-03-18
Solution

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -70,7 +70,7 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {todos.length !== 0 && (
+          {todos.length && (
             <ToggleButton isActive={isAllTodosActive} />
           )}
 
@@ -87,7 +87,7 @@ export const App: React.FC = () => {
           <TodoList todos={filteredArray} />
         </section>
 
-        {todos.length !== 0 && (
+        {todos.length && (
           <Footer
             amountOfItemsLeft={amountOfItemsLeft}
             amountOfItems={todos.length}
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
-          <button type="button" className="todoapp__toggle-all active" />
+          {todos.length !== 0 && (
```
OleziO, line 73: Replace with `todos.length`, it will be the same as `todos.length !== 0`

OleziO: Check other similar cases.

OleziO: Check other similar cases.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+          className="todo__status"
+          checked={completed}
+        />
+      </label>
+
+      <span className="todo__title">
+        {title}
+      </span>
+
+      {/* Remove button appears only on hover */}
```
OleziO, line 32: Remove all comments.

OleziO (changes requested):

TarasHoliuk (approved):
LGTM :+1: 

FYI: No need to create a separate folder and index.ts for a component with only one file. Just keep `ComponentExample.tsx` file in the `components` folder

PR #380, opened 2023-03-18
done

ghost:
[DEMO LINK](https://l4st1m0za.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -13,15 +13,11 @@ export const Footer: React.FC<Props> = ({
   sortType,
   activeTodosLength,
 }) => {
-  const handleSortType = (value: SortType) => {
-    setSortType(value);
-  };
-
   return (
     <footer className="todoapp__footer">
       <span className="todo-count">
-        { activeTodosLength }
-        items left
+        {/* eslint-disable react/jsx-one-expression-per-line */}
+        { activeTodosLength } items left
       </span>
 
       {/* Active filter should have a 'selected' class */}
@@ -31,7 +27,7 @@ export const Footer: React.FC<Props> = ({
           className={classNames('filter__link', {
             selected: sortType === SortType.ALL,
           })}
-          onClick={() => handleSortType(SortType.ALL)}
+          onClick={() => setSortType(SortType.ALL)}
         >
           All
         </a>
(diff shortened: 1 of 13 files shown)


TarasHoliuk:
The gap between the footer and the input is redundant, footer's background-color should be white:
Знімок екрана з 2023-03-19 11-25-09

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
+      <nav className="filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: sortType === SortType.ALL,
+          })}
+          onClick={() => handleSortType(SortType.ALL)}
```
TarasHoliuk, line 34:
```suggestion
          onClick={() => setSortType(SortType.ALL)}
```
`handleSortType` wrapper is redundant

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const handleSortType = (value: SortType) => {
+    setSortType(value);
+  };
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        { activeTodosLength }
+        items left
```
TarasHoliuk, line 24: There is no space between the number and the text - use a string template instead, i.e 
```
`${count} text`
```

TarasHoliuk: also, active todos are todos with `compleated === false`

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+type Props = {
+  activeTodosLength: number,
+};
+
+export const Header: React.FC<Props> = ({ activeTodosLength }) => (
+  <header className="todoapp__header">
+    {/* this buttons is active only if there are some active todos */}
```
TarasHoliuk, line 9: code says the same - comment is redundant (there is no complication logic that should be described in comments)
fix the same everywhere

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main">
```
TarasHoliuk, line 11: since it's a list it's better to use semantic tag:
```suggestion
    <ul className="todoapp__main">
```
P.S. the same inside `TodoInfo` - use `li`

TarasHoliuk (changes requested):
GJ :+1: 
But a few fixes are needed

SanyaBratashchuk (approved):

PR #383, opened 2023-03-23
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,65 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useState, useCallback, useEffect, useMemo,
+} from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { TodoList } from './components/TodoList/TodoList';
+// import collection from './api/collection.json';
+import { FilterType } from './types/FilterType';
 
-const USER_ID = 0;
+const USER_ID = 6704;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorText, setErrorText] = useState('');
+  const [filterType, setFilterType] = useState(FilterType.All);
+
+  const fetchTodos = useCallback(async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,155 +1,91 @@
...
+  const fetchTodos = useCallback(async () => {
+    try {
+      const todosFromServer = await getTodos(USER_ID);
+
+      setTodos(todosFromServer);
+    } catch (error) {
+      setErrorText('Unable to load todos');
+      setTimeout(() => {
+        setErrorText('');
+      }, 3000);
```
SanyaBratashchuk, line 28:
```suggestion
      setTimeout(closeError, 3000);
```

src/App.tsx
```diff
@@ -1,155 +1,91 @@
...
+        default:
+          return todos;
+      }
+    });
+  }, [todos, filterType]);
+
+  useEffect(() => {
+    closeError();
+    fetchTodos();
+  }, [fetchTodos]);
```
SanyaBratashchuk, line 60: move all hooks to the top of your component

src/App.tsx
```diff
@@ -1,155 +1,91 @@
...
+
+        case FilterType.Completed:
+          return todo.completed;
+
+        case FilterType.Active:
+          return !todo.completed;
+
+        default:
+          return todos;
+      }
```
SanyaBratashchuk, line 53: Use filter inside switch and not vice versa.
To avoid extra filter call in `ALL` case

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,37 @@
...
+        ))}
+        {todos
+        && (
+          <Footer
+            activeTodos={activeTodos}
+            filterType={filterType}
+            setFilterType={setFilterType}
+            completedTodos={completedTodos}
+          />
+        )}
```
SanyaBratashchuk, line 33: Why Footer should be inside TodoList component?

SanyaBratashchuk (changes requested):

SanyaBratashchuk (changes requested):
comments ⏫

src/components/TodoList/TodoList.tsx
IVxVI: My bad, confused with basic sections

SanyaBratashchuk (approved):

PR #386, opened 2023-03-27
Filter and error parts were implemented

lilia-obushenko:
[DEMO LINK](https://lilia-obushenko.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -14,143 +54,25 @@ export const App: React.FC = () => {
...
-        <button type="button" className="delete" />
-
-        {/* show only one message at a time */}
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      {error && (
```
timoxa2t, line 70: What would happen when we clear an error? I think it just suddenly disapear, and it should fade away for some time. Thats why we use here hidden class. But I dont know how it works for you. So if it works fine than fine)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+  todos: Todo[],
+  filterType: FilterType,
+  onChangeFilterType: (filterType: FilterType) => void,
+}
+
+export const Footer: FC<Props> = (props) => {
+  const { todos, filterType, onChangeFilterType } = props;
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
```
timoxa2t, line 16: Is it realy nessesary to filter your todos twise? And without useMemo? I suggest to wrap activeTodos calculation in useMemo. And because you need only size of active and completed todos for completedTodos you can just 'todos.length - activeTodos.length'

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+              selected: value === filterType,
+            })}
+            onClick={() => onChangeFilterType(value)}
+          >
+            {`${value.slice(0, 1).toUpperCase()}${value.slice(1)}`}
+          </a>
+        ))}
+      </nav>
+
+      {completedTodos.length > 0 && (
```
timoxa2t, line 39: button is a part of a flex content so if you just remova it the rest of this content will slide. In example I saw that this button fades away(like error message should).

timoxa2t (approved):
Good work. You can make some optimization.

src/components/Footer/Footer.tsx
lilia-obushenko: My final logic for this is in my final implementation) I've used useMemo there)

src/components/Footer/Footer.tsx
lilia-obushenko: Grear idea! I agree, going to fix it in my last task, thanks!

src/App.tsx
lilia-obushenko: I've made a modal for errors in my last implementation.Just some optimization, for users to see an error, no matter how much todos are. It's not a problem, it disappears in 3 sec, as in task is written. But thanks)

src/components/Footer/Footer.tsx
lilia-obushenko: Agree. Fixed it in my final solution)

OleksiiBlezniuk (approved):
reviewed in last part

PR #390, opened 2023-03-28
loading_todos

Vadym-Prydatok:


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const Footer: React.FC<Props> = ({ filter, onFilter }) => {
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        3 items left
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
+      <nav className="filter">
+        <a
```
OlEzhA0, line 19: nb: you can reduce your code, if you move some general keys to array and render this links with `.map`

OlEzhA0 (changes requested):
hey! I left few comments; check it please

src/App.tsx
```diff
@@ -1,10 +1,56 @@
...
+        return !todo.completed;
+
+      case Filter.COMPLETED:
+        return todo.completed;
+
+      default:
+        return todo;
+    }
+  });
+  const completedTodos = visibleTodos?.find(todo => (todo.completed));
```
IvanFesenko, line 52: Thats not `completedTodos`. Better namig will be `hasCompleted` and you can use `Array.some` to get boolean value as you use it
```suggestion
  const hasCompleted = visibleTodos?.some(todo => (todo.completed));
```

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,32 @@
...
+type Props = {
+  error: boolean,
+  errorNotice: string,
+  hasError: (state: boolean) => void,
+};
+
+export const Notification: React.FC<Props> = ({
+  error,
+  errorNotice,
+  hasError,
```
IvanFesenko, line 13: Naming one more time. `hasError` means that error is exist, fix to `setError`

src/helper/filterLinks.tsx
```diff
@@ -0,0 +1,5 @@
+export const filterLinks = [
+  { title: 'All', url: '' },
+  { title: 'Active', url: 'active' },
+  { title: 'Completed', url: 'completed' },
+];
```
IvanFesenko, line 5: Also you can use `Filter` enum for title

IvanFesenko (changes requested):
Left some comments, please check

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
+      <button
+        aria-label="btn"
+        type="button"
+        className="todoapp__toggle-all active"
+      />
+
+      {/* Add a todo on form submit */}
```
OleziO, line 13: Remove all comments from the code.

OleziO (approved):

PR #391, opened 2023-03-28
First part of TodoApp

Tanya-Grebenkina:
[DEMO LINK](https://Tanya-Grebenkina.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,63 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React,
+{
+  useState, useEffect, useMemo, useCallback,
+} from 'react';
+import { Todo } from './types/Todo';
+import { FilterBy } from './types/FilterBy';
+import { getTodos } from './api/todos';
+import Footer from './components/Footer';
+import Header from './components/Header';
+import Notification from './components/Notification';
+import TodoList from './components/TodoList';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 6511;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoFilterType, setTodoFilterType] = useState<FilterBy>(FilterBy.all);
+  const [query, setQuery] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+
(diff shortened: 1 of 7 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+            className={classNames('filter__link', {
+              selected: todoFilterType === FilterBy.completed,
+            })}
+            onClick={() => setTodoFilterType(FilterBy.completed)}
+          >
+            Completed
+          </a>
+        </nav>
+
+        {/* don't show this button if there are no completed todos */}
```
Anastasiia-Svintsova, line 60: Pay attention to this commented line. (And remove this line from the code)
image

src/components/Header.tsx
```diff
@@ -0,0 +1,40 @@
...
+type Props = {
+  todos: Todo[];
+  query: string;
+  onFormInputChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+const Header: React.FC<Props> = ({ todos, query, onFormInputChange }) => {
+  const activeTodos = useMemo(() => {
+    return todos.filter(todo => !todo.completed);
+  }, [todos]);
```
Anastasiia-Svintsova, line 14: Seems that you have the same calculation in the footer. So, you can move it to the highest level and pass the value as a prop

src/components/Notification.tsx
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
Anastasiia-Svintsova, line 1: Try not to disable eslint

Anastasiia-Svintsova (changes requested):
- Left items amount shouldn't change from filter to filter. It should always display the number of active todos.
image
image

- The todos can't be empty 
image

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+          return !todo.completed;
+
+        case FilterBy.completed:
+          return todo.completed;
+
+        case FilterBy.all:
+          return true;
+
+        default:
+          return todo;
```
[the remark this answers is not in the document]
witflash: @Tanya-Grebenkina 
What about this comment?
If you don't familiar with this syntax you could ask about this in Slack.

PR #402, opened 2023-04-05
develop

Anatolii-Polishchuk:
https://Anatolii-Polishchuk.github.io/react_todo-app-loading-todos/


SerhiiKirik (changes requested):
Great 🚀 
Let's make some improvements 🧑‍💻 

Filters doesn't works
image
image
image

src/components/TodosList/TodosList.tsx
```diff
@@ -0,0 +1,17 @@
...
+type Props = {
+  todos: Todo[];
+};
+
+export const TodosList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main">
+      {todos.map(todo => {
+        return <TodoItem todo={todo} key={todo.id} />;
+      })}
```
anstsot, line 14: can be simplified
```suggestion
      {todos.map(todo => (
        <TodoItem todo={todo} key={todo.id} />
      ))}
```

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+import { Filter } from './types/Filter';
+import { filterTodos } from './utils/helpers';
+import { getTodos } from './api/todos';
+import { ErrorAction } from './types/ErrorAction';
+
+const USER_ID = 6796;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```
anstsot, line 18: this state can be removed and you can use `useMemo` hook to calculate visible todos

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  useMemo(() => {
+    const filteredTodos = filterTodos(todos, filterBy);
+
+    setVisibleTodos([...filteredTodos]);
+  }, [filterBy]);
```
anstsot, line 44: `useMemo` hook returns a value so it can be rewritten to:
```suggestion
  const filteredTodos = useMemo(() => filterTodos(todos, filterBy), [todos, filterBy]);
```

src/components/NotificationError/NotificationError.tsx
```diff
@@ -0,0 +1,44 @@
...
+import classNames from 'classnames';
+import { ErrorAction } from '../../types/ErrorAction';
+
+type Props = {
+  action: ErrorAction;
+  resetError: () => void;
+};
+
+export const NotificationError: React.FC<Props> = ({ action, resetError }) => {
+  const [isNotificationVisible, setIsNotificationVisible] = useState(true);
```
anstsot, line 12: you don't need this state variable because the component won't be rendered if there is no error (you have the check in `App.tsx`)
```suggestion
```

src/components/NotificationError/NotificationError.tsx
```diff
@@ -0,0 +1,44 @@
...
+          'has-text-weight-normal',
+          { hidden: !isNotificationVisible },
+        )}
+      >
+        <button
+          type="button"
+          className="delete"
+          onClick={() => {
+            setIsNotificationVisible(false);
+          }}
```
anstsot, line 37:
```suggestion
          onClick={resetError}
```

anstsot (changes requested):

src/App.tsx
sTorba24: still not fixed

src/App.tsx
sTorba24: still not fixed

src/utils/helpers.ts
```diff
@@ -0,0 +1,19 @@
...
+export const filterTodos = (todos: Todo[], filterBy: Filter) => {
+  const filteredTodos = todos.filter(todo => {
+    switch (filterBy) {
+      case Filter.ALL:
+        return todo;
+      case Filter.ACTIVE:
+        return todo.completed === false;
+      case Filter.COMPLETED:
+        return todo.completed === true;
+      default:
```
sTorba24, line 13:
```suggestion
        return !todo.completed;
      case Filter.COMPLETED:
        return todo.completed;
      default:
```

src/utils/helpers.ts
```diff
@@ -0,0 +1,19 @@
...
+        return false;
```
sTorba24, line 14: you can 'connect' it with Filter.ALLL case and use 
```suggestion
        return true;
```

sTorba24 (changes requested):
GJ! I left you some comments

PR #404, opened 2023-04-06
Todo app loading todos

Igor-Yaremchuk:
[DEMO LINK](https://Igor-Yaremchuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,70 @@
+import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+  filter: string;
+  onSetFilter: (filter: string) => void;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  filter,
+  onSetFilter,
+}) => (
+  <footer className="todoapp__footer">
+    <span className="todo-count">
+      {`${todos.filter(todo => !todo.completed).length} items left`}
+    </span>
+
+    <nav className="filter">
+      <a
+        href="#/"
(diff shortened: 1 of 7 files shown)


Igor-Yaremchuk:
I have a problem with npm run deploy, that because demolink is not working ((

src/components/Todolist.tsx
```diff
@@ -0,0 +1,80 @@
...
+          )}
+        >
+          <label
+            className="todo__status-label"
+          >
+            <input
+              type="checkbox"
+              className="todo__status"
+              id="todo__selected"
+            />
```
polosanya, line 37: The input element has a static id attribute. It is not recommended to have the same id for multiple elements. Consider generating a unique id for each todo using todo.id.

src/components/Footer.tsx
```diff
@@ -0,0 +1,70 @@
...
+        }}
+      >
+        Completed
+      </a>
+    </nav>
+    {todos.find(todo => todo.completed) && (
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        onClick={() => {}}
```
polosanya, line 64: What's the purpose of an empty callback?

src/components/Todolist.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </label>
+          {editing
+            ? (
+              <Form
+                title={title}
+                onSubmit={() => {}}
```
polosanya, line 43: Same empty handler here

polosanya (changes requested):
Almost done!

PR #407, opened 2023-04-08
finished task

valeriiayarzhemska:
[DEMO LINK](https://valeriiayarzhemska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,74 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import {
+  FC,
+  useState,
+  useEffect,
+  useMemo,
+} from 'react';
+import classNames from 'classnames';
+import { Hearts } from 'react-loader-spinner';
+
 import { UserWarning } from './UserWarning';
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,10 +1,74 @@
...
+  const visibleTodos = useMemo(() => {
+    return getFilteredTodos(todos, filterType);
+  }, [todos, filterType]);
+
+  const activeTodos = useMemo(() => {
+    return getActiveTodos(visibleTodos);
+  }, [visibleTodos]);
+
+  const hasActiveTodos = visibleTodos
+    .some(({ completed }) => !completed);
```
vladyslav-mosor, line 52: I think you can just check if your helper function getActiveTodos return something bigger than 0

vladyslav-mosor (approved):

src/App.tsx
valeriiayarzhemska: > I think you can just check if your helper function getActiveTodos returns something bigger than 0

Id need to call function 2 times or write the result in const, think better like this, but thanks)

illiarerun (approved):

PR #413, opened 2023-04-10
solution: implement add, error, filter

abcdesign1:
[DEMO LINK](https://vadym4che.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,14 +1,108 @@
...
 
+  const handleSetNewTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { target: { value } } = event;
+
+    return setAddingTodoTitle(value);
+  };
+
+  const getNewTodoId = () => {
+    return Math.max(...todos.map(todo => todo.id)) + 1;
+  };
```
Bogdan-Mykhailov, line 57: you don't need that here and also in the next hw

src/App.tsx
```diff
@@ -19,11 +113,17 @@ export const App: React.FC = () => {
...
+          >
             <input
               type="text"
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              value={addingTodoTitle}
+              onChange={handleSetNewTitle}
             />
           </form>
         </header>
```
Bogdan-Mykhailov, line 129: all inside header should be different component

src/App.tsx
```diff
@@ -119,15 +220,27 @@ export const App: React.FC = () => {
...
 
-            <a href="#/completed" className="filter__link">
+            <a
+              href="#/completed"
+              className="filter__link"
+              onClick={handleFilterTodo}
+            >
               Completed
             </a>
           </nav>
```
Bogdan-Mykhailov, line 246: also footer should be different component

src/App.tsx
```diff
@@ -109,6 +209,7 @@ export const App: React.FC = () => {
               <div className="loader" />
             </div>
           </div>
+          {todosToShow.map(todo => <TodoItem key={todo.id} todo={todo} />)}
```
Bogdan-Mykhailov, line 212: also for correct structure of your project need to add TodoList component and inside that component you can map your TodoListItems

src/api/todos.ts
```diff
@@ -5,4 +5,6 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const addTodo = (userId: number, todo: Todo) => {
+  return client.post<Todo[]>(`/todos?userId=${userId}`, todo);
+};
```
Bogdan-Mykhailov, line 10: in this task we don't need to add that logic

Bogdan-Mykhailov (changes requested):
good job, but in readme file was said that no need to implement extra logic, also please separate your code to different components

Andr1yK (approved):

PR #434, opened 2023-04-11
todos load done

novak-k:
[DEMO LINK](https://novak-k.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,75 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
+import { getTodos } from './api/todos';
+import { SortType } from './types/SortType';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Errors } from './components/Errors';
+import { ErrorMessage } from './types/ErrorMessage';
 
-const USER_ID = 0;
+const USER_ID = 6713;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortType, setSortType] = useState<SortType>(SortType.ALL);
+  const [errorMessage, setErrorMessage]
+    = useState<ErrorMessage>(ErrorMessage.NONE);
+
+  const visibleTodos = useMemo(() => {
+    return (todos.filter((todo) => {
+      switch (sortType) {
+        case SortType.ACTIVE:
+          return !todo.completed;
+        case SortType.COMPLETE:
+          return todo.completed;
+        default:
+          return true;
+      }
+    })
+    );
+  }, [sortType, todos]);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const todosFromServer = await getTodos(USER_ID);
+
+        setTodos(todosFromServer);
+      } catch (error) {
+        setErrorMessage(ErrorMessage.LOAD);
+      }
+    };
+
+    fetchData();
+  }, []);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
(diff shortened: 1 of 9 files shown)


src/types/ErrorMessage.ts
```diff
@@ -0,0 +1,7 @@
+export enum ErrorMessage {
+  NONE = '',
+  LOAD = 'Unable to load a todo',
+  ADD = 'Unable to add a todo',
+  DELETE = 'Unable to delete a todo',
+  UPDATE = 'Unable to apdate a todo',
```
TarasHoliuk, line 6: There is a typo in the description of the UPDATE error message. Change 'apdate' to 'update' 📝

src/components/Errors.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ErrorMessage } from '../types/ErrorMessage';
+
+type Props = {
+  errorMessage: ErrorMessage,
+  closeError: () => void,
+};
+
+export const Errors: React.FC<Props> = ({ errorMessage, closeError }) => {
+  useEffect(() => {
+    setTimeout(() => {
```
TarasHoliuk, line 12: Clear the setTimeout by returning a cleanup function in the useEffect to avoid errors when the component unmounts. 🧹

src/App.tsx
```diff
@@ -1,156 +1,75 @@
...
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        <Header activeTodosLength={visibleTodos.length} />
+        <TodoList todos={visibleTodos} />
+
+        {TodoList.length > 0 ? (
```
TarasHoliuk, line 56: Replace 'TodoList.length' with 'todos.length' to correctly check if there are any todos.
Use conditional rendering with `&&` operator instead of ternary in a such cases

src/App.tsx
```diff
@@ -1,156 +1,75 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
TarasHoliuk, line 1: Consider removing the 'eslint-disable jsx-a11y/control-has-associated-label' comment since it's not needed anymore.

src/components/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+  setSortType,
+  sortType,
+  activeTodosLength,
+}) => {
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {activeTodosLength}
+        {' '}
+        items left
```
TarasHoliuk, line 21: use interpolation instead

TarasHoliuk (changes requested):
GJ! I've added a few minor comments

src/components/Errors.tsx
novak-k: not sure that I understand that point correctly

src/components/TodoList.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <ul className="todoapp__main">
+      {
+        todos.map(todo => {
+          return <TodoInfo key={todo.id} todo={todo} />;
+        })
+      }
```
anstsot, line 16: it can be simplified
```suggestion
      {todos.map(todo => (
        <TodoInfo key={todo.id} todo={todo} />
      )}
```

src/components/Errors.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const Errors: React.FC<Props> = ({ errorMessage, closeError }) => {
+  useEffect(() => {
+    return () => {
+      setTimeout(() => {
+        closeError();
+      }, 3000);
+    }
+  }, []);
+
```
anstsot, line 18: comment from the previous mentor means that you have to clean up the `setTimeout` on unmount
```suggestion
  useEffect(() => {
    let timerId = setTimeout(() => { closeError(); }, 3000);
    
    return () => {
      clearTimeout(timerId);
    };
  }, []);

```

anstsot (approved):
Good work, approving! But please check the comments I've left

src/components/Errors.tsx
novak-k: thanks, I will fix it in the next task

PR #455, opened 2023-04-22
Solution by DK

DimaK123:
[DEMO LINK](https://Dimak123.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -1,61 +1,23 @@
-import React, { useState } from 'react';
+import React from 'react';
 import classNames from 'classnames';
+import { FilterParams } from './FilterParams';
 
 type Props = {
-  setFilterParamHandler: (param: string) => void
+  setFilterParamHandler: (param: FilterParams) => void
+  filterParam: FilterParams
 };
 
-export const Filter: React.FC<Props> = ({ setFilterParamHandler }) => {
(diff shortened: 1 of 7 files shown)


src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  setFilterParamHandler: (param: string) => void
+};
+
+export const Filter: React.FC<Props> = ({ setFilterParamHandler }) => {
+  const [activeLink, setActiveLink] = useState({
```
agniya-i, line 9: Instead of adding an extra state, you can pass `filterParam` to this component and use it to display active link

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+  };
+
+  return (
+    <nav className="filter">
+      <a
+        href="#/"
+        className={classNames(
+          'filter__link',
+          { selected: activeLink.all },
```
agniya-i, line 56:
```suggestion
          { selected: filterParam === FilterParams.all },
```

Here's an example

agniya-i (changes requested):

src/components/Filter/Filter.tsx
DimaK123: All done
Check please :)

PR #457, opened 2023-04-25
done

andrudenko:
[DEMO LINK](https://andrudenko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,42 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { Notification } from './components/Notification/Notification';
+import { TodoList } from './components/TodoList/TodoList';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 9943;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<Status>(Status.All);
+  const [error, setError] = useState(false);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => setTodos(todosFromServer))
+      .catch(() => {
+        setError(true);
+        setTimeout(() => {
+          setError(false);
+        }, 3000);
+      });
+  }, [todos]);
+
+  const filteredTodos = () => {
+    switch (filterType) {
+      case Status.Active:
+        return todos.filter(todo => !todo.completed);
+      case Status.Completed:
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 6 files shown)


src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,23 @@
...
+    )}
+    >
+      <button
+        type="button"
+        className="delete"
+        aria-label="Mute volume"
+        onClick={() => setOnButtonDelete(false)}
+      />
+
+      Unable to add a todo
```
loralevitska, line 20: You do not add there, but only upload, the text must correspond to the actions
```suggestion
      Unable to load a todo
```

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,14 @@
...
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <>
+    {todos.map((todo: Todo) => (
+      <TodoItem key={todo.id} todo={todo} />
+    ))}
+  </>
```
loralevitska, line 13: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,30 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  return (
+    <div
```
loralevitska, line 10: Use `<li>` here due to comment about semantic tags

loralevitska (changes requested):
Almost done) Let's improve

src/App.tsx
```diff
@@ -1,10 +1,42 @@
...
+const USER_ID = 9943;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<Status>(Status.All);
+  const [error, setError] = useState(false);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(todosFromServer => setTodos(todosFromServer))
```
DarkMistyRoom, line 20:
```suggestion
      .then(setTodos)
```

DarkMistyRoom: you can use the short form

src/App.tsx
```diff
@@ -1,10 +1,42 @@
...
+  const filteredTodos = () => {
+    switch (filterType) {
+      case Status.Active:
+        return todos.filter(todo => !todo.completed);
+      case Status.Completed:
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  };
```
DarkMistyRoom, line 38: Looks like a function. So would be better to call it `filterTodos`

DarkMistyRoom: Or use a variable with defined value. Here is no need to use a function

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,16 @@
...
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => (
+  <ul style={{ listStyleType: 'none' }}>
+    {todos.map((todo: Todo) => (
+      <li key={todo.id}>
+        <TodoItem todo={todo} />
+      </li>
```
DarkMistyRoom, line 13: you can use `li` inside the `TodoItem`, instead of `div`. It will reduce tags

DarkMistyRoom (changes requested):
Almost there! A few suggestions:

OleziO (approved):

PR #458, opened 2023-04-26
Created a todo app

prokivskaaa:
[DEMO LINK](https://prokivskaaa.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,16 +12,16 @@ const USER_ID = 9953;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [error, setError] = useState<string | null>(null);
+  const [error, setError] = useState<string>('');
   const [status, setStatus] = useState<Statuses>(Statuses.ALL);
 
-  const visibleTodos = todos.filter((todo) => {
+  const visibleTodos = todos.filter(({ completed }) => {
     switch (status) {
       case Statuses.ACTIVE:
-        return !todo.completed;
+        return completed;
 
       case Statuses.COMPLETED:
-        return todo.completed;
+        return completed;
 
       default:
         return true;
@@ -43,7 +43,7 @@ export const App: React.FC = () => {
 
     if (error) {
       timeout = setTimeout(() => {
-        setError(null);
+        setError('');
       }, 3000);
     }
 
@@ -102,7 +102,7 @@ export const App: React.FC = () => {
         <button
           type="button"
           className="delete"
-          onClick={() => setError(null)}
+          onClick={() => setError('')}
         />
         {error}
       </div>


src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
+      .then((fetchedTodos: Todo[]) => {
+        setTodos(fetchedTodos);
+      })
+      .catch((fetchedError: Error) => {
+        setError(fetchedError?.message ?? 'Something went wrong');
+      });
+  }, []);
+
+  useEffect(() => {
+    let timeout: NodeJS.Timeout;
```
PashaMalyshkin, line 42:
```
I am not sure but if you want to save your timer in variable you should do this, but it's just suggestion
    let timerId: number | null = null;
```

src/types/Common.ts
```diff
@@ -0,0 +1,5 @@
+export enum Statuses {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
```
PashaMalyshkin, line 5:
```
I think you don't need values if you don't use it 
export enum Statuses {
  ALL,
  ACTIVE ,
  COMPLETED,
}
```

src/App.tsx
```diff
@@ -1,10 +1,57 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import classNames from 'classnames';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { TodoList } from './components/Todolist/TodoList';
+import { TodosFilter } from './components/TodosFilter/TodosFilter';
```
PashaMalyshkin, line 6: I think it will be better when you add index.ts or index.tsx export file with `export * from './TodoList`...

PashaMalyshkin (approved):
Good Job!

igorkruz (approved):
LGTM 👍

src/App.tsx
ManoylovAC, line 1: you can split this component into smaller ones like header, footer, etc

ManoylovAC (changes requested):
Hey, good start!
There are several comments for improvement
Please review your main component: it can be split it into smaller components.
Also, please review the filtering logic, as a result, it does not display the data quite correctly.
I believe you will succeed!

src/App.tsx
manoylov: Sorry, I was in a hurry with the comment, but in general, such timeouts in the browser return numbers, but in the NodeJs environment, they are really Timers. Within this project, the typescript settings work with NodeJs type for timers, so you can leave it as is.

src/App.tsx
```diff
@@ -28,109 +75,13 @@ export const App: React.FC = () => {
...
-          {/* Active filter should have a 'selected' class */}
-          <nav className="filter">
-            <a href="#/" className="filter__link selected">
-              All
-            </a>
-
-            <a href="#/active" className="filter__link">
-              Active
-            </a>
+          <span className="todo-count">3 items left</span>
```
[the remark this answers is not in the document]
prokivskaaa: @ManoylovAC It's not required in this task, we should do it in the next one

prokivskaaa:
@ManoylovAC Thanks, done!

liliia-livandovskaya (approved):
Great job! 👍

PR #480, opened 2023-05-08
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/react_todo-app-loading-todos/)


src/utils/constants.ts
```diff
@@ -0,0 +1 @@
+export const FILTER_OPTIONS = ['All', 'Active', 'Completed'];
```
nazarmatsevych, line 1: Consider adding a type for the FILTER_OPTIONS constant to improve type safety.

src/types/Error.ts
```diff
@@ -0,0 +1,4 @@
+export type Error = {
+  title: string,
+  isDanger: boolean,
```
nazarmatsevych, line 3: Rename 'isDanger' to 'isError' or another suitable name to indicate that the property represents an error state.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  setTodosToRender: (arr: Todo[]) => void;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  todosToRender,
+  setTodosToRender,
+}) => {
+  const [selectedButton, setSelectedButton] = useState('All');
+  const itemsLeft
```
nazarmatsevych, line 17: Consider moving the itemsLeft calculation into a useCallback or useMemo hook to avoid unnecessary re-renders.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+  = todos.reduce((acc, item) => (!item.completed ? acc + 1 : acc), 0);
+
+  const filterTodos = (e: React.MouseEvent<HTMLAnchorElement>) => {
```
nazarmatsevych, line 20: Consider using the event handler type React.MouseEventHandler<HTMLAnchorElement> to avoid using 'e.currentTarget.innerText'.

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todosToRender: Todo[] | undefined;
```
nazarmatsevych, line 5: Change the type of todosToRender from 'Todo[] | undefined' to 'Todo[]'. This will make the code less prone to errors.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Error } from '../../types/Error';
+
+type Props = {
+  errors: Error[];
+  setErrors: (cb: (prev: Error[]) => Error[]) => void;
+};
+
+export const Notification: React.FC<Props> = ({ errors, setErrors }) => {
+  const [isHidden, setIsHidden] = useState(false);
+  const { title, isDanger } = errors[0];
```
nazarmatsevych, line 12: Accessing 'errors[0]' directly may cause an error if the 'errors' array is empty. It's better to destructure the first element with a default value or check if the array is not empty before accessing it.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,37 @@
...
+  useEffect(() => {
+    setInterval(closeError, 3000);
+  }, []);
+
+  return (
+    <div className={`notification${isDanger ? ' is-danger' : ' is-light'} has-text-weight-normal${isHidden ? ' hidden' : ''}`}>
+      <button
+        type="button"
+        className="delete"
+        onClick={() => closeError()}
```
nazarmatsevych, line 31: The 'closeError' function is already wrapped in an arrow function, so you can directly pass 'closeError' as the onClick handler without the extra arrow function.

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const closeError = () => {
+    setIsHidden(true);
+    wait(900).then(() => {
+      setErrors(prev => prev.slice(1));
+      setIsHidden(false);
+    });
+  };
+
+  useEffect(() => {
```
nazarmatsevych, line 22: You are using 'setInterval' to close the error after 3 seconds, but you are not clearing the interval when the component is unmounted. Use 'clearInterval' inside a useEffect cleanup function to avoid potential memory leaks.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Todo } from '../../types/Todo';
+import { Loader } from '../Loader';
+
+type Props = {
+  todo: Todo,
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const { completed, title } = todo;
+  const [isBeingEdited] = useState(false);
```
nazarmatsevych, line 11: The isBeingEdited state should have a setter function, e.g., const [isBeingEdited, setIsBeingEdited] = useState(false);

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const [isLoading] = useState(false);
```
nazarmatsevych, line 12: The isLoading state should have a setter function, e.g., const [isLoading, setIsLoading] = useState(false);

nazarmatsevych (changes requested):
Good work, 
pls check comments and implemend filtering todos by status
image

src/types/Error.ts
vlad-demchuk: https://github.com/mate-academy/react_todo-app-loading-todos/pull/480/files#r1187301818
not fixed

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  todos,
+  todosToRender,
+  setTodosToRender,
+}) => {
+  const [selectedButton, setSelectedButton] = useState('All');
+  const itemsLeft = useMemo(() => {
+    return todos.reduce((acc, item) => (!item.completed ? acc + 1 : acc), 0);
+  }, [todos]);
+
+  // console.log(itemsLeft);
```
vlad-demchuk, line 21: remove comments

vlad-demchuk (changes requested):

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+  todosToRender: Todo[]
+  setTodosToRender: (arr: Todo[]) => void;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  todosToRender,
+  setTodosToRender,
+}) => {
+  const [selectedButton, setSelectedButton] = useState('All');
```
anstsot, line 16: it will be great introduce `enum FilterType` and use it everywhere
```suggestion
  const [selectedButton, setSelectedButton] = useState(FilterType.All);
```

anstsot (approved):
Good work, approved

PR #499, opened 2023-05-11
Start

vksazonov:
[DEMO LINK](https://vksazonov.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,20 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useState, useEffect } from 'react';
-import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { getTodos } from './api/todos';
 import { Todo } from './types/Todo';
+import { Sort } from './utils/enums';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { Header } from './components/Header';
 
 const USER_ID = 10284;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [sort, setSort] = useState('All');
+  const [sort, setSort] = useState(Sort.All);
   const [hasError, setHasError] = useState(false);
 
   const loadTodos = async () => {
@@ -29,15 +34,13 @@ export const App: React.FC = () => {
     loadTodos();
   }, []);
 
-  const filteredTodos = todos.filter(todo => {
+  const visibleTodos = todos.filter(todo => {
     switch (sort) {
-      case 'Active': return !todo.completed;
-      case 'Completed': return todo.completed;
+      case Sort.Active: return !todo.completed;
+      case Sort.Completed: return todo.completed;
       default: return todos;
     }
   });
-
-  const visibleTodos = filteredTodos;
 
   if (!USER_ID) {
     return <UserWarning />;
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,22 +1,60 @@
...
+
+  const filteredTodos = todos.filter(todo => {
+    switch (sort) {
+      case 'Active': return !todo.completed;
+      case 'Completed': return todo.completed;
+      default: return todos;
+    }
+  });
+
+  const visibleTodos = filteredTodos;
```
ManoylovAC, line 40: So now you have two variables with same value and purpose.
It looks like you can use filteredTodos directly or call it whatever you want at the beginning

src/App.tsx
```diff
@@ -29,128 +67,184 @@ export const App: React.FC = () => {
...
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+          ))}
+
+          {/* This is a completed todo
```
ManoylovAC, line 99: you can actually delete old code

src/App.tsx
```diff
@@ -29,128 +67,184 @@ export const App: React.FC = () => {
...
+                  { selected: sort === 'Active' })}
+                onClick={() => setSort('Active')}
+              >
+                Active
+              </a>
+
+              <a
+                href="#/completed"
+                className={classNames('filter__link',
+                  { selected: sort === 'Completed' })}
```
ManoylovAC, line 218: Literals like `Completed`, `Active`, `All` and so on  better to keep in some constants or enums. This bring you less potential typos and quick changes in case of this names updatings

ManoylovAC (approved):
Good job!
It would be great if you could add some types for the literals and clean up the old code.
You can also practice and split the project into smaller components.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,17 @@
+import { FC } from 'react';
+import { TodoInfo } from '../TodoInfo/TodoInfo';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: FC<Props> = ({ todos }) => {
+  return (
```
manoylov, line 10: you can use implicit return

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FC } from 'react';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const Header:FC<Props> = ({ todos }) => {
+  return (
```
manoylov, line 10: you can use implicit return

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  filter: Sort;
+  setSort: (sort: Sort) => void;
+}
+
+export const Footer:FC<Props> = ({
+  todos,
+  filter,
+  setSort,
+}) => {
+  return (
```
manoylov, line 17: you can use implicit return

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+import { Sort } from '../../utils/enums';
+
+interface Props {
+  todos: Todo[];
+  filter: Sort;
+  setSort: (sort: Sort) => void;
```
manoylov, line 9: it would be better call  like `onChangeSort`

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FC } from 'react';
+
+interface Props {
+  setHasError: (boolean: boolean) => void
+}
+
+export const ErrorNotification:FC<Props> = ({ setHasError }) => {
+  return (
```
manoylov, line 9: you can use implicit return

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FC } from 'react';
+
+interface Props {
+  setHasError: (boolean: boolean) => void
+}
+
+export const ErrorNotification:FC<Props> = ({ setHasError }) => {
```
manoylov, line 8: Try to avoid using names for props the way you use them for useState. Sometimes this is confusing and sounds like a hard binding.
```suggestion
export const ErrorNotification:FC<Props> = ({ onChangeError }) => {
```

src/App.tsx
```diff
@@ -1,156 +1,72 @@
...
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+const USER_ID = 10284;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sort, setSort] = useState(Sort.All);
```
manoylov, line 17: The name `sort` is a bit confusing in this case, because sorting is more about changing the order of items, and what you're actually doing is filtering.
```suggestion
  const [filter, setFilter] = useState(FilterType.All);
```

src/App.tsx
```diff
@@ -1,156 +1,72 @@
...
+import { Todo } from './types/Todo';
+import { Sort } from './utils/enums';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+// eslint-disable-next-line max-len
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+const USER_ID = 10284;
```
manoylov, line 13: as a good practice, it is better to store such constants in a separate file

manoylov (approved):
Very clean and good
I left a few small comments and suggestions for names
Keep up the good work 👍

manoylov (approved):
Very clean and good
I left a few small comments and suggestions for names
Keep up the good work 👍

PR #503, opened 2023-05-11
load and render todos

StasLukianchuk:
[DEMO LINK](https://StasLukianchuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,21 +1,54 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useState, useCallback, useEffect } from 'react';
 
-const USER_ID = 0;
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+
+import { getTodos } from './api/todos';
+
+import { Todo } from './types/Todo';
+import { Filter } from './types/FilterEnum';
+
+const USER_ID = 10336;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(Filter.ALL);
+
+  const loadTodos = useCallback(() => {
+    getTodos(USER_ID).then((loadedTodos) => {
+      setTodos(loadedTodos);
+    });
+  }, []);
+
+  const filteredTodos = (currentTodos: Todo[], option: string) => {
+    const visibleTodos = [...currentTodos];
+
+    switch (option) {
+      case Filter.ACTIVE:
+        return visibleTodos.filter((todo) => !todo.completed);
+      case Filter.COMPLETED:
+        return visibleTodos.filter((todo) => todo.completed);
+      default:
+        return visibleTodos;
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const visibleTodos = filteredTodos(todos, filter);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
+
         <header className="todoapp__header">
           {/* this buttons is active only if there are some active todos */}
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
           <button type="button" className="todoapp__toggle-all active" />
 
           {/* Add a todo on form submit */}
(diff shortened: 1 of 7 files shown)


src/components/ErrorMesage/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './ErrorMessage';
```
liliia-livandovskaya, line 1: There is a typo in the directory name. It should be 'ErrorMessage' instead of 'ErrorMesage'.

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main">
+      {todos.map(({ title, completed }) => (
```
liliia-livandovskaya, line 12: Need to add a key prop to the mapped div element to avoid warning and improve rendering performance.

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

src/App.tsx
```diff
@@ -1,21 +1,67 @@
...
+  useMemo,
+} from 'react';
 
-const USER_ID = 0;
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+import { ErrorMessage } from './components/ErrorMessage';
+
+import { getTodos } from './api/todos';
+
```
ManoylovAC, line 14: no need extra lines between internals imports

src/App.tsx
```diff
@@ -1,21 +1,67 @@
...
+      setErrorMessage('Unable to load todos');
+    }
+  }, []);
+
+  const closeError = useCallback(() => {
+    setErrorMessage('');
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    const visibleTodos = [...todos];
```
ManoylovAC, line 40: You don't modify this array so you no need to create new variable with array spread. In this case you can use todos directly.

src/App.tsx
```diff
@@ -28,129 +74,25 @@ export const App: React.FC = () => {
...
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
-      </div>
+        {todos.length !== 0 && (
```
ManoylovAC, line 77: it is better and more common:
```suggestion
        {todos.length > 0 && (
```

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,75 @@
...
+interface Props {
+  todoCounter: number;
+  filterTodos: Filter;
+  setFilterTodos: (filter: Filter) => void;
+}
+
+export const TodoFooter: React.FC<Props> = ({
+  todoCounter,
+  filterTodos,
+  setFilterTodos,
```
ManoylovAC, line 13: props naming can be better, hints
- you already know you in `Todo`Footer so any `todo` prefixes or suffixes should have a good reason or it's just extra duplicating
- try to avoid names like you use for useState

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
```
ManoylovAC, line 10: you can use implicit return

src/App.tsx
```diff
@@ -28,129 +74,25 @@ export const App: React.FC = () => {
...
-        <button type="button" className="delete" />
-
-        {/* show only one message at a time */}
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
+            <TodoFooter
+              todoCounter={todos.length}
```
ManoylovAC, line 82: I believe you need not all todos count but only left todos count

ManoylovAC (changes requested):
Good work!
Please check and fix some comments I left.
Almost on place, few small changes and ready to go.

ManoylovAC (changes requested):
Hey, your code is mostly ok but I see some lapse in logic.
When you click on todo statuses (all | active | completed) you have wrong calculation. In all cases you should show only active count of todos. Please check and fix this part.
Everything else is pretty good 👍 

image

ManoylovAC (approved):
Rock!
Good work, keep it up!

PR #507, opened 2023-05-12
 react_todo-app-loading-todos 

ymys1:
[DEMO LINK](https://ymys1.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,67 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { request } from './utils/fetchClient';
+import { Todo, TodoStatus } from './types/Todo';
+import { ErrorNotification } from './ErrorNotification';
+import { TodosMain } from './TodosMain';
+import { TodosHeader } from './TodosHeader';
+import { TodosFooter } from './TodosFooter';
 
-const USER_ID = 0;
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFileter, setStatusFilter] = useState<
+  TodoStatus.All
+  | TodoStatus.Completed
+  | TodoStatus.Uncompleted
+  >(TodoStatus.All);
+
+  const loadTodos = async () => {
+    try {
+      const response = await request<Todo[]>('/todos?userId=10332');
+
+      setTodos([...todos, ...response]);
+    } catch {
+      setError('server response error');
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const handleStatusFilter = useCallback(
+    (status: TodoStatus.All
+    | TodoStatus.Completed
+    | TodoStatus.Uncompleted) => {
+      setStatusFilter(status);
+    }, [],
+  );
+
+  let preparedTodos = todos;
+
+  if (statusFileter) {
+    preparedTodos = preparedTodos.filter(todo => {
+      switch (statusFileter) {
+        case TodoStatus.Uncompleted:
+          return !todo.completed;
+        case TodoStatus.Completed:
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+  }
+
+  const completedTodosLength = preparedTodos
+    .filter(todo => todo.completed).length;
+
+  const uncompletedTodosLength = preparedTodos
+    .filter(todo => !todo.completed).length;
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,10 +1,67 @@
...
-const USER_ID = 0;
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFileter, setStatusFilter] = useState<
+  TodoStatus.All
+  | TodoStatus.Completed
+  | TodoStatus.Uncompleted
```
Esteban4769, line 19: Not sure you need set types for `statusFilter` state

src/App.tsx
```diff
@@ -14,143 +71,29 @@ export const App: React.FC = () => {
...
+        <TodosHeader
+          uncompletedTodosLength={uncompletedTodosLength}
+        />
+
+        <TodosMain
+          todos={preparedTodos}
+        />
+
+        {todos && (
+          <TodosFooter
```
Esteban4769, line 83: It seems you dont add new components so i cant review it.

Esteban4769 (approved):

src/utils/fetchClient.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-const BASE_URL = 'https://mate.academy/students-api';
+const BASE_URL = 'https://mate.academy/students-ap';
```
liliia-livandovskaya, line 2: The BASE_URL seems to be incorrect. Please update it to the correct API URL.

src/App.tsx
```diff
@@ -1,10 +1,67 @@
...
+import { TodosHeader } from './TodosHeader';
+import { TodosFooter } from './TodosFooter';
 
-const USER_ID = 0;
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFileter, setStatusFilter] = useState<
```
liliia-livandovskaya, line 16: Typo in the variable name 'statusFileter'. Consider changing it to 'statusFilter'.

src/App.tsx
```diff
@@ -14,143 +71,29 @@
...
-        </footer>
+        <TodosHeader
+          uncompletedTodosLength={uncompletedTodosLength}
+        />
+
+        <TodosMain
+          todos={preparedTodos}
+        />
+
+        {todos && (
```
liliia-livandovskaya, line 82: Use 'todos.length > 0' instead of just 'todos' to conditionally render the TodosFooter. This ensures that the TodosFooter is only rendered when there are todos.

src/App.tsx
```diff
@@ -1,10 +1,67 @@
...
+  TodoStatus.All
+  | TodoStatus.Completed
+  | TodoStatus.Uncompleted
+  >(TodoStatus.All);
+
+  const loadTodos = async () => {
+    try {
+      const response = await request<Todo[]>('/todos?userId=10332');
+
+      setTodos([...todos, ...response]);
```
liliia-livandovskaya, line 26: Consider using 'setTodos(response)' instead of 'setTodos([...todos, ...response])' since the 'todos' state should be replaced with the new response, not merged.

src/App.tsx
```diff
@@ -1,10 +1,67 @@
...
+    } catch {
+      setError('server response error');
```
liliia-livandovskaya, line 28: It's better to catch specific errors and provide a more meaningful error message. Consider updating the catch block to handle specific error types.

liliia-livandovskaya (changes requested):
Excellent work! Please take note of the comments. Additionally, it might be beneficial to structure your code into components (such as todoList, todoItem, Error) and extract the corresponding code from the App for a more organized approach.

src/App.tsx
ymys1: Ok, on the next task there'll be more specific errors, I'll do it when will implement all other functionality.

src/utils/fetchClient.ts
ymys1: That was for test, corrected)

src/TodoComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    This todo is in loadind state
+    <div className="todo">
+      <label className="todo__status-label">
+        <input type="checkbox" className="todo__status" />
+      </label>
+
+      <span className="todo__title">Todo is being saved now</span>
+      <button type="button" className="todo__remove">×</button>
+
```
liliia-livandovskaya, line 75: remove comments

src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+const USER_ID = 10332;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [statusFilter, setStatusFilter] = useState<TodoStatus>(TodoStatus.All);
+
+  const loadTodos = async () => {
+    try {
+      const response = await client.get<Todo[]>('/todos?userId=10332');
```
liliia-livandovskaya, line 20:
```suggestion
      const response = await client.get<Todo[]>('/todos?userId=10332');
```
move `10332` to the variable, do not hardcode it. Also, you already have a function `getTodos` use it
```suggestion
      const response = await getTodos(USER_ID);
```

liliia-livandovskaya (approved):

PR #513, opened 2023-05-14
solution

Nikoramo:
[DEMO LINK](https://Nikoramo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,44 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { Notification } from './components/Notification/Notification';
+import { TodoList } from './components/TodoList/TodoList';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 10392;
 
 export const App: React.FC = () => {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,10 +1,44 @@
...
+      })
+      .catch(() => {
+        setError(true);
+        setTimeout(() => {
+          setError(false);
+        }, 3000);
+      });
+  }, [todos]);
+
+  const filteredTodos = () => {
```
anstsot, line 31: let's use `useMemo` hook here

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const Footer: React.FC<Props> = ({
+  setFilterType,
+  filterType,
+}) => {
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        3 items left
```
anstsot, line 17: remove hardcode

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,24 @@
...
+    >
+      <button
+        type="button"
+        className="delete"
+        aria-label="Mute volume"
+        onClick={() => setOnButtonDelete(false)}
+      />
+
+      {/* show only one message at a time */}
+      Unable to add a todo
```
anstsot, line 21: it will be good to remove hardcode and introduce the `enum ErrorType` to manage different types of error

anstsot (changes requested):
Good job, but there are few fixes to improve your code.
Here is wrong number because there are 2 todo items
image

PR #518, opened 2023-05-14
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,45 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 10380;
+
+type Filter = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [isError, setIsError] = useState<string | null>(null);
+  const [filterStatus, setFilterStatus] = useState<Filter>('all');
+
+  function filterTodos() {
+    switch (filterStatus) {
+      case 'all':
+        return todos;
+      case 'completed':
+        return todos?.filter(todo => todo.completed === true);
+      case 'active':
+        return todos?.filter(todo => todo.completed === false);
+      default:
+        return todos;
+    }
+  }
+
+  const visibleTodos: Todo[] | null | undefined = filterTodos();
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => setTodos(data))
+      .catch(() => {
+      });
+  });
+
+  const handleFilter = (status: Filter) => {
+    setFilterStatus(status);
+  };
+
   if (!USER_ID) {
     return <UserWarning />;
   }
@@ -15,10 +50,8 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
           <button type="button" className="todoapp__toggle-all active" />
 
-          {/* Add a todo on form submit */}
           <form>
             <input
               type="text"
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -28,128 +61,127 @@ export const App: React.FC = () => {
...
+                      </div>
+                    </li>
+                  ))}
+                </ul>
+              )}
+            </section>
+
+            <footer className="todoapp__footer">
+              <span className="todo-count">
+                3 items left
```
anstsot, line 114: remove the hardcoded value and calculate the number of left todo items

anstsot (changes requested):
It’s almost done! Let’s improve a little.
It would be great to create separate components for different elements link `TodoItem`, `ErrorNotification`, `Footer` etc.

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+        return todos?.filter(todo => todo.completed === false);
+      default:
+        return todos;
+    }
+  }
+
+  visibleTodos = useMemo(
+    filterTodos,
+    [todos, filterStatus],
+  );
```
anstsot, line 54: let's create the const variable here
```suggestion
  const visibleTodos = useMemo(
    filterTodos,
    [todos, filterStatus],
  );
```

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+  Post = 'post',
+  Delete = 'delete',
+  Patch = 'patch',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [error, setError] = useState<ErrorType | null>(null);
+  const [filterStatus, setFilterStatus] = useState<Filter>(Filter.All);
+  let visibleTodos: Todo[] | null | undefined = todos;
```
anstsot, line 32:
```suggestion
```

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+  );
+
+  function filterTodos() {
+    switch (filterStatus) {
+      case Filter.All:
+        return todos;
+      case Filter.Completed:
+        return todos?.filter(todo => todo.completed === true);
+      case Filter.Active:
+        return todos?.filter(todo => todo.completed === false);
```
anstsot, line 45:
```suggestion
        return todos?.filter(todo => todo.completed);
      case Filter.Active:
        return todos?.filter(todo => !todo.completed);
```

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+    [todos, filterStatus],
+  );
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(data => setTodos(data))
+      // eslint-disable-next-line no-console
+      .catch(() => {
+        setError(ErrorType.Get);
+        setTimeout(() => setError(null), 3000);
```
anstsot, line 62: you can clear an error inside the `ErrorNotification` component. just use `useEffect` and set error to `null` in 3000ms

anstsot (changes requested):

src/App.tsx
```diff
@@ -1,10 +1,74 @@
...
+
+export enum ErrorType {
+  Get = 'get',
+  Post = 'post',
+  Delete = 'delete',
+  Patch = 'patch',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
```
IvanFesenko, line 29: My assumption is better use empty array as default value
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type Props = {
+  error: ErrorType | null,
+  onError(type: ErrorType | null): void,
+};
+
+export const ErrorNotification: React.FC<Props> = ({ error, onError }) => {
+  useEffect(() => {
+    setTimeout(() => onError(null), 3000);
+  });
```
IvanFesenko, line 16: Do not forget add dependecies, or it will be called on each render
```suggestion
  useEffect(() => {
    setTimeout(() => onError(null), 3000);
  }, []);
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <nav className="filter">
+        <a
+          href="#/"
+          className={classNames(
+            'filter__link',
+            {
+              selected: filterStatus === Filter.All,
+            },
+          )}
+          onClick={() => onFilterChange(Filter.All)}
```
IvanFesenko, line 31: It's better co reacte separate handler instead of anonymous function

IvanFesenko (approved):
Good job. Check my comments to avoid same mistakes in the future

PR #538, opened 2023-05-21
Develop

Peti51:
[DEMO LINK](https://Peti51.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,429 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, {
+  useEffect, useState, FormEvent, useMemo,
+} from 'react';
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { client } from './utils/client';
 
-const USER_ID = 0;
+const USER_ID = 10377;
+
+function getRandomNumber(): number {
+  const generatedNumbers: number[] = [];
+  const maxAttempts = 1000;
+
+  for (let i = 0; i < maxAttempts; i += 1) {
+    const randomNumber = Math.floor(Math.random() * 1001);
+
+    if (!generatedNumbers.includes(randomNumber)) {
+      generatedNumbers.push(randomNumber);
+
+      return randomNumber;
+    }
+  }
+
+  return 0;
+}
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo[]>([]);
+  const [selectedTab, setSelectedTab] = useState('All');
+  const [inputValue, setInputValue] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isThereActiveTodo, setIsThereActiveTodo] = useState(false);
+  const [isThereCompletedTodos, setIsThereCompletedTodos] = useState(false);
+  const [isAddingTodoAllowed, setIsAddingTodoAllowed] = useState(false);
+  const [isDeleteingTodoAllowed, setIsDeleteingTodoAllowed] = useState(false);
+  const [isEditingTodoAllowed, setIsEditingTodoAllowed] = useState(false);
+  const [numberOfActiveTodos, setNumberOfActivTodos] = useState(0);
+  const [errorMessageField, setErrorMessageField] = useState(false);
+  const [newTodo, setNewTodo] = useState({
+    title: '',
+    userId: 10377,
+    completed: false,
+    id: getRandomNumber(),
+  });
+
+  const handleKeyPress = async (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter' && event.target.value.trim() !== '') {
+      event.preventDefault();
+
+      const updatedObjectTitle = {
+        ...newTodo,
+        title: event.target.value,
+        id: getRandomNumber(),
+      };
+
+      setNewTodo(updatedObjectTitle);
+      setTodo([...todo, updatedObjectTitle]);
+
+      const resetedTitle = {
+        ...newTodo,
+        title: '',
+        id: 0,
+      };
+
+      setNewTodo(resetedTitle);
+
+      setInputValue(resetedTitle.title);
+
+      try {
+        await client.post('/todos', updatedObjectTitle);
+      } catch (error) {
+        console.error('Error occurred:', error);
+      }
+    }
+
+    if (event.key === 'Enter' && inputValue.trim() === '') {
+      setIsAddingTodoAllowed(true);
+      setErrorMessageField(true);
+    }
+  };
+
+  const deleteTodo = async (id: number) => {
+    const newTodos = todo.filter((element) => {
+      return element.id !== id;
+    });
+
+    if (newTodos === todo) {
+      setIsDeleteingTodoAllowed(true);
+      setErrorMessageField(true);
+    }
+
+    setTodo(newTodos);
+
+    try {
+      await client.delete(`/todos/${id}`);
+    } catch (error) {
+      console.error('Error occurred:', error);
+    }
+  };
+
+  const changeAll = () => {
+    const chnagedArr = todo.map((element) => {
+      return {
+        ...element,
+        completed: !element.completed,
+      };
+    });
+
+    setTodo(chnagedArr);
+  };
+
+  const handleSubmit = (event: FormEvent) => {
+    event.preventDefault();
+  };
+
+  const searchTodo = async (id: number) => {
+    const updatedTodo = todo.map((obj) => {
+      if (obj.id === id) {
+        return {
+          ...obj,
+          completed: !obj.completed,
+        };
+      }
+
+      return obj;
+    });
+
+    const none = todo.some((element) => {
+      return element.id === id;
+    });
+
+    if (!none) {
+      setIsEditingTodoAllowed(true);
+      setErrorMessageField(true);
+    }
+
+    setTodo(updatedTodo);
+
+    try {
+      const todoToUpdate = todo.find((elem) => elem.id === id);
+
+      if (todoToUpdate) {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+    const todoLength = todo.filter((obj) => {
+      return obj.completed === false;
+    });
+
+    setIsThereActiveTodo(isActive);
+    setIsThereCompletedTodos(isFalse);
+    setNumberOfActivTodos(todoLength.length);
+
+    getTodo();
+  }, [newTodo, todo]);
```
Anastasiia-Svintsova, line 189: The useEffect hook should not have 'newTodo' and 'todo' as dependencies, since it may cause unnecessary re-renders. You can split the useEffect into two separate useEffects to handle the side effects independently.

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
 
-          {/* This todo is not completed */}
-          <div className="todo">
-            <label className="todo__status-label">
-              <input
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
+            {/* This todo is not completed
```
Anastasiia-Svintsova, line 281: Remove the commented code everywhere

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+import { Todo } from './types/Todo';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { client } from './utils/client';
 
-const USER_ID = 0;
+const USER_ID = 10377;
+
+function getRandomNumber(): number {
+  const generatedNumbers: number[] = [];
```
Anastasiia-Svintsova, line 15: Refactor getRandomNumber function to return a random number without using the loop. For example: `return Math.floor(Math.random() * 1001);`

src/App.tsx
```diff
@@ -1,156 +1,429 @@
+/* eslint-disable max-len */
+/* eslint-disable no-console */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
Anastasiia-Svintsova, line 3: Remove all eslint disabling

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+      return randomNumber;
+    }
+  }
+
+  return 0;
+}
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo[]>([]);
+  const [selectedTab, setSelectedTab] = useState('All');
```
Anastasiia-Svintsova, line 33: Create an enum for the filter types

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+  const [inputValue, setInputValue] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [isThereActiveTodo, setIsThereActiveTodo] = useState(false);
+  const [isThereCompletedTodos, setIsThereCompletedTodos] = useState(false);
+  const [isAddingTodoAllowed, setIsAddingTodoAllowed] = useState(false);
+  const [isDeleteingTodoAllowed, setIsDeleteingTodoAllowed] = useState(false);
```
Anastasiia-Svintsova, line 39: You don't need to implement adding, deleting, and updating todos in this part of the task. It doesn't work correctly for now anyway. So you can remove it. Follow the requirements, please 
image

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+  const [isEditingTodoAllowed, setIsEditingTodoAllowed] = useState(false);
+  const [numberOfActiveTodos, setNumberOfActivTodos] = useState(0);
+  const [errorMessageField, setErrorMessageField] = useState(false);
+  const [newTodo, setNewTodo] = useState({
+    title: '',
+    userId: 10377,
```
Anastasiia-Svintsova, line 45: You have USER_ID as a const already

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+  const visibleTodos: Todo[] = useMemo(() => todo.filter((element) => {
+    switch (selectedTab) {
+      case 'Completed':
+        return element.completed === true;
+      case 'Active':
+        return element.completed === false;
+      case 'All':
+        return todo;
+      default:
+        return todo;
```
Anastasiia-Svintsova, line 200:
```suggestion
      case 'All':
      default:
        return todo;
```

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+
+    getTodo();
+  }, [newTodo, todo]);
+
+  const visibleTodos: Todo[] = useMemo(() => todo.filter((element) => {
+    switch (selectedTab) {
+      case 'Completed':
+        return element.completed === true;
+      case 'Active':
+        return element.completed === false;
```
Anastasiia-Svintsova, line 196: Use logical opeartors 
```suggestion
        return element.completed;
      case 'Active':
        return !element.completed;
```

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+      case 'All':
+        return todo;
+      default:
+        return todo;
+    }
+  }), [todo, newTodo.completed, selectedTab]);
```
Anastasiia-Svintsova, line 202: You don't use newTodo.completed here, so remove it from the dependencies 
```suggestion
  }), [todo, selectedTab]);
```

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
-          {/* This is a completed todo */}
-          <div className="todo completed">
-            <label className="todo__status-label">
+        <div className="todoapp__content">
+          <header className="todoapp__header">
+            <button
+              type="button"
+              className={isThereActiveTodo
+                ? 'todoapp__toggle-all active'
+                : 'todoapp__toggle-all'}
```
Anastasiia-Svintsova, line 219: You can use the classnames library here

src/App.tsx
```diff
@@ -1,156 +1,429 @@
...
+          <section className="todoapp__main">
+            {todo.length > 0 && (
+              <>
+                {visibleTodos.map((task) => {
+                  if (!isLoading) {
+                    return (
+                      <div
+                        className={task.completed
+                          ? 'todo completed'
+                          : 'todo'}
```
Anastasiia-Svintsova, line 243: Same

src/utils/client.ts
```diff
@@ -1,32 +1,29 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-const BASE_URL = 'https://mate.academy/students-api';
+// const BASE_URL = 'https://mate.academy/students-api';
+const BASE_URL = 'https://mate.academy/students-api/todos?userId=10377';
```
Anastasiia-Svintsova, line 3: Don't change the base url

Anastasiia-Svintsova (changes requested):
- Filters should always be aligned to the center. Check the working example 
image

- Check these errors after loading, please 
image

qoa11a (changes requested):
Please, consider fixing all the problems the previous mentor has mentioned.

src/App.tsx
```diff
@@ -1,156 +1,371 @@
...
+  Active = 'Active',
+  Completed = 'Completed',
+}
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todo, setTodo] = useState<Todo[]>([]);
+  const [selectedTab, setSelectedTab] = useState('All');
```
polosanya, line 25:
```suggestion
  const [selectedTab, setSelectedTab] = useState(SortType.All);
```
Use Enum instead of hardcoded values

src/App.tsx
polosanya: Not fixed. In this task only loading, filtering and errors should be implemented

src/App.tsx
```diff
@@ -1,156 +1,371 @@
...
+  const [isThereActiveTodo, setIsThereActiveTodo] = useState(false);
+  const [isThereCompletedTodos, setIsThereCompletedTodos] = useState(false);
+  const [isAddingTodoAllowed, setIsAddingTodoAllowed] = useState(false);
+  const [isDeleteingTodoAllowed, setIsDeleteingTodoAllowed] = useState(false);
+  const [isEditingTodoAllowed, setIsEditingTodoAllowed] = useState(false);
+  const [numberOfActiveTodos, setNumberOfActivTodos] = useState(0);
+  const [errorMessageField, setErrorMessageField] = useState(false);
+  const [newTodo, setNewTodo] = useState({
+    title: '',
+    userId: 10377,
```
polosanya, line 37:
```suggestion
    userId: USER_ID,
```

polosanya (changes requested):
Hi! Some comments are still not fixed. Feel free to ask for help in js_chat if you have any difficulties)

src/App.tsx
Anastasiia-Svintsova: Hmm, still not fixed. Do you have any blockers here? Feel free to ask questions in the chat. Please, implement only the required tasks. You'll have the possibility to implement adding, deleting, and editing in the next parts of this task. There will be checklists and you'll be able to check yourself and find some hints. 
For now, it doesn’t work correctly. So this is the reason for rejection.
image

src/App.tsx
```diff
@@ -1,156 +1,302 @@
...
+            <label htmlFor="nameInput">
+              <button
+                id="nameInput"
+                type="button"
+                className={classNames('todoapp__toggle-all', {
+                  active: isThereActiveTodo,
+                })}
+                onClick={changeAll}
+              >
+                {null}
```
Anastasiia-Svintsova, line 154: Are you sure that it should be null here?

src/App.tsx
```diff
@@ -1,156 +1,302 @@
...
-              <input
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
+              <nav className="filter">
+                <a
+                  href="#/"
+                  className={classNames('filter__link', {
+                    selected: selectedTab === 'All',
```
Anastasiia-Svintsova, line 223: Use the enum instead of string values

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #539, opened 2023-05-24
Solution

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/footer/Footer.tsx
+++ b/src/components/footer/Footer.tsx
@@ -1,0 +1,73 @@
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+import { Select } from '../../types/Select';
+
+type Props = {
+  filterBy: string,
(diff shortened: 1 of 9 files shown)


src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+    let count = 0;
+
+    todoList.forEach(todo => {
+      if (!todo.completed) {
+        count += 1;
+      }
+    });
+
+    return count;
+  }
```
anstsot, line 28: try to calculate the number by using the `filter` method and `length`. use `useMemo` hook

anstsot (changes requested):
Add the `padding` or `margin` if the todo item is empty
image

qoa11a (approved):

PR #543, opened 2023-05-25
Develop

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,12 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
-import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { UserWarning } from './UserWarning';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { TodoForm } from './components/TodoForm/TodoForm';
+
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,156 +1,74 @@
...
 export const App: React.FC = () => {
+  const [activeFilter, setActiveFilter] = useState<Filters>('all');
+
+  const { todos, error } = useTodos(USER_ID);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const activeTodos = todos.filter(todo => !todo.completed);
```
Radoslaw-Czerniawski, line 23: I would personally do something like this:
image
image

Currently you are doing the filtering everytime the `activeFilter` changes, with `useMemo` it will only recalculate when the todos fetched from the server changes, also with this implementation you are using only one loop instead iterating over todos list two times. If you were writing an application that fetches hundreds of data it would be very beneficial.

Radoslaw-Czerniawski (changes requested):
Left one comment, overall good implementation adjust if you deem it optimal

src/App.tsx
Zibi95: Ohh I see. That's nice

Radoslaw-Czerniawski (approved):
Approved :)

PR #549, opened 2023-05-27
Solution-1

MykolaButylkov:
[DEMO LINK](https://MykolaButylkov.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,142 +1,81 @@
...
+
+  async function loadedTodos(f: (USER_ID: number) => Promise<Todo[]>) {
+    const result = await f(USER_ID);
+
+    setTodos(result);
+  }
+
+  useEffect(() => {
+    loadedTodos(getTodos);
+  }, []);
```
VitaliyBondarenko1982, line 26:
```suggestion
  async function loadedTodos() {
    const result = await getTodos(USER_ID);

    setTodos(result);
  }

  useEffect(() => {
    loadedTodos();
  }, []);
```
redundant pass function through arguments

src/App.tsx
```diff
@@ -1,142 +1,81 @@
...
+
+  let visibleTodos = todos.filter((todo) => (todo.title.includes(input)));
+
+  visibleTodos = visibleTodos.filter((todo) => {
+    switch (status) {
+      case 'active':
+        return !todo.completed;
+
+      case 'completed':
+        return !!todo.completed;
```
VitaliyBondarenko1982, line 46:
```suggestion
        return todo.completed;
```

src/App.tsx
```diff
@@ -1,142 +1,81 @@
...
+    setInput(event.target.value);
+  };
+
+  const handleStatus = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    setStatus(event.currentTarget.type);
+  };
+
+  let visibleTodos = todos.filter((todo) => (todo.title.includes(input)));
```
VitaliyBondarenko1982, line 38: for what this filtering?

src/components/main_todos-list/main_todos-list.tsx
```diff
@@ -0,0 +1,122 @@
...
+        );
+      })}
+      <div className="todo completed">
+        <label className="todo__status-label">
+          <input
+            type="checkbox"
+            className="todo__status"
+            checked
+          />
+        </label>
```
VitaliyBondarenko1982, line 48:
```suggestion
     
```
here and below remove redundant jsx
it is just examples for todo.
should be only list of todos from server

VitaliyBondarenko1982 (changes requested):
chrome-capture-2023-4-29
remove redundant jsx.
should be only list of todos from server
add logic for errors
left items amount should always render amount of not completed todos.

src/App.tsx
MykolaButylkov: oh yeah, I did it because, in my previous task experience, it was))

IvanFesenko (changes requested):
Waiting last changes

MykolaButylkov:
there are the last changes

PR #551, opened 2023-05-29
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,70 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoInfo } from './components/TodoInfo';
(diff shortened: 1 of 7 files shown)


polosanya (changes requested):
Almost done!

src/App.tsx
```diff
@@ -1,10 +1,67 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await client.get(URL);
+        const todosData = response as Todo[];
+
+        setTodos(todosData);
+      } catch (error) {
+        setIsAddError(true);
```
agniya-i, line 31: Here's the problem - in the case of the failed request on **loading** todos, you will get all three errors
Screenshot 2023-06-12 at 20 49 48

There should be only one error showing that loading is failed

agniya-i (changes requested):

agniya-i (approved):
👍

PR #552, opened 2023-05-30
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { client } from './utils/fetchClient';
 import { Todo } from './types/Todo';
@@ -147,10 +148,12 @@ export const App: React.FC = () => {
             <nav className="filter">
               <a
                 href="#/"
-                className={`filter__link ${
-                  selectedTodos === ShowTodos.All
-                    ? 'selected'
-                    : ''}`}
+                className={classNames(
+                  'filter__link',
+                  {
+                    selected: selectedTodos === ShowTodos.All,
+                  },
+                )}
                 onClick={handleFilterSelect}
               >
                 All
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,10 +1,70 @@
...
+  Active,
+  Completed,
+}
+
+const todosFromServer = client
+  .get<Todo[]>(`/todos?userId=${USER_ID}`);
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
```
demetrxx, line 21: Better use null when there is no value, instead of empty string, as it is more explicit and non-ambigious indication that the value is empty.

src/App.tsx
```diff
@@ -112,44 +136,81 @@ export const App: React.FC = () => {
...
+            {/* don't show this button if there are no completed todos */}
+            <button type="button" className="todoapp__clear-completed">
+              Clear completed
+            </button>
+          </footer>
+        )}
+
       </div>
 
       {/* Notification is shown in case of any error */}
```
demetrxx, line 195: Remove unnecessary comments

demetrxx (changes requested):
Great job! I like it!

Some moments, mentioned bellow, may be improved, though. 
Besides, we have the following mandatory requirement:
> also hide the notification before any next request

src/App.tsx
yuran007: 👍, I'm understood, thank you. Done.

src/App.tsx
yuran007: Done

sTorba24 (approved):
Good work! I added some comments to improve it

src/App.tsx
```diff
@@ -111,45 +136,76 @@ export const App: React.FC = () => {
...
+              3 items left
+            </span>
+
+            <nav className="filter">
+              <a
+                href="#/"
+                className={`filter__link ${
+                  selectedTodos === ShowTodos.All
+                    ? 'selected'
+                    : ''}`}
```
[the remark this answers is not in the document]
yuran007: Done

PR #554, opened 2023-05-31
Solution

NadieinOleh:
https://NadieinOleh
.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,155 +1,105 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Notification } from './components/Notification';
+
+const USER_ID = 10586;
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [selected, setSelected] = useState<string>('All');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hideError, setHideError] = useState<boolean>(false);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const loadTodos = async () => {
+    try {
+      setHideError(false);
+
+      const data = await getTodos(USER_ID);
+
+      if ('Error' in data) {
+        throw new Error('Error1');
+      } else {
+        setTodos(data);
+      }
+    } catch {
+      setHideError(true);
+
+      throw new Error('Error2');
+    }
+  };
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    const timerId = setTimeout(() => {
+      setHideError(false);
+    }, 3000);
+
+    return () => {
+      clearInterval(timerId);
+    };
+  }, []);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filterTodo = useCallback(() => {
+    switch (selected) {
+      case 'All':
+
+        return todos;
+
+      case 'Active':
+        return [...todos].filter(todo => !todo.completed);
+
+      case 'Completed':
+        return [...todos].filter(todo => todo.completed);
+
+      default:
+        return todos;
+    }
+  }, [selected, todos]);
+
+  const visibleTodos = filterTodo();
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this buttons is active only if there are some active todos */}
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,155 +1,105 @@
...
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filterTodo = useCallback(() => {
+    switch (selected) {
+      case 'All':
```
polosanya, line 61: Create a Enum will all possible filters instead of hardcoded values

src/App.tsx
```diff
@@ -1,155 +1,105 @@
...
+
+        return todos;
+
+      case 'Active':
+        return [...todos].filter(todo => !todo.completed);
```
polosanya, line 66:
```suggestion
        return todos.filter(todo => !todo.completed);
```
Filter is a non-mutating method. No need to create a copy before

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export const Footer: React.FC<Props> = React.memo(({
+  selectTodo,
+  selected,
+}) => {
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        3 items left
```
polosanya, line 18: It would be great to calculate left items

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  setHideErrorBtn: (hide: boolean) => void;
+  hideError: boolean
```
polosanya, line 6: Consider naming boolean variables according to these rules
image

src/App.tsx
```diff
@@ -1,155 +1,103 @@
...
+
+  useEffect(() => {
+    const timerId = setTimeout(() => {
+      setHideError(false);
+    }, 3000);
+
+    return () => {
+      clearInterval(timerId);
+    };
+  }, []);
```
vovakpro13, line 28: It's false by default, why do you need to set it false again after 3 seconds??

src/App.tsx
```diff
@@ -1,155 +1,103 @@
...
+
+  const filterTodo = useCallback(() => {
+    switch (selected) {
+      case SelectTodo.All:
```
vovakpro13, line 32:
```suggestion
      case Filters.All:
```

src/App.tsx
```diff
@@ -1,155 +1,103 @@
...
+
+      if ('Error' in data) {
+        throw new Error('Error - impossible load todo');
+      } else {
+        setTodos(data);
+      }
+    } catch {
+      setHideError(true);
+
+      throw new Error('Error 404 no connection to the server');
```
vovakpro13, line 61: this error won't be handled anyway, so it's redundant

src/App.tsx
```diff
@@ -1,155 +1,103 @@
...
+import { Notification } from './components/Notification';
+import { SelectTodo } from './types/SelectTodo';
+
+const USER_ID = 10586;
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [selected, setSelected] = useState<string>('All');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hideError, setHideError] = useState<boolean>(false);
```
vovakpro13, line 18: why wouldn't you save a certian error message instead of boolean value?

src/App.tsx
```diff
@@ -1,155 +1,103 @@
...
-            <input
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
+        <Header
+          query={query}
+          setQuery={setQuery}
+        />
+
+        {todos?.length > 0 && (
```
vovakpro13, line 85:
```suggestion
        {!!todos?.length && (
```

vovakpro13 (changes requested):

src/App.tsx
IvanFesenko: agree. Remove useless useEffect

src/App.tsx
IvanFesenko: Add todos is always array, so no need optional chaining

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <span className="todo-count">
+        {`${todoCount} items left`}
+      </span>
+
+      <nav className="filter">
+        <a
+          href="#/"
+          className={classNames('filter__link',
+            { selected: selected === SelectTodo.All })}
+          onClick={() => selectTodo(SelectTodo.All)}
```
IvanFesenko, line 28: Create separate handler instead of anonymous function

IvanFesenko (changes requested):
Not fixed previous comments

src/App.tsx
```diff
@@ -1,155 +1,101 @@
...
+    }, 3000);
+
+    return () => {
+      clearInterval(timerId);
+    };
+  }, []);
+
+  const filterTodo = useCallback(() => {
+    switch (selected) {
+      case SelectTodo.All:
```
agniya-i, line 32: Still not fixed 
It would be better to rename it to `Filters` because it shows selected filter, not selected todos

agniya-i (approved):
I approved, but please pay attention to the comment below

PR #560, opened 2023-06-01
message

EvgenyNikitinEvgeny:
https://EvgenyNikitinEvgeny.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import { TodoHeader } from './components/TodoHeader/TodoHeader';
 import { TodoFooter } from './components/TodoFooter/TodoFooter';
 import { SortEnum } from './types/sort';
 
-const USER_ID = 1058;
+const USER_ID = 10589;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,53 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoHeader } from './components/TodoHeader/TodoHeader';
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { SortEnum } from './types/sort';
 
-const USER_ID = 0;
+const USER_ID = 1058;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [select, setSelect] = useState('all');
+  const [error, setError] = useState(false);
+  const lengTodos = todos.filter(todo => todo.completed === false).length;
+
+  const getTodosAll = async () => {
+    try {
+      const receivedTodos = await getTodos(USER_ID);
+
+      setTodos(receivedTodos);
+    } catch {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,10 +1,53 @@
...
+import { TodoFooter } from './components/TodoFooter/TodoFooter';
+import { SortEnum } from './types/sort';
 
-const USER_ID = 0;
+const USER_ID = 10589;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [select, setSelect] = useState('all');
+  const [error, setError] = useState(false);
```
loralevitska, line 16:
```suggestion
  const [isError, setIsError] = useState(false);
```
Fix all cases if present
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  },
+) => {
+  return (
+
+    <footer className="todoapp__footer">
+      {/* Hide the footer if there are no todos */}
+      <span className="todo-count">
+        {lengTodos}
+        {' '}
+        items left
```
loralevitska, line 25: Add condition and right text for `1 item left`

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </span>
+
+      {/* Active filter should have a 'selected' class */}
```
loralevitska, line 28: Remove comments. Fix everywhere

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,41 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos:Todo[]
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main">
```
loralevitska, line 10: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- remove redundant rows from app 
image
- add error message according requirement 
image

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+
+        case SortEnum.ALL:
+          return !completed || completed;
+
+        case SortEnum.COMPLETED:
+          return completed;
+
+        default:
+          return !completed || completed;
+      }
```
SanyaBratashchuk, line 44: move filter inside switch case

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -1,10 +1,52 @@
...
+import { SortEnum } from './types/sort';
 
-const USER_ID = 0;
+const USER_ID = 10589;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [select, setSelect] = useState('all');
+  const [isError, setIsError] = useState(false);
+  // const [isDeleteError, setDeleteIsError] = useState(false);
```
TarasHoliuk, line 17: it's better remove comments (you will remember it till the next task : ) )
also, talking about errors, I'd recommend creating an enum with error types and setting the corresponding type in the state instead of creating a separate state for each error type

but for now, you can keep `const [isError, setIsError] = useState(false);`

FYI

TarasHoliuk (approved):
Approved :+1: 
I left one comment FYI

PR #565, opened 2023-06-02
initial solution

Ozwiena:
[DEMO LINK](https://ozwiena.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,50 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoList } from './TodoList';
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodoFilter } from './TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 10603;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState('all');
+  const [isHidden, setIsHidden] = useState(false);
+  const [query, setQuery] = useState('');
+  const [error, setError] = useState<null | string>(null);
+
+  useEffect(() => {
+    client.get<Todo[]>(`/todos?userId=${USER_ID}`)
+      .then(setTodos).catch(() => {
+        setError('Unable to load todos');
+        setTimeout(() => {
+          setError(null);
+        }, 3000);
+      });
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let filteredTodos = todos;
+
+    switch (filterType) {
+      case 'all':
+        filteredTodos = todos;
+        break;
+      case 'completed':
+        filteredTodos = todos.filter(todo => todo.completed);
+        break;
+      case 'active':
+        filteredTodos = todos.filter(todo => !todo.completed);
+        break;
+      default:
+        break;
+    }
+
+    return filteredTodos;
+  }, [filterType, todos]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
 
-const USER_ID = 0;
+const USER_ID = 10603;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState('all');
+  const [isHidden, setIsHidden] = useState(false);
+  const [query, setQuery] = useState('');
+  const [error, setError] = useState<null | string>(null);
```
d-wasilewski, line 16: You can just go with empty string initially. If the string will be empty it will have the same effect as null in terms of being truthy/falsy
```suggestion
  const [error, setError] = useState('');
```

src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
+import { client } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodoFilter } from './TodoFilter';
 
-const USER_ID = 0;
+const USER_ID = 10603;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState('all');
```
d-wasilewski, line 13: Let's create an enum for that so you won't have to use simple strings in the code
```suggestion
  const [filterType, setFilterType] = useState(FilterType.ALL);
```

src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
+      });
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let filteredTodos = todos;
+
+    switch (filterType) {
+      case 'all':
+        filteredTodos = todos;
+        break;
```
d-wasilewski, line 34: let's just make it a default case, we don't need to filter for this case

src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
+      .then(setTodos).catch(() => {
+        setError('Unable to load todos');
+        setTimeout(() => {
+          setError(null);
+        }, 3000);
+      });
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let filteredTodos = todos;
```
d-wasilewski, line 29:
```suggestion
    let filteredTodos;
```

src/App.tsx
```diff
@@ -15,141 +55,70 @@ export const App: React.FC = () => {
...
+        <button
+          type="button"
+          className="delete"
+          onClick={() => setIsHidden(true)}
+        />
+        {error && <p>{error}</p>}
         {/* show only one message at a time */}
-        Unable to add a todo
+
+        {/* {query && 'Unable to add a todo'}
```
d-wasilewski, line 116: let's remove the comments

src/TodoCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const TodoCard = ({ todo, key }: TodoCardProps) => {
+  const isCompleted = todo.completed === true;
+  const [isShown, setIsShown] = useState(false);
+
+  return (
+    <div
+      className={`todo ${isCompleted ? 'completed' : ''}`}
+      key={key}
```
d-wasilewski, line 16: You don't need a key here, you already put them on the list in TodoList

src/App.tsx
```diff
@@ -15,141 +55,70 @@ export const App: React.FC = () => {
...
+              filterType={filterType}
+              setFilterType={setFilterType}
+            />
+
+            {todos.find(todo => todo.completed) && (
+              <button type="button" className="todoapp__clear-completed">
+                Clear completed
+              </button>
+            )}
+          </footer>
```
d-wasilewski, line 97: let's create a separate component for that

src/App.tsx
```diff
@@ -15,141 +55,70 @@ export const App: React.FC = () => {
...
 
+      <div
+        className={`notification is-danger is-light has-text-weight-normal ${isHidden ? 'hidden' : ''}`}
+      >
+        <button
+          type="button"
+          className="delete"
+          onClick={() => setIsHidden(true)}
+        />
+        {error && <p>{error}</p>}
```
d-wasilewski, line 113: this can be a notification component either

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #569, opened 2023-06-04
solution

natalia-klonowska:
- [DEMO LINK](https://natalia-klonowska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,52 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { TodoList } from './components/TodoList';
+import { Notification } from './components/Notification';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,156 +1,52 @@
...
-        <button type="button" className="delete" />
-
-        {/* show only one message at a time */}
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      {hasError && (
```
d-wasilewski, line 47: You don't need `hasError` variable at all because you don't need to check for string's lenght if you want to check if it's truthy. You can just go with `{error && (` here

src/components/TodoList.tsx
```diff
@@ -0,0 +1,54 @@
...
+  todos: Todo[],
+  filter: string,
+}
+
+export const TodoList: React.FC<TodoListProps> = ({ todos, filter }) => {
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'All':
+        return true;
+      case 'Active':
```
d-wasilewski, line 13: Let's create an enum for all those values to avoid using simple strings

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #570, opened 2023-06-05
Solution

mateusz-michalowski:
[DEMO LINK](https://mateusz-michalowski.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,50 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodosList } from './TodosList';
+import { InputField } from './InputField';
+import { ErrorMessage } from './handleError';
+import { Footer } from './Footer';
 
-const USER_ID = 0;
+type TodoStatus = 'all' | 'active' | 'completed';
+
+const USER_ID = 10627;
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
 
-const USER_ID = 0;
+type TodoStatus = 'all' | 'active' | 'completed';
+
+const USER_ID = 10627;
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<TodoStatus>('all');
```
d-wasilewski, line 18: Create an enum for this instead, so that you will avoid using simple strings in the code
```suggestion
  const [status, setStatus] = useState(TodoStatus.ALL);
```

src/App.tsx
```diff
@@ -1,10 +1,50 @@
...
+  const [itemsLeft, setItemsLeft] = useState(0);
+
+  const [isError, setIsError] = useState(false);
```
d-wasilewski, line 21: You can set the error message to a string instead
```suggestion
  const [isError, setIsError] = useState('');
```

src/handleError.tsx
```diff
@@ -0,0 +1,23 @@
...
+          type="button"
+          className="delete"
+          onClick={() => handleSetError(false)}
+        />
+
+        Unable to add a todo
+        <br />
+        Unable to delete a todo
+        <br />
+        Unable to update a todo
```
d-wasilewski, line 20: Maybe just pass the error message here
```suggestion
        {errorMessage}
```

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #572, opened 2023-06-06
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,55 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Filter } from './components/Filter/Filter';
+import { NewTodo } from './components/NewTodo/NewTodo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Notification } from './components/Notification/Notification';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 10596;
 
 export const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


SuaBy514:
I have decided to start a pull request regardless of my problem in order to ask about it; why won't the title scale to the width? I have verified and every other css tag applies perfectly fine, I couldn't figure it out

src/App.tsx
```diff
@@ -1,156 +1,87 @@
...
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
-      </div>
+        <TodoList todos={currentTodos} />
+
+        {todos.length && (
+          <footer className="todoapp__footer">
```
d-wasilewski, line 68: you can create a separate component for footer

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  return (
+    <nav className="filter">
+      {filterTypes.map((filter) => {
+        let title: string | string[] = filter.split('');
+
+        title[0] = title[0].toUpperCase();
+
+        title = title.join('');
```
d-wasilewski, line 27: If the goal of this is to uppercase the first letter you can do it like this

`const title = filter.charAt(0).toUpperCase() + filter.slice(1);`

d-wasilewski (changes requested):

src/components/Filter/Filter.tsx
SuaBy514: It is, thank you

d-wasilewski (approved):

PR #582, opened 2023-06-08
solution

Nikoramo:
[DEMO LINK](https://Nikoramo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,83 @@
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header/Header';
+import { Main } from './components/Main/Main';
+import { Footer } from './components/Footer/Footer';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,156 +1,83 @@
...
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { FilterBy } from './types/FilterBy';
+import { getTodos } from './api/todos';
+
+const USER_ID = 10392;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState(false);
```
SanyaBratashchuk, line 18:
```suggestion
  const [hasError, setHasError] = useState(false);
```

src/App.tsx
```diff
@@ -1,156 +1,83 @@
...
+      switch (filterBy) {
+        case FilterBy.ACTIVE:
+          return !todo.completed;
+        case FilterBy.COMPLETED:
+          return todo.completed;
+        case FilterBy.ALL:
+          return true;
+        default:
+          return todo;
+      }
```
SanyaBratashchuk, line 37: move filter inside switch case to avoid extra loop when filterBy is "All"

SanyaBratashchuk (changes requested):

PR #586, opened 2023-06-13
Task24

Ryshon:
[DEMO LINK](https://Ryshon.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,72 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useMemo, useState } from 'react';
 
-const USER_ID = 0;
+import { TodoInfo } from './components/todoinfo/todoinfo';
+import { Header } from './components/header/header';
+import { Footer } from './components/footer/footer';
+import { Error } from './components/error/error';
+
+import { UserWarning } from './UserWarning';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,10 +1,72 @@
...
+        const todosData = response as Todo[];
+
+        setTodos(todosData);
+        setIsAddError(false);
+        setIsDeleteError(false);
+        setIsUpdateError(false);
+      } catch (error) {
+        setIsAddError(true);
+        setIsDeleteError(true);
+        setIsUpdateError(true);
```
agniya-i, line 38: You should set only one error in case of failed **loading**, not of other actions

src/App.tsx
```diff
@@ -1,10 +1,72 @@
...
-const USER_ID = 0;
+const USER_ID = 10683;
+const URL = `/todos?userId=${USER_ID}`;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<null | Todo[]>(null);
+  const [filterValue, setFilterValue] = useState(Filter.All);
+  const [isAddError, setIsAddError] = useState(false);
+  const [isDeleteError, setIsDeleteError] = useState(false);
+  const [isUpdateError, setIsUpdateError] = useState(false);
```
agniya-i, line 23: It would be better to create an enum for `ErrorType` and set the error type instead of using different flags

agniya-i (changes requested):

PR #604, opened 2023-07-01
complete first part

okosohor:
[DEMO LINK](https://okosohor.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,52 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { showError } from './helpers/helpers';
+import { Error } from './components/Error';
+import { Footer } from './components/Footer';
+import { Todos } from './components/Todos';
+import { Header } from './components/Header';
 
-const USER_ID = 0;
+const USER_ID = 10881;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => (showError('Failed to get list of todos', setError)));
+  }, []);
+
+  let visibleTodos = todos;
+
+  if (searchQuery) {
+    visibleTodos = visibleTodos
+      .filter(todo => (
+        todo.title
+          .toLowerCase()
+          .includes(searchQuery.toLowerCase())
+      ));
+  }
+
+  switch (filter) {
+    case 'completed':
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+
+    case 'active':
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+
+    default:
+      break;
+  }
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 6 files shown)


src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+        <a
+          href="#/completed"
+          className={cn('filter__link', {
+            selected: filter === 'completed',
+          })}
+          onClick={() => setFilter('completed')}
+        >
+          Completed
+        </a>
```
serentinos, line 46: I think, in this section will be better using Enums instead of just hardcoded values. You even can map Enum values in nav after that. Create some Enum and then use it:)

serentinos (approved):

src/App.tsx
```diff
@@ -1,10 +1,52 @@
...
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+
+    case 'active':
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+
+    default:
+      break;
+  }
```
dffUqp, line 48: to find a `visibleTodos` we can use only one filter.

what is a `searchQuery`? I think in this task we don't need to implement a search🤔

src/App.tsx
```diff
@@ -1,10 +1,52 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => (showError('Failed to get list of todos', setError)));
```
dffUqp, line 23: unnecessary helper

src/App.tsx
```diff
@@ -14,143 +56,20 @@ export const App: React.FC = () => {
...
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
-      </div>
+        <Todos todos={visibleTodos} />
```
dffUqp, line 65: hide the list if there are no todos yet

src/components/Header.tsx
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      {/* Add a todo on form submit */}
+      <form>
+        <input
+          type="text"
+          className="todoapp__new-todo"
+          placeholder="What needs to be done?"
+          value={searchQuery}
+          onChange={(event) => (setSearchQuery(event.target.value))}
```
dffUqp, line 38: let's create separate handler for this

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/components/Footer.tsx
```diff
@@ -0,0 +1,54 @@
+import cn from 'classnames';
+import { Todo } from '../types/Todo';
+
+interface Props {
+  filter: string,
+  todos: Todo[]
+  setFilter: (text: string) => void;
+}
+
+export const Footer: React.FC<Props> = ({ filter, todos, setFilter }) => {
```
niaYaro, line 10: Good enough!

Take a look that, according to [Demo Page](https://mate-academy.github.io/react_todo-app-with-api/), the quantity of the items left should stay unchanged regardless of the filter we use.


230703-review-1
230703-review-2
230703-review-3

niaYaro (approved):

src/App.tsx
okosohor: I will use it for next tasks. I can write SetErorr and and setTimeout for every error but i think will be beter  use this function, because we have similar cases, just different text.
image

silchyn (approved):
The next part of this task will be reviewed

PR #607, opened 2023-07-01
solution

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,102 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+// import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Header } from './components/Header';
+import { ErrorType, FilterType } from './types/HelperTypes';
+import { ErrorMessage } from './components/ErrorMessage';
+import { getFilteredTodos, getTodosInfo } from './Helper';
+import { Footer } from './components/Footer';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 10884;
+
+// I did this just in case https://mate-academy.github.io/react_todo-app-with-api/ doesn't respond.
+const fakeTodos: Todo[] = [
+  {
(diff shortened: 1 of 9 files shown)


src/components/Todo.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { Todo as TodoType } from '../types/Todo';
```
bohdan-mykhailenko, line 3: Using an alias, the same name of component and type is a little confusing to understand

bohdan-mykhailenko (approved):
Good job 👍 
Here are some points to make the code logic and style more clear

src/App.tsx
```diff
@@ -1,10 +1,136 @@
...
+
+  const removeError = () => {
+    setErrorType(null);
+  };
+
+  const errorMessage = useMemo(() => {
+    const timerId = setTimeout(() => {
+      setErrorType(null);
+      clearTimeout(timerId);
+    }, 3000);
```
dinaminahi, line 118: Does this work? As the timer created by setTimeout will fire the callback function after 3000 milliseconds, and before that, you're trying to clear the timer using clearTimeout(timerId), which won't work as timerId will be null at that point.
Instead we should clear timeout on `unmount`

dinaminahi (changes requested):
Good job 👍 
Let's do some fixes

src/App.tsx
```diff
@@ -1,10 +1,117 @@
...
+  userId: USER_ID,
+  completed: false,
+  title: '',
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterType, setFilterType] = useState<FilterType>(FilterType.ALL);
+  const [errorType, setErrorType] = useState<ErrorType | null>(null);
+  const [todosInfo, setTodosInfo] = useState<TodosInfo>(initialTodosInfo);
```
dinaminahi, line 36: Do we need to have this in state? Or we can just calculate it when we need?

dinaminahi (changes requested):
Well done 👍 
Let's just think about todosInfo variable. I don't think we need to keep it in state as it's values can be calculated based on state values

silchyn (approved):
The next part of this task will be reviewed

PR #645, opened 2023-07-13
Develop

sasdaiv:
[DEMO LINK](https://sasdaiv.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,62 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
+
+import { getTodos } from './api/todos';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+const USER_ID = 11073;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [filterStatus, setFilterStatus] = useState(Status.ALL);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+      })
+      .catch((error: Error) => {
+        setHasError(true);
+        throw new Error(error.message);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (hasError) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,10 +1,62 @@
...
+        setHasError(false);
+      }, 3000);
+
+      return () => clearTimeout(timeout);
+    }
+
+    return () => {};
+  }, [hasError]);
+
+  const completedTodos = todos?.filter((todo) => todo.completed);
```
Viktor-Kost, line 41: You are filtering todos twice to get completed and uncompleted todos. It's better to filter todos once and use array partitioning to separate completed and uncompleted todos.

src/App.tsx
```diff
@@ -29,127 +81,100 @@ export const App: React.FC = () => {
...
+            <div
+              key={todo.id}
+              className={cn('todo', {
+                completed: todo?.completed,
+              })}
+            >
+              <label className="todo__status-label">
+                <input
+                  type="checkbox"
+                  className="todo__status"
```
Viktor-Kost, line 94: You are using 'defaultChecked' which is uncontrolled component. It's better to use 'checked' property and handle 'onChange' event to control component.

src/App.tsx
```diff
@@ -29,127 +81,100 @@ export const App: React.FC = () => {
...
+              >
+                Completed
+              </a>
+            </nav>
+
+            <button
+              type="button"
+              className="todoapp__clear-completed"
+              style={{ opacity: completedTodos.length > 0 ? '1' : '0' }}
+              disabled={!completedTodos}
```
Viktor-Kost, line 153: You are using '!completedTodos' for disabling the button. It should be '!completedTodos?.length' as completedTodos is an array and if it's empty, the button should be disabled.

src/App.tsx
```diff
@@ -1,10 +1,62 @@
...
+  const [filterStatus, setFilterStatus] = useState(Status.ALL);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+      })
+      .catch((error: Error) => {
+        setHasError(true);
+        throw new Error(error.message);
```
Viktor-Kost, line 25: You are throwing error in catch block. This will result in the error being unhandled, leading to application crash. It's better to log the error message instead of throwing it.

Viktor-Kost (changes requested):
Good job!
Some moments can be fixed

Viktor-Kost (approved):
🔥

PR #647, opened 2023-07-13
Solution

Andrij-Nalyvayko:
[DEMO LINK](https://Andrij-Nalyvayko.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,156 +1,70 @@
...
+const USER_ID = 11076;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [showError, setShowError] = useState(false);
+
+  const errorMessage = useRef<ErrorMessages>(ErrorMessages.noError);
```
etojeDenys, line 19: We can combine it into a single errorMessage state using useState, storing the error message, and checking if ErrorMessages.noError means that the error doesn't exist.

etojeDenys (changes requested):

src/App.tsx
Andrij-Nalyvayko: When we add `errorMessage` to state, we'll have changing error message in component `ErrorNotification.tsx` immediately, so we will have problem with animation. 

https://github.com/mate-academy/react_todo-app-loading-todos/assets/83166880/98adcf43-ce37-4b03-84c6-a8100864521b

Andrij-Nalyvayko: Or I can just add state with` {message: string, showMessage: boolean }` but it add new rerender

VitaliyBondarenko1982 (approved):

PR #654, opened 2023-07-18
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,156 +1,92 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Todo } from './types/Todo';
+import { createTodo, getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { Notification } from './components/Notification/Notification';
 
-const USER_ID = 0;
+const USER_ID = 11085;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        const fetchedTodos = await getTodos(USER_ID);
+
+        setTodos(fetchedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to fetch todos');
+      }
+    };
+
+    fetchTodos();
+  }, []);
+
+  const addTodo = async (title: string) => {
+    try {
+      const newTodo = await createTodo({
+        userId: USER_ID,
+        title,
+        completed: false,
+      });
+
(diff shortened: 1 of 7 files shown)


loralevitska (changes requested):
Almost done) Let's improve

src/App.tsx
```diff
@@ -1,156 +1,89 @@
...
+import { createTodo, getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { Notification } from './components/Notification/Notification';
 
-const USER_ID = 0;
+const USER_ID = 11085;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
TarasHoliuk, line 15:
```suggestion
  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
you write `'all' | 'active' | 'completed'` a lot of times - create type for it and export everywhere you need

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const getFilteredTodosCount = () => {
+    switch (filter) {
+      case 'all':
+        return todosCount;
+      case 'active':
+        return todosCount - completedCount;
+      case 'completed':
+        return completedCount;
+      default:
+        return completedCount;
```
TarasHoliuk, line 30:
```suggestion
      case 'completed':
      default:
        return completedCount;
```
jou may join a few cases in this way

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,82 @@
...
+          className={classNames('filter__link', {
+            selected: filter === 'all',
+          })}
+          onClick={() => handleFilterChange('completed')}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {completedCount > 0 && (
```
TarasHoliuk, line 72: let's add a class with `visibility: hidden` style instead of conditional rendering to keep the filter in the center of the footer: 
image

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const handleClose = () => {
+    setShowNotification(false);
+    onClose();
+  };
+
+  return (
+    <div
+      className={`notification is-danger is-light has-text-weight-normal ${showNotification
+        ? ''
+        : 'hidden'}`}
```
TarasHoliuk, line 38: it's better to use classnames library for such cases

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -1,156 +1,90 @@
...
+
+  const fetchTodos = async () => {
+    try {
+      const fetchedTodos = await getTodos(USER_ID);
+
+      setTodos(fetchedTodos);
+    } catch (error) {
+      setErrorMessage('Unable to fetch todos.');
+    }
+  };
```
DarkMistyRoom, line 27: you can declare this function inside the `useEffect` to avoid redeclaring on each render

src/components/Notification/Notification.tsx
DarkMistyRoom: not fixed. feel free to write in the chat if you have questions
image

PR #657, opened 2023-07-20
Develop

DanyloPasko:
[DEMO LINK](https://DanyloPasko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import { Todo } from './types/Todo';
 
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 11073;
+const USER_ID = 10;
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,10 +1,79 @@
...
+  };
+
+  const [completedTodos, uncompletedTodos] = useMemo(() => {
+    const completed = todos?.filter((todo) => todo.completed);
+    const uncompleted = todos?.filter((todo) => !todo.completed);
+
+    return [completed, uncompleted];
+  }, [todos]);
+
+  const todoIsActive = todos?.find((todo) => todo.completed === false);
```
vlad-demchuk, line 61: you can check length of uncompleted array

src/App.tsx
```diff
@@ -29,127 +97,99 @@ export const App: React.FC = () => {
...
+                onClick={() => setFilterStatus(Status.COMPLETED)}
+              >
+                Completed
+              </a>
+            </nav>
+
+            <button
+              type="button"
+              className="todoapp__clear-completed"
+              style={{ opacity: completedTodos.length > 0 ? '1' : '0' }}
```
vlad-demchuk, line 168: it will be better to use another approach to hide button, since it is still clickable using opacity

vlad-demchuk (changes requested):

volodymyr-soltys97 (approved):

PR #658, opened 2023-07-20
solution

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/react_todo-app-loading-todos/


src/App.tsx
```diff
@@ -1,10 +1,27 @@
...
+const USER_ID = 11096;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState(Filter.All);
+  const [hasError, setHasError] = useState(Error.Not);
+
+  useEffect(() => {
+    getTodos(USER_ID).then(data => {
+      setTodos(data);
```
GGsmou, line 21: It might be needed to add .cath later to detect some errors

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,107 @@
...
+}) => {
+  const countItems = todos.filter(todo => !todo.completed).length;
+
+  const filterAllHandler = (e: React.MouseEvent<HTMLAnchorElement>) => {
+    e.preventDefault();
+
+    setFilter(Filter.All);
+  };
+
+  const filterActiveHandler = (e: React.MouseEvent<HTMLAnchorElement>) => {
```
GGsmou, line 29: It is not really needed to move each filter to separate function
If you want separate func you can pass argument instead

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    setTodos(todos.filter(t => !t.completed));
+  };
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {countItems}
+        {' '}
+        items left
```
GGsmou, line 61: If items count === 1 text must be `item` not itmes
In future tasks fix this issue

src/types/Todo.ts
```diff
@@ -4,3 +4,16 @@ export interface Todo {
...
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+export enum Error {
+  Not = '',
+  Add = 'add',
+  Delete = 'delete',
+  Update = 'update',
```
GGsmou, line 18: There also might be case when data just not fetched from server

GGsmou (approved):
GJ, approving!
Check some problems to fix in future tasks

And678 (approved):
:heavy_check_mark:

PR #659, opened 2023-07-21
solution

vpdrabynko:
[DEMO LINK](https://vpdrabynko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,15 +26,15 @@ export const App: React.FC = () => {
   }, []);
 
   useEffect(() => {
+    let timeout: NodeJS.Timeout;
+
     if (hasError) {
-      const timeout = setTimeout(() => {
+      timeout = setTimeout(() => {
         setHasError(false);
       }, 3000);
-
-      return () => clearTimeout(timeout);
     }
 
-    return () => { };
+    return () => clearTimeout(timeout);
   }, [hasError]);
 
   const completedTodos = useMemo(() => {


src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+  const handleCheckboxChange = (todoId: number) => {
+    const updatedTodos = todos.map((todo) => {
+      if (todo.id === todoId) {
+        return { ...todo, completed: !todo.completed };
+      }
+
+      return todo;
+    });
+
+    setTodos(updatedTodos);
```
Anastasiia-Svintsova, line 48: It's better to use functional setState to ensure that the state update is based on the latest state value

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+  };
+
+  const [completedTodos, uncompletedTodos] = useMemo(() => {
```
Anastasiia-Svintsova, line 51: In my opinion, it's better to create 2 separate variables to memoize the results for each calculation independently

src/App.tsx
```diff
@@ -29,127 +95,101 @@ export const App: React.FC = () => {
...
-            </label>
-
-            <span className="todo__title">Todo is being saved now</span>
-            <button type="button" className="todo__remove">×</button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
+          {visibleTodos?.map((todo: Todo) => (
```
Anastasiia-Svintsova, line 98: Create a separate component for the TodoList and don't render it if there is no visibleTodos

src/App.tsx
```diff
@@ -29,127 +95,101 @@ export const App: React.FC = () => {
...
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        {todos?.length > 0 && (
+          <footer className="todoapp__footer">
```
Anastasiia-Svintsova, line 127: Would be great to create a separate component for the footer as well. To declutter the App component and make it more clear and readable.

Anastasiia-Svintsova (changes requested):
Keep up the great job! Check my comments, please

src/App.tsx
```diff
@@ -1,10 +1,76 @@
...
+  useEffect(() => {
+    if (hasError) {
+      const timeout = setTimeout(() => {
+        setHasError(false);
+      }, 3000);
+
+      return () => clearTimeout(timeout);
+    }
+
+    return () => { };
```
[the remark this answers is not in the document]
vpdrabynko: how to fix the problem 'Not all code paths return a value.'  then?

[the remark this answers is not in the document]
witflash: Your return on the 37 line haven't any sense. It looks like a hack.
It's better to do the next way:
```
useEffect(() => {
  let timeout;

  if (hasError) {
    timeout = setTimeout(// your callback);
  }

  return () => clearTimeout(timeout);
}, [hasError]);
```

So if you have an error timeout will be cleared. Or if you don't have an error, clearTimeout just do nothing, it's ok.

witflash (changes requested):
Good job!

Pay attention to my comment to fix your `useEffect` usage

src/App.tsx
```diff
@@ -1,10 +1,70 @@
...
+const USER_ID = 11098;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [filterStatus, setFilterStatus] = useState(Status.ALL);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((todosfromserver) => {
```
agniya-i, line 20:
```suggestion
      .then((todosFromServer) => {
```
Don't forget about camelCase)

agniya-i (approved):

PR #661, opened 2023-07-22
Develop

VolodymyrPst:
https://volodymyrpst.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,87 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import * as todosService from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 11105;
 
 export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,10 +1,87 @@
...
+  const addTodo = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const newTodo = {
+      userId: USER_ID,
+      title: newTitle.trim(),
+      completed: false,
+    };
+
+    todosService.createTodo(newTodo).then(getAllTodo);
```
VitaliyBondarenko1982, line 32: after add new todo no need to get allTodo from server. just update todos state in case success response from server 
something like that
```
todosService.createTodo(newTodo).then(response => setTodos(prevTodos => [ ...prevTodos, response ]))
```
but anyway add/remove todo should be implemented in next task.

Screenshot from 2023-07-22 16-49-00

VitaliyBondarenko1982 (changes requested):
Check [working example](https://mate-academy.github.io/react_todo-app-with-api/#/completed)
when selected completed filter dissappears footer
also do not render errors when no any errors
chrome-capture-2023-6-22

nazarmatsevych (approved):

PR #665, opened 2023-07-25
solution

MariiaMonakhova:
[DEMO LINK](https://MariiaMonakhova.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useContext, useMemo, useState } from 'react';
 import cn from 'classnames';
@@ -20,11 +21,13 @@ export const App: React.FC = () => {
 
   const { todos, addTodo, resetError } = todosContext;
 
-  const activeTodosCount = todos.filter((todo: Todo) => !todo.completed).length;
+  const activeTodosCount = useMemo(() => {
+    return todos.filter((todo: Todo) => !todo.completed).length;
+  }, [todos]);
 
-  const completedTodosCount = todos.filter(
-    (todo: Todo) => todo.completed,
-  ).length;
+  const completedTodosCount = useMemo(() => {
+    return todos.filter((todo: Todo) => todo.completed).length;
+  }, [todos]);
 
   function createTodo(event: React.FormEvent<HTMLFormElement>) {
     event.preventDefault();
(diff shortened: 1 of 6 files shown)


src/components/TodoContext.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    return client
+      .post<Todo>('/todos', newTodo)
+      .then((createdTodo) => {
+        setTodos(prevTodos => [createdTodo, ...prevTodos]);
+        setError('');
+      })
+      .catch(() => {
+        setError(ErrorType.addTodo);
+      });
```
artkas03, line 50: instead of importing client and working directly with this object you could create new methods in api/todos.ts, import those methods and call them

src/components/TodoErrors.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { error, resetError } = todoContext;
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    if (error) {
+      const timeoutId = setTimeout(() => {
+        resetError();
+      }, 3000);
+
+      return () => clearTimeout(timeoutId);
```
artkas03, line 21: it's not really necessary to call `clearTimeout` after it has been finished

artkas03 (approved):
GJ, but check some tips below

src/App.tsx
```diff
@@ -1,156 +1,101 @@
...
+
+  const todosContext = useContext(TodosContext);
+
+  if (!todosContext) {
+    return null;
+  }
+
+  const { todos, addTodo, resetError } = todosContext;
+
+  const activeTodosCount = todos.filter((todo: Todo) => !todo.completed).length;
```
anstsot, line 23: better to use `useMemo` hook

src/App.tsx
```diff
@@ -1,156 +1,101 @@
...
+
+  const completedTodosCount = todos.filter(
```
anstsot, line 25: better to use `useMemo` hook

src/components/TodoContext.tsx
anstsot: agree with [comment](https://github.com/mate-academy/react_todo-app-loading-todos/pull/665/files#r1273848615)

anstsot (changes requested):
Good work, let's improve a little and fix one bug.
The filters should always be centered
image

anstsot (approved):
Good work, approving:sparkles:

PR #667, opened 2023-07-25
app_load_todos

hrynkevych:
[DEMO LINK](https://hrynkevych.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,8 +19,8 @@ export const App: React.FC = () => {
   const [filter, setFilter] = useState(FilterType.All);
   const [errorMessage, setErrorMessage] = useState<string>('');
   const [value, setValue] = useState('');
-  const completedTodos = todos.filter(todo => todo.completed);
-  const notCompletedTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter((todo) => todo.completed);
+  const notCompletedTodos = todos.filter((todo) => !todo.completed);
 
   useEffect(() => {
     async function fetchTodos() {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,10 +1,83 @@
...
+  const filteredTodos = todos.filter((todo) => {
+    switch (filter) {
+      case FilterType.Active:
+        return !todo.completed;
+      case FilterType.Completed:
+        return todo.completed;
+      default:
+        return todo;
+    }
+  });
```
IvanFesenko, line 79: No mistake on this task, but it future do not filter in case FilterType is All
Also filter callback should return boolean - you have object instead

src/App.tsx
```diff
@@ -15,141 +88,78 @@ export const App: React.FC = () => {
...
+          )}
 
-          {/* Add a todo on form submit */}
           <form>
             <input
               type="text"
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              value={value}
+              onChange={(event) => setValue(event.target.value)}
```
IvanFesenko, line 109: Recomendation: 
Create separate handler inctead of using anonymous function

src/App.tsx
```diff
@@ -15,141 +88,78 @@ export const App: React.FC = () => {
...
+                </button>
+              ) : (
+                <button
+                  style={{ visibility: 'hidden' }}
+                  type="button"
+                  className="todoapp__clear-completed"
+                >
+                  Clear completed
+                </button>
+              )}
```
IvanFesenko, line 143: No need to render button and hide it
```suggestion
              {completedTodos.length > 0 ? (
                <button
                  type="button"
                  className="todoapp__clear-completed"
                >
                  Clear completed
                </button>
              ) : null}
```

IvanFesenko (changes requested):
Good job. Check my comments

src/App.tsx
hrynkevych: Got it, thank you!

VitaliyBondarenko1982 (approved):

PR #674, opened 2023-07-27
add task solution

NazarBaraban:
https://nazarbaraban.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,58 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useMemo, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { Error } from './components/Error';
+import { TodoFilter } from './components/TodoFilter';
+import { TodoItem } from './components/TodoItem';
+import { getTodos } from './api/todos';
+import { Todo, FilterStatus } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 11220;
 
 export const App: React.FC = () => {
+  const [selectedStatus, setSelectedStatus] = useState(FilterStatus.All);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todosContent, setTodosContent] = useState<Todo[]>([]);
+
+  const todos = useMemo(() => {
+    return todosContent.filter(todo => {
+      switch (selectedStatus) {
+        case FilterStatus.Completed:
+          return todo.completed;
+
+        case FilterStatus.Active:
+          return !todo.completed;
+
+        default:
+          return todo;
+      }
+    });
+  }, [selectedStatus, todosContent]);
+
+  const completeTodo = useMemo(() => {
+    return todosContent.filter(todo => todo.completed);
+  }, [todosContent]);
+
+  const closeError = () => {
+    setErrorMessage('');
+  };
+
+  const filterOnStatus = (status: FilterStatus) => {
+    setSelectedStatus(status);
+  };
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(response => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
 
-const USER_ID = 0;
+const USER_ID = 11220;
 
 export const App: React.FC = () => {
+  const [selectedStatus, setSelectedStatus] = useState(FilterStatus.All);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [todosContent, setTodosContent] = useState<Todo[]>([]);
+
+  const todos = useMemo(() => {
```
olena-ole, line 17: Consider naming your variables todos and filteredTodos

src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
+        case FilterStatus.Active:
+          return !todo.completed;
+
+        default:
+          return todo;
+      }
+    });
+  }, [selectedStatus, todosContent]);
+
+  const completeTodo = useMemo(() => {
```
olena-ole, line 32: this variable name seems suitable for a function, maybe it's better to call it completedTodos

src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
+    getTodos(USER_ID)
+      .then(response => {
+        setTodosContent(response);
+      })
+      .catch(err => {
+        setErrorMessage(err);
+      });
+  }, []);
+
+  const amount = todosContent.length - completeTodo.length;
```
olena-ole, line 54: it is not clear what amount you mean, activeTodosCount would sound better

src/components/Error.tsx
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }, [message]);
+
+  return (
+    <div className={
+      classNames('notification is-danger is-light has-text-weight-normal',
+        {
+          hidden: message.length !== 0,
+        })
+    }
```
olena-ole, line 23:
```suggestion
      classNames('notification is-danger is-light has-text-weight-normal',
        { hidden: message.length !== 0 })
      }
```

olena-ole (approved):
Good Job! Your app works as expected and the logic is well-built, + you used enum, which is excellent! But consider renaming your variables, cause they seem misleading. Also, the memoization doesn't seem to make sense here

src/App.tsx
anstsot: agree with [comment](https://github.com/mate-academy/react_todo-app-loading-todos/pull/674/files#r1276541678)

src/App.tsx
```diff
@@ -29,128 +80,27 @@ export const App: React.FC = () => {
...
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      {errorMessage && (
+        <Error
+          message={errorMessage}
+          whenClose={closeError}
```
anstsot, line 101:
```suggestion
          onClose={closeError}
```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,63 @@
...
+  selected,
+  filter,
+  todosTotalAmount,
+  completedTodosAmount,
+}) => {
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {todosTotalAmount}
+        items left
```
anstsot, line 22:
```suggestion
        {`${todosTotalAmount} items left`}
```

anstsot (changes requested):
Good work, let's improve a little and fix one bug.
The error isn't shown
image

anstsot (changes requested):
Good job, but the error is still not shown
image

YegorKorenkov (approved):
Good job!

PR #704, opened 2023-07-28
Develop

AlyonaSarapina:
[DEMO LINK](https://alyonasarapina.github.io/react_todo-app-loading-todos/)


anstsot (changes requested):
Good work, let's fix one bug.
The error doesn't disappear in 3 seconds
image

anstsot (approved):
Good work, approving:sparkles:

PR #722, opened 2023-07-31
add task solution

diamat1997:
[DEMO LINK](https://diamat1997.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoApp.tsx
+++ b/src/components/TodoApp.tsx
@@ -1,0 +1,125 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useState, useContext, useEffect } from 'react';
+import classNames from 'classnames';
+import { FilterStatus } from '../types/FilterStatus';
+import { TodosContext } from '../context/TodosContext';
+import { TodoList } from './TodoList';
+import { TodosFilter } from './TodosFilter';
+import { wait } from '../utils/wait';
+
+export const TodoApp: React.FC = () => {
+  // const [todoTitle, setTodoTitle] = useState('');
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.All);
+
+  /* Unable to add a todo
+    <br />
+    Unable to delete a todo
+    <br />
+    Unable to update a todo
+
+    hide err on request
+  */
+
+  const [todosPrep, setTodos, errorMsg, setErrorMsg] = useContext(TodosContext);
+
+  // left for future tasks so lint doesn't complain
+  if (todosPrep.length === -1) {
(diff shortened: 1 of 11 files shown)


src/components/TodoApp.tsx
```diff
@@ -0,0 +1,125 @@
...
+    <br />
+    Unable to update a todo
+
+    hide err on request
+  */
+
+  const [todosPrep, setTodos, errorMsg, setErrorMsg] = useContext(TodosContext);
+
+  // left for future tasks so lint doesn't complain
+  if (todosPrep.length === -1) {
```
loralevitska, line 26: array length cannot be -1
https://stackoverflow.com/questions/32386253/can-the-js-array-length-property-ever-return-a-negative-value

src/components/TodoApp.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  const todos = todosPrep.filter((todo) => {
+    switch (filterStatus) {
+      case FilterStatus.Active:
+        return !todo.completed;
+      case FilterStatus.Completed:
+        return todo.completed;
+      default:
+        return true;
```
loralevitska, line 37:
```suggestion
    switch (filterStatus) {
      case FilterStatus.Active:
        return !todo.completed;
        
      case FilterStatus.Completed:
        return todo.completed;
        
      default:
        return true;
```

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  // const [todos, setTodos] = useContext(TodosContext);
+  // add setter
+  const [tempTitle] = useState(todo.title);
```
loralevitska, line 20: Just use `todo.title` where you need and all

src/components/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <>
+      {todos.map((todo) => (
+        <TodoItem
+          key={todo.id}
+          todo={todo}
+          selectedTodo={selectedTodo}
+          setSelectedTodo={setSelectedTodo}
+        />
+      ))}
+    </>
```
loralevitska, line 24: Use semantic tags for list `ul` `li`. Don't forget to remove default `li` dots and `ul` paddings

loralevitska (changes requested):
Almost done) Let's improve

PR #725, opened 2023-08-01
Develop

allagutsul27:
[DEMO LINK](https://allagutsul27.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,36 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { TodoList } from './components/TodoList/TodoList';
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+import { prepareTodos } from './utils/filterTodos';
 import { UserWarning } from './UserWarning';
+import { TodoError } from './components/TodoError/TodoError';
+import { Error } from './types/Error';
 
-const USER_ID = 0;
+const USER_ID = 6909;
 
 export const App: React.FC = () => {
+  const [todoTitle, setTodoTitle] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortField, setSortField] = useState(Status.all);
+  const [error, setError] = useState(Error.NONE);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setError(Error.LOAD));
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const completedTodos = todos.filter(todo => todo.completed);
+  const notCompletedTodos = todos.filter(todo => !todo.completed);
+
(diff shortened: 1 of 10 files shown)


src/utils/filterTodos.ts
```diff
@@ -0,0 +1,17 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export function prepareTodos(allTodos: Todo[], filterBy: Status) {
+  const filteredTodos = [...allTodos];
```
polosanya, line 5: Filter is a non-mutating method. There is no need to create a copy

src/App.tsx
```diff
@@ -1,156 +1,87 @@
...
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        {todos.length > 0
+          && (
```
polosanya, line 62: There is a repeated check for 'todos.length > 0'. It would be more efficient to check this condition once and return a fragment with both TodoList and Footer if true.

src/types/Status.ts
```diff
@@ -0,0 +1,5 @@
+export enum Status {
+  all,
+  active,
+  completed,
```
polosanya, line 4:
```suggestion
export enum Status {
  All,
  Active,
  Completed,
```
Enum properties should be capitalized
https://www.typescriptlang.org/docs/handbook/enums.html

polosanya (changes requested):
Almost done! Let's improve your code a bit)

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,38 @@
+import { Status } from '../../types/Status';
+import { Todo } from '../../types/Todo';
+import { TodoFilter } from '../TodoFilter/TodoFilter';
+
+type Props = {
+  completedTodos: Todo[],
+  notCompletedTodos: Todo[],
```
etojeDenys, line 7: footer does not need to have the arrays of todos, let's pass only the length here
```suggestion
  nrOfCompletedTodos:  number,
  nrOfNotCompletedTodos: number,
```

etojeDenys (changes requested):
one nit

src/App.tsx
```diff
@@ -1,156 +1,88 @@
...
-              Completed
-            </a>
-          </nav>
-
-          {/* don't show this button if there are no completed todos */}
-          <button type="button" className="todoapp__clear-completed">
-            Clear completed
-          </button>
-        </footer>
+        {todosExist
```
anastasiia-tilikina, line 62: I believe this is not what was required to fix)
Since the condition for rendering `TodoList` and `Footer` is the same, we don't need to check it separately for both of them. Let's just wrap `Todolist` and `Footer`  in `Fragment` and render this fragment only if `todosExist` 

If it is still not clear for you feel free to ask your questions in slack chat!

anastasiia-tilikina (changes requested):
Great Job!
Let's just fix one more thing. 
Check the comment)

VitaliyBondarenko1982 (approved):
Well done

PR #726, opened 2023-08-02
react_todo-app-loading-todos

vintwp:
https://vintwp.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,55 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { TodoHeader } from './components/TodoHeader';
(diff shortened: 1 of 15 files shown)


vintwp:
> Do not load todos, need to check it image

I've fixed url, and it will works

volodymyr-soltys97 (changes requested):

src/App.tsx
```diff
@@ -1,156 +1,84 @@
...
+  const todosToRender = todos.filter(todo => {
+    switch (filterBy) {
+      case FilterBy.active:
+        return !todo.completed;
+      case FilterBy.completed:
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
```
IvanFesenko, line 35: NOTE: for next tasks - no need to filter in case FilterBy is all

IvanFesenko (approved):
Check my comment to avoid mistake in the future

PR #727, opened 2023-08-02
solution

LiliiaDanylenko:
[DEMO LINK](https://LiliiaDanylenko.github.io/react_todo-app-loading-todos/)

Не впевнена в useEffect для повідомлення про помилку

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,62 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Filter } from './components/Filter';
+import { NewTodo } from './components/NewTodo';
+import { FILTERS } from './types/FILTERS';
 
-const USER_ID = 0;
+const USER_ID = 11253;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [filter, setFilter] = useState(FILTERS.ALL);
+  const visibleTodos = useMemo(() => {
+    switch (filter) {
+      case FILTERS.ALL:
+        return todos;
+
+      case FILTERS.ACTIVE:
+        return todos.filter(todo => !todo.completed);
+
+      case FILTERS.COMPLITED:
+        return todos.filter(todo => todo.completed);
+
+      default:
+        return todos;
+    }
+  }, [todos, filter]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch();
+  }, []);
+
+  useEffect(() => {
+    if (errorMessage !== '') {
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
+
+  const allTodoComplited = useMemo(() => {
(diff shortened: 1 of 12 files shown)


src/types/FILTERS.ts
```diff
@@ -0,0 +1,5 @@
+export enum FILTERS {
+  ALL = 'All',
+  ACTIVE = 'Active',
+  COMPLITED = 'Completed',
```
volodymyr-soltys97, line 4: There's a spelling mistake in the 'COMPLETED' filter. It's currently spelled as 'COMPLITED'.

src/api/todos.ts
```diff
@@ -5,4 +5,6 @@ export const getTodos = (userId: number) => {
   return client.get<Todo[]>(`/todos?userId=${userId}`);
 };
 
-// Add more methods here
+export const createTodo = ({ userId, title, completed }: Omit<Todo, 'id'>) => {
+  return client.post<Todo>('/todos', { userId, title, completed });
```
volodymyr-soltys97, line 9: You should handle errors for the API request. Using try/catch blocks can help you gracefully manage when the API call fails.

src/App.tsx
```diff
@@ -1,10 +1,62 @@
...
+
+      default:
+        return todos;
+    }
+  }, [todos, filter]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch();
```
volodymyr-soltys97, line 37: Empty catch block. It's important to handle errors in some way to prevent potential crashes or unpredictable behavior

src/App.tsx
```diff
@@ -1,10 +1,62 @@
...
+  }, []);
+
+  useEffect(() => {
```
volodymyr-soltys97, line 40: Using setTimeout in useEffect could lead to memory leaks if the component unmounts before the timeout finishes. Consider using clearTimeout in the cleanup function of useEffect.

src/components/NewTodo/NewTodo.tsx
```diff
@@ -0,0 +1,51 @@
...
+    event.preventDefault();
+
+    if (query.trim() === '') {
+      setQuery('');
+
+      return;
+    }
+
+    createTodo({
+      userId: 11253,
```
volodymyr-soltys97, line 25: Hardcoding 'userId' is not a good practice. It should be dynamic or taken from the user's session or state.

src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,49 @@
...
+        onClick={() => setFilter(FILTERS.ACTIVE)}
+      >
+        Active
+      </a>
+
+      <a
+        href="#/completed"
+        className={classNames(
+          'filter__link',
+          { selected: filter === FILTERS.COMPLITED },
```
volodymyr-soltys97, line 41: There is a typo in the filter name. It should be 'FILTERS.COMPLETED' instead of 'FILTERS.COMPLITED'.

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,60 @@
+import { useState } from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoItem: React.FC<Props> = ({ todo }) => {
+  const [edited] = useState(false);
```
volodymyr-soltys97, line 10: You set the state 'edited' to false and never change it. That makes the edit form in your JSX code unreachable. If you want to allow users to edit the todo, you should create a function that will change 'edited' state when necessary.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

polosanya (approved):
👍

PR #733, opened 2023-08-06
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/react_todo-app-loading-todos/)
 
Question: TodoError.tsx, line 27. When load error occurs and disappears, the error message goes without animation, how am i suppused to fix that issue? When in TodoError error changes to Error.None after 3s, no rendering occurs, also there is a check in the parent element, 67 line App.tsx.
Is that why it doesn't work?

PS. i made mistake in 2 line of fetchClient on purpose just to show you the problem

### DIFF
--- a/src/components/TodoError.tsx
+++ b/src/components/TodoError.tsx
@@ -1,0 +1,38 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { useEffect, useRef } from 'react';
+import classNames from 'classnames';
+import { Error } from '../types/Error';
+
+type Props = {
+  error: Error;
+  onErrorChange: (error: Error) => void;
+};
+
+export const TodoError: React.FC<Props> = ({ error, onErrorChange }) => {
+  const timeRef = useRef(0);
+
+  useEffect(() => {
+    window.clearTimeout(timeRef.current);
+
+    if (error) {
+      timeRef.current = window.setTimeout(() => {
+        onErrorChange(Error.None);
+      }, 3000);
+    }
+  }, [error]);
+
+  return (
+    <div className={classNames('notification',
+      'is-danger', 'is-light', 'has-text-weight-normal', {
+        hidden: error === Error.None,
+      })}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => onErrorChange(Error.None)}
+      />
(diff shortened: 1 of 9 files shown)


maxim2310:
yes when errorMassage is appeared your todoError also suddenly appear on screen,  you need to add style to todoError that make it opacity 0 when it appear and increase smooth or always render this component on screen with opacity and change it by the condition.
And next time will be better to ask your question in our chat.

witflash (changes requested):
Just for re-request enabling

src/components/TodoError.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  useEffect(() => {
+    window.clearTimeout(timeRef.current);
+
+    if (error) {
+      timeRef.current = window.setTimeout(() => {
+        onErrorChange(Error.None);
+      }, 3000);
+    }
+  }, [error]);
```
lerastarynets, line 22:
```suggestion
  useEffect(() => {
    if (error) {
      setTimeout(() => onErrorChange(Error.None), 3000);
    }
  }, [error]);
```
you can simplify it

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <footer className="todoapp__footer">
+      <span className="todo-count">
+        {`${todosLeft.length} items left`}
+      </span>
+
+      <nav className="filter">
+        {filters.map(({ href, status }) => (
+          <a
```
lerastarynets, line 31: i don't see key attribute for list items

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,35 @@
...
+import classNames from 'classnames';
+import React from 'react';
+import { Todo } from '../types/Todo';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoHeader: React.FC<Props> = ({ todos }) => {
+  const isSomeVisible = todos.length !== 0;
```
lerastarynets, line 11:
```suggestion
  const isSomeVisible = !!todos.length;
```

agniya-i (approved):

PR #736, opened 2023-08-07
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
+/* eslint-disable @typescript-eslint/no-shadow */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { TodoFooter } from './components/TodoFooter';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,22 +1,72 @@
...
+      default:
+        return true;
+    }
+  });
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setError('Error');
```
agniya-i, line 36: The error message is too generic. Consider providing more specific error messages for better debugging and user experience.

src/App.tsx
```diff
@@ -1,22 +1,72 @@
...
 
-const USER_ID = 0;
+const USER_ID = 11226;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [status, setStatus] = useState<Status>(Status.all);
+
+  const filteredTodos = todos.filter(todo => {
```
agniya-i, line 19: It would be better to wrap this funciton into useMemo

agniya-i (changes requested):

PR #737, opened 2023-08-07
todos loading, errors and filtering were implement

2pasha:
[DEMO LINK](https://2pasha.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,71 @@
+/* eslint-disable max-len */
+/* eslint-disable react/jsx-no-bind */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+import { SelectedCategory } from './types/SelectedCategory';
+import { ErrorMessage } from './types/ErrorMessage';
+
+import * as todoService from './api/todos';
+
+import { USER_ID } from './utils/constants';
+
+import { ToggleAll } from './component/ToggleAll';
+import { TodoInList } from './component/TodoInList';
+import { NewTodo } from './component/NewTodo';
+import { Filter } from './component/Filter';
+import { ClearCompleted } from './component/ClearCompleted';
+import { ErrorNotification } from './component/ErrorNotification';
+
+function filter(todos: Todo[], category: SelectedCategory) {
+  switch (category) {
+    case SelectedCategory.Active:
+      return [...todos].filter(todo => !todo.completed);
+
+    case SelectedCategory.Completed:
+      return [...todos].filter(todo => todo.completed);
+
+    default:
+      return [...todos];
+  }
+}
+
+const showError = (
+  message: ErrorMessage,
+  set: (error: ErrorMessage | null) => void,
+) => {
+  set(message);
+
+  setTimeout(() => {
+    set(null);
+  }, 3000);
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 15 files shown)


src/component/ErrorNotification.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <div
+      className={cn(
+        'notification is-danger is-light has-text-weight-normal',
+        { hidden: !error },
+      )}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setError}
```
Viktor-Kost, line 26: You're not actually calling the setError function on click. The parentheses are missing after setError. It should be onClick={() => setError()}

src/App.tsx
```diff
@@ -1,10 +1,71 @@
...
+  useEffect(() => {
+    todoService
+      .getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => showError(ErrorMessage.Load, setError));
+  }, []);
+
+  const filteredTodos = filter(todos, category);
+
+  const countActiveTodos = todos.reduce((prev, todo) => {
```
Viktor-Kost, line 61: Consider memoizing this with the useMemo hook to avoid recalculating on every render.

src/App.tsx
```diff
@@ -1,10 +1,71 @@
...
+  const [error, setError] = useState<ErrorMessage | null>(null);
+
+  useEffect(() => {
+    todoService
+      .getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => showError(ErrorMessage.Load, setError));
+  }, []);
+
+  const filteredTodos = filter(todos, category);
```
Viktor-Kost, line 59: Consider memoizing this with the useMemo hook to avoid recalculating on every render.

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

src/component/ErrorNotification.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <div
+      className={cn(
+        'notification is-danger is-light has-text-weight-normal',
+        { hidden: !error },
+      )}
+    >
+      <button
+        type="button"
+        className="delete"
+        onClick={() => setError()}
```
anastasiia-tilikina, line 26:
```suggestion
        onClick={setError}
```

when we use the function without params we can simplify this

src/App.tsx
```diff
@@ -1,10 +1,75 @@
...
+      .getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => showError(ErrorMessage.Load, setError));
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    return filter(todos, category);
+  }, [todos, category]);
+
+  const countActiveTodos = useMemo(() => {
```
anastasiia-tilikina, line 63: FYI
Try to use proper vars names) since `countActiveTodos` is a variable name, not a method it is better to name it as a noun like `activeTodosCount`. and if it was the method, for getting count it could be `getActiveTodosCount`, and eventually for methods that do something but not return we would use `countActiveTodos` the same as you used for `showError`

This is not a mistake though, but just keep in mind for the future)

src/component/ToggleAll.tsx
```diff
@@ -0,0 +1,18 @@
...
+  activeTodos: number,
+};
+
+export const ToggleAll: React.FC<Props> = ({
+  activeTodos,
+}) => {
+  return (
+    <button
+      type="button"
+      className={cn('todoapp__toggle-all', { active: activeTodos > 0 })}
```
anastasiia-tilikina, line 15:
```suggestion
      className={cn('todoapp__toggle-all', { active: activeTodos })}
```

anastasiia-tilikina (approved):
well done!
check my comments and consider this improvements for the future)

PR #739, opened 2023-08-09
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,66 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Status } from './enum/Status';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorMessage } from './enum/ErrorMessages';
+import { Error } from './components/Error/Error';
 
-const USER_ID = 0;
+const USER_ID = 11298;
+
+const getFilterdTodos = (
+  todos: Todo[],
+  filterBy: Status,
+) => {
+  const filteredTodos = [...todos];
+
+  switch (filterBy) {
+    case Status.all:
+      return filteredTodos;
+
+    case Status.active:
+      return filteredTodos.filter(todo => !todo.completed);
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,10 +1,66 @@
...
+      .then(setTodos)
+      .catch(() => setErrorWithTimeout(ErrorMessage.Add, 30000))
+      .finally(() => setIsLoading(false));
+  }, [todos]);
+
+  const filteredTodos = useMemo(() => {
+    return getFilterdTodos(todos, filterBy);
+  }, [todos, filterBy]);
+
+  const activeTodos = filteredTodos.filter(todo => !todo.completed);
```
Viktor-Kost, line 62: add useMemo for this filtering

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+            {`${activeTodosQuantity} items left`}
+          </span>
+
+          <nav className="filter">
+            <a
+              href="#/"
+              className={classNames('filter__link', {
+                selected: filterBy === Status.all,
+              })}
+              onClick={() => setFilterBy(Status.all)}
```
Viktor-Kost, line 30: it's better to move this callback to an own handler function

Viktor-Kost (changes requested):
Great job!
Some minor fixes needed

src/components/Footer/Footer.tsx
ShamievDavid: ok.
but still it will be callback, smth like onClick= {() => handleFilterStatus(Status.all)}

or you can give me some advice?

src/App.tsx
```diff
@@ -1,10 +1,68 @@
...
+
+    case Status.active:
+      return filteredTodos.filter(todo => !todo.completed);
+
+    case Status.completed:
+      return filteredTodos.filter(todo => todo.completed);
+
+    default:
+      return filteredTodos;
+  }
```
IvanFesenko, line 32: No need to spread
```suggestion
  switch (filterBy) {
    case Status.active:
      return todos.filter(todo => !todo.completed);

    case Status.completed:
      return todos.filter(todo => todo.completed);

    default:
      return todos;
  }
```

IvanFesenko (approved):
Check my cooment to avoid same mistake in the next task

PR #751, opened 2023-08-20
add task solution

IvanVaverchak:


src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,58 @@
...
+    setStatus(Status.ALL);
+  };
+
+  const handleActiveButton = () => {
+    setStatus(Status.ACTIVE);
+  };
+
+  const handleCompletedButton = () => {
+    setStatus(Status.COMPLETED);
+  };
```
VitaliyBondarenko1982, line 23: instead three funcion you can make one function factory and call with proper field for each button

```suggestion
  const handleFilter = (filter: Status) => () => {
    setStatus(filter);   
  };

```

src/components/TodoFilter.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <nav className="filter">
+      <a
+        href="#/"
+        className={classNames('filter__link', {
+          selected: status === Status.ALL,
+        })}
+        onClick={handleAllButton}
```
VitaliyBondarenko1982, line 32:
```suggestion
        onClick={handleFilter(Status.All)}
```
same for other filter buttons

VitaliyBondarenko1982 (changes requested):
Please provide demo link in PR description.

VitaliyBondarenko1982 (approved):

PR #757, opened 2023-08-24
add task solution

akozlovska:
[DEMO LINK](https://akozlovska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/TodosContext.tsx
+++ b/src/TodosContext.tsx
@@ -1,0 +1,65 @@
+import React, { useMemo, useState } from 'react';
+import { Error } from './types/Error';
+import { Filter } from './types/Filter';
+import { Todo } from './types/Todo';
+
+interface TodosGlobalContext {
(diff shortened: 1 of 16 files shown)


maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

src/TodosContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+          return !todo.completed;
+
+        case Filter.Completed:
+          return todo.completed;
+
+        case Filter.All:
+          return true;
+
+        default:
+          return true;
```
TarasHoliuk, line 48:
```suggestion
        case Filter.All:
        default:
          return true;
```
just FYI - you can do it in this way :arrow_up:

TarasHoliuk (approved):

PR #758, opened 2023-08-25
Todo app with loading with incorrect filters

ValeraViachalo:
[DEMO LINK](https://valeraviachalo.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,30 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+import { getTodos } from './api/todos';
 
-const USER_ID = 0;
+const USER_ID = 11368;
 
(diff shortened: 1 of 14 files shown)


ValeraViachalo:
Still doesnt works, but little bit better(i hope so😃).  Have some errors in console
image

src/App.tsx
```diff
@@ -1,24 +1,81 @@
...
+
+  const numberOfActive = useMemo(() => {
+    return getQuantityOfActiveTodos(todos);
+  }, [todos]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((currentTodo) => {
+        setTodo(currentTodo);
+      })
```
etojeDenys, line 38:
```suggestion
      .then(setTodo)
```
we can simplify it

src/App.tsx
```diff
@@ -1,24 +1,81 @@
...
+
+        return filteredTodos;
+
+      case Filter.completed:
+        filteredTodos = todos.filter(currentTodo => currentTodo.completed);
+
+        return filteredTodos;
+
+      default:
+        break;
```
etojeDenys, line 65: why do need filteredTodos var if we can just return the filter result directly

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #770, opened 2023-09-12
first commit

AnnaViolin23:
[DEMO LINK](https://AnnaViolin23.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoError.tsx
+++ b/src/components/TodoError.tsx
@@ -1,0 +1,35 @@
+import { useEffect } from 'react';
+import classNames from 'classnames';
+import { Error } from '../types/Error';
+
+type Props = {
+  error: Error,
+  onErrorChange: (error: Error) => void,
+};
+
+export const TodoError: React.FC<Props> = ({ error, onErrorChange }) => {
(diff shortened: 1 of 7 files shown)


etojeDenys (changes requested):
1. should be at the center
image

2. why do we need these two footers
image

src/components/TodoError.tsx
```diff
@@ -0,0 +1,35 @@
...
+  onErrorChange: (error: Error) => void,
+};
+
+export const TodoError: React.FC<Props> = ({ error, onErrorChange }) => {
+  useEffect(() => {
+    if (error) {
+      const timeotId = setTimeout(() => onErrorChange(Error.None),
+        3000);
+
+      clearTimeout(timeotId);
```
anastasiiavorobiova, line 16: The clearTimeout function should also be called in the cleanup function of useEffect to clear the timeout when the component unmounts or when the error changes.

anastasiiavorobiova (changes requested):
Great work! You shouldn't show clear completed button if there are no completed todos

PR #802, opened 2023-09-19
implement just loading of todos

Vannnkof:
[DEMO LINK](https://vannnkof.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,25 +1,75 @@
...
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const filteredTodos = getFilteredTodos(todos, filterTodos);
+  const activeTodos = getFilteredTodos(todos, FilterTodos.Active);
+  const completedTodos = getFilteredTodos(todos, FilterTodos.Completed);
```
alexander-ignatow, line 57: This is not optimal

It is better to calculate only on list of todos you should display, based on selected filters

alexander-ignatow (changes requested):
Conflict on the branch means you have to fork the repo again, copy your code, and re-submit PR

Also, your demo looks non-working - please make sure it performs like it should
Final comment - please break your App into components - for instance it makes sense to create `TodoList` and `TodoListItem` components - that will make your code easier to read and to maintain

PR #811, opened 2023-09-20
added ReactTodoApp LoadingTodos solution

NazarVotkanych:
[Demo](https://NazarVotkanych.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,47 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
+    getTodos(USER_ID)
+      .then(setTodo)
+      .catch((error) => {
+        setErrorMessage('Unable to load todos');
+        // eslint-disable-next-line no-console
+        console.warn(error);
+      });
+    const timerId = setInterval(() => {
+      setErrorMessage('');
+    }, 3000);
```
alexander-ignatow, line 35: You are cleaning the message every 3000 mSec - looks like it is a bit wrong

Consider using `setTimeout` instead

alexander-ignatow (changes requested):
Looks good - just a minor error is there

Optionally - you may consider better separation into components - `TodoList`, `TodoListItem`, `Error`, etc - it will make maintaining the code simpler

PR #972, opened 2023-11-13
add task solution

andrew17752985:
https://andrew17752985.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/components/TodoList.tsx
+++ b/src/components/TodoList.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import { Todo } from '../types/Todo';
+import { TodoItems } from './TodoItems';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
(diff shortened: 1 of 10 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,17 @@
...
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <TodoItems todo={todo} />
+      ))}
```
etojeDenys, line 14: do not forget to add key prop

etojeDenys (changes requested):
let's fix that when todos are loading
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)
still have this issue 
image

nazarmatsevych (approved):
Well done 👍

PR #977, opened 2023-11-17
add task solution

MariaSnegireva:


src/services/filterTodos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from "../types/Todo";
+import { Sort } from "../types/Sort";
+
+export function filterTodos(todos: Todo[], sortType: Sort) {
+  let newTodo = [...todos];
```
etojeDenys, line 5: this var looks redundant, you can just filter the todos directly

src/App.tsx
```diff
@@ -32,173 +104,119 @@ export const App: React.FC = () => {
...
+          </footer>
+        )}
       </div>
-
-      {/* Notification is shown in case of any error */}
-      {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
-        className="notification is-danger is-light has-text-weight-normal"
+        className={`notification is-danger is-light has-text-weight-normal ${errorMessage ? '' : 'hidden'}`}
```
etojeDenys, line 209: we usually use classnames package for such cases

src/App.tsx
```diff
@@ -1,28 +1,100 @@
...
+
+    fetchTodos();
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const isActiveTodo = todos.every(todo => !todo.completed);
+  const isNotCompleted = todos.filter(todo => !todo.completed).length;
```
etojeDenys, line 69: we usually name boolean vars using the "is", it does not look like the boolean var
```suggestion
  const nrOfNotCompleted = todos.filter(todo => !todo.completed).length;
```

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work! Go ahead!

PR #979, opened 2023-11-18
Developed

StepanMustafin:
[DEMO LINK](https://StepanMustafin.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './Components/UserWarning/UserWarning';
 import { TodoList } from './Components/TodoList';
 import { Footer } from './Components/Footer';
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,204 +1,84 @@
...
+import { USER_ID } from './utils/userId';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
-
-  return (
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>(todos);
```
VitaliyBondarenko1982, line 16: it is redundant to keep state for visible todos.
filter todos before render. whenever will change filterStatus - app will rerender and filter todos with current filter status.

src/App.tsx
```diff
@@ -1,204 +1,84 @@
...
+
+      case Status.Completed:
+        getTodosByStatus(USER_ID, true)
+          .then(setVisibleTodos)
+          .catch(() => setError(Errors.LoadError));
+        break;
+
+      default:
+        break;
+    }
```
VitaliyBondarenko1982, line 48:
```suggestion
```
looks like redundant to fetch todos each time when changed filterStatus. and every time will be fethc all todos and then also byStatus. 
fetch ont time all todos and just filter them when changing filterStatus

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

src/App.tsx
```diff
@@ -1,204 +1,78 @@
...
+      .catch(() => setError(Errors.LoadError));
+  }, []);
 
-            <span data-cy="TodoTitle" className="todo__title">
-              Not Completed Todo
-            </span>
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
+  const getVisibleTodos = () => {
```
anastasiia-tilikina, line 27:
```typescript
const getVisibleTodos = () => {
  return todos.filter(({ completed }) => {
    switch (filterStatus) {
      case Status.Active:
        return !completed;

      case Status.Completed:
        return completed;

      default:
        return true;
    }
  })
}
```

anastasiia-tilikina: let's simplify your function a bit.

anastasiia-tilikina (changes requested):
Well done! let's just add some improvement

SanyaBratashchuk (approved):

PR #988, opened 2023-11-26
stage-1

no4kar:
[DEMO LINK](https://no4kar.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,74 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+// import { UserWarning } from './UserWarning';
+import { TodoHeader } from './components/TodoHeader';
+import { TodoFooter } from './components/TodoFooter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Todo } from './types/Todo';
+import { TodoError } from './types/TodoError';
+import { getTodos } from './api/todos';
+import { Filter } from './types/Filter';
+import { getPraperedTodos } from './services/todos';
+import { TodoList } from './components/TodoList';
 
-const USER_ID = 0;
+const USER_ID = 11968;
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  // #region STATE
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMsg, setErrorMsg] = useState<TodoError>(TodoError.NONE);
+  const [filter, setFilter] = useState<Filter>(Filter.ALL);
+  // #endregion
+
+  // #region EFFECT
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setErrorMsg(TodoError.LOAD))
+      .finally(() => setIsLoading(false));
+  }, [USER_ID]);
+  // #endregion
+
+  const visibleTodos: Todo[] = getPraperedTodos(todos, filter);
+  const activeTodos: Todo[] = getPraperedTodos(todos, Filter.ACTIVE);
+  const handleFilterChange = (v: Filter) => {
+    setFilter(v);
+  };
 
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,204 +1,74 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-
-const USER_ID = 0;
+import React, { useEffect, useState } from 'react';
+// import { UserWarning } from './UserWarning';
```
VitaliyBondarenko1982, line 3: do not keep commented code

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,48 @@
...
+  errorMsg,
+  onErrorDelete = () => { },
+}) => {
+  useEffect(() => {
+    const timeoutID = setTimeout(onErrorDelete, 3000);
+
+    return () => {
+      clearTimeout(timeoutID);
+    };
+  }, []);
```
VitaliyBondarenko1982, line 19: fix eslint warning wiht missing dependency
Screenshot from 2023-11-26 16-21-39

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+      {/* Active filter should have a 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={cn('filter__link', {
+            selected: filter === Filter.ALL,
+          })}
+          data-cy="FilterLinkAll"
+          onClick={() => onFilterChange(Filter.ALL)}
```
VitaliyBondarenko1982, line 31: suggest to make one handler factory for all filter buttons
```
const handleFilter = (filterField: Filter) => () => onFilterChange(filterField);
```
```suggestion
          onClick={handleFilter(Filter.ALL)}
```

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check a couple comments.

src/components/ErrorNotification.tsx
no4kar: but I dont need side effects after rendering. start the timer only after it appears and stop if the user closed the message earlier. without dependency array the timers will be start after every render

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export const TodoItem: React.FC<Props> = ({
+  todo,
+}) => {
+  const [isEditing] = useState(false);
+  const { id, title, completed } = todo;
+
+  return (
+    <div
+      key={id}
```
lerastarynets, line 17: key prop shouldn't be passed like so, it should be passed while mapping items

src/utils/fetchClient.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 const BASE_URL = 'https://mate.academy/students-api';
+// const BASE_URL = 'https://CORRUPTED/students-api';
```
lerastarynets, line 3: pls no redundant comments

src/App.tsx
```diff
@@ -1,204 +1,69 @@
...
-            <form>
-              <input
-                data-cy="TodoTitleField"
-                type="text"
-                className="todo__title-field"
-                placeholder="Empty todo will be deleted"
-                value="Todo is being edited now"
-              />
-            </form>
+        {todos.length > 0 && (
```
lerastarynets, line 47:
```suggestion
        {!!todos.length && (
```

src/App.tsx
```diff
@@ -1,204 +1,69 @@
...
+
+  // #region EFFECT
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setErrorMsg(TodoError.LOAD))
+      .finally(() => setIsLoading(false));
+  }, [USER_ID]);
```
lerastarynets, line 31: unnecessary dependency: 'USER_ID', this value doesn't change, so it shouldn't be in dependencies array

lerastarynets (changes requested):
Almost there) Just several more things needs to be fixed

etojeDenys (approved):
good job

PR #997, opened 2023-11-30
Develop

xnquaint:
[DEMO LINK](https://xnquaint.github.io/react_todo-app-loading-todos/)


src/types/ErrorNotification.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorNotification {
```
ivbutsykin, line 1: Better to use default `const` instead `enum` here.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const Footer: React.FC<Props> = ({ setFilter, filterOption, todos }) => {
+  const completedTodo = todos.filter(todo => todo.completed).length;
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {
+          todosLeft === 1
+            ? `${todosLeft} item left`
```
ivbutsykin, line 21: Best way it's create `pluralize` helper in separate file to such cases: https://stackoverflow.com/a/39835908

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,80 @@
...
+          }
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilter(Filter.Completed)}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* don't show this button if there are no completed todos */}
+      {completedTodo && (
```
ivbutsykin, line 69: `completedTodo` it's not boolean type. In case that you cover `completedTodo` will be `0` and you receive `0` instead nothing:
```
console.log(0 && 'foo');
// 0
```

Options to fix:
1. `!!completedTodo && ...`
2.  `const hasCompleteTodo = todos.some(...);`

ivbutsykin (approved):
Good job!
Keep doing this project and take into account my notes.

PR #998, opened 2023-11-30
implemented todos loading, errors, filtering

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/react_todo-app-loading-todos/)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setHidden(false)}
+      />
+    </div>
+  ) : (
+    null
+  );
+};
```
etojeDenys, line 40:
```suggestion

  if (hidden) {
    return null
  }

  return (
    <div
      data-cy="ErrorNotification"
      className="notification is-danger is-light has-text-weight-normal"
    >
      {errorMessage}
      <button
        data-cy="HideErrorButton"
        type="button"
        className="delete"
        onClick={() => setHidden(false)}
      />
    </div>
  );
};
```
it will look much better

etojeDenys (approved):
good job

PR #1000, opened 2023-12-03
Solution

VKdrvtsv:
[DEMO LINK](https://VKdrvtsv.github.io/react_todo-app-loading-todos/)

My tests are pending, so I can`t wach all my errors. 

I implement only 1 error - which tell about empty title. I don`t realise which conditions should pretend them, so now it`s the only one.


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+          </form>
+        ) : (
+          <>
+            <span
+              data-cy="TodoTitle"
+              className="todo__title"
+              onDoubleClick={() => {
+                return setEditing(true);
+              }}
+            >
```
lerastarynets, line 79:
```suggestion
              onDoubleClick={() => setEditing(true)}
            >
```

PR #1004, opened 2023-12-06
result

errisi:
[DEMO LINK](https://errisi.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoFooter/TodoFooter.tsx
+++ b/src/components/TodoFooter/TodoFooter.tsx
@@ -1,0 +1,86 @@
+import React from 'react';
+import cn from 'classnames';
+
+import { Todo } from '../../types/Todo';
+import { Status } from '../../types/Status';
+
+type Props = {
+  todos: Todo[];
+  todoFilter: Status;
+  setTodoFilter: (todoFilter: Status) => void;
+};
+
+export const TodoFooter: React.FC<Props> = ({
+  todos,
+  todoFilter,
+  setTodoFilter,
+}) => {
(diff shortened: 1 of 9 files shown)


src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <footer
+      className="todoapp__footer"
+      data-cy="Footer"
+    >
+      <span
+        className="todo-count"
+        data-cy="TodosCounter"
+      >
+        {activeTodosCount}
```
ArtemSydun, line 33: and use jsx here like {`${activeTodos.length} items left`}  smth like that

src/types/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum Error {
+  LoadTodosError = 'Unable to load todos',
+  EmptyTitleError = 'Title should not be empty',
+  AddTodoError = 'Unable to add a todo',
+  DeleteTodoError = 'Unable to delete a todo',
+  UpdateTodoError = 'Unable to update a todo',
```
ArtemSydun, line 6: i don`t think it is necessary to use 'Error' in names cause you have already declared that it is enum Error, so it will be Error.LoadTodosError which is duplicating (i guess)

ArtemSydun (approved):

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

PR #1016, opened 2023-12-12
Task solution

Shostyy:
[DEMO LINK](https://shostyy.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,57 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
 
-const USER_ID = 0;
+import { Todo } from './types/Todo';
+
+import { getTodos } from './api/todos';
+
+import { Header } from './components/ErrorMessage/Header/Header';
+import { TodoList } from './components/ErrorMessage/TodoList/TodoList';
+import { Footer } from './components/ErrorMessage/Footer/Footer';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+
+const USER_ID = 12018;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[] | null>(null);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(response => setTodoList(response))
+      .catch(() => setErrorMessage('Unable to load todos'));
+  }, []);
+
+  const [activeTodos, completedTodos] = useMemo(() => {
+    let active = 0;
+    let completed = 0;
+
+    if (todoList) {
+      active = todoList.filter(todo => !todo.completed).length;
+      completed = todoList.filter(todo => todo.completed).length;
+    }
+
+    return [active, completed];
+  }, [todoList]);
+
+  const todoListToShow: Todo[] | null = useMemo(() => {
+    if (todoList) {
+      switch (filter) {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
+      completed = todoList.filter(todo => todo.completed).length;
+    }
+
+    return [active, completed];
+  }, [todoList]);
+
+  const todoListToShow: Todo[] | null = useMemo(() => {
+    if (todoList) {
+      switch (filter) {
+        case 'completed':
```
lerastarynets, line 41: pls add 'completed' and 'active' values to an enum and use where needed

src/App.tsx
```diff
@@ -14,191 +61,38 @@ export const App: React.FC = () => {
...
+            todoList={todoListToShow}
+          />
+        )}
+
+        {todoList && (
+          <Footer
+            activeTodos={activeTodos}
+            completedTodos={completedTodos}
+            filter={filter}
+            setFilter={(newFilter: string) => setFilter(newFilter)}
```
lerastarynets, line 79:
```suggestion
            setFilter={setFilter}
```

src/components/ErrorMessage/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  todoList,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todoList.map(todo => (
+        <div
+          data-cy="Todo"
+          className={cn('todo', {
+            completed: todo.completed,
+          })}
```
lerastarynets, line 20: where is `key` prop?

lerastarynets (changes requested):
when todos are loading, the field looks weird, pls add a loader
Screenshot at Dec 12 16-34-39

src/App.tsx
```diff
@@ -1,10 +1,59 @@
...
+const USER_ID = 12018;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[] | null>(null);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<TodoListState>(TodoListState.All);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(response => setTodoList(response))
```
etojeDenys, line 24:
```suggestion
      .then(setTodoList)
```
we can simplify it

src/App.tsx
```diff
@@ -1,10 +1,59 @@
...
+import { TodoList } from './components/ErrorMessage/TodoList/TodoList';
+import { Footer } from './components/ErrorMessage/Footer/Footer';
+import { ErrorMessage } from './components/ErrorMessage/ErrorMessage';
+import { TodoListState } from './types/TodoListState';
+import { Errors } from './types/Errors';
+
+const USER_ID = 12018;
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[] | null>(null);
```
etojeDenys, line 18:
```suggestion
  const [todoList, setTodoList] = useState<Todo[]>([]);
```
let's just set the empty array by default

src/App.tsx
```diff
@@ -1,10 +1,59 @@
...
+        case TodoListState.Active:
+          return todoList.filter(todo => todo.completed);
+
+        default:
+          return todoList;
+      }
+    }
+
+    return null;
+  }, [todoList, filter]);
```
etojeDenys, line 55: now you can simplify this logic(remove the if and return null logic)

etojeDenys (changes requested):
probably, you forgot to re-deploy your work, 'cause I still can see it
image

lerastarynets (approved):
You're on the right track!

PR #1017, opened 2023-12-13
Solution

BoblykNazar:
[DEMO LINK](https://BoblykNazar.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,0 +1,131 @@
+import cn from 'classnames';
+import { FC } from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+};
+
+export const TodoList: FC<Props> = ({
+  todos,
+}) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {/* This is a completed todo */}
+      {todos.map(todo => {
+        const {
+          title,
+          id,
+        } = todo;
+
+        return (
+          <div
+            data-cy="Todo"
+            className={cn('todo', { completed: todo.completed })}
+            key={id}
+          >
+            <label className="todo__status-label">
+              <input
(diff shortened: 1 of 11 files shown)


src/helpers.ts
```diff
@@ -0,0 +1,26 @@
+import { Status } from './types/Status';
+import { Todo } from './types/Todo';
+
+export function preperedTodos(
+  todos: Todo[],
+  selectFilter: Status,
+): Todo[] {
+  let filteredTodo = [...todos];
```
VeDono, line 8: filter not mutate original data, so i that case you can forgot about cloning original array.

(but i can wrong, mentors know better)

src/types/Error.ts
```diff
@@ -0,0 +1,7 @@
+export enum Error {
```
VeDono, line 1: Good idea, maybe i litle-bit steal that 👀

VeDono (approved):
Nice.

Maybe I will `steal` some things 👀

src/helpers.ts
BoblykNazar: Agree with you=)

src/types/Error.ts
BoblykNazar: Your welcome and thanks for review=)

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,131 @@
...
+              <div className="loader" />
+            </div>
+          </div>
+        );
+      })}
+    </section>
+  );
+};
+
+// eslint-disable-next-line no-lone-blocks
```
ArtemTeslenko, line 56: delete unnecessary comments

src/helpers.ts
ArtemTeslenko: right filter does not mutate array

ArtemTeslenko (approved):
Good job 👌

PR #1022, opened 2023-12-14
done first part of ToDo app

Anton-Kuchmasov:
[DEMO LINK](https://Anton-Kuchmasov.github.io/react_todo-app-loading-todos/)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,73 @@
...
+  todosLength,
+  onStatus,
+  status,
+  hasCompletedTodos,
+  handleClear,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todosLength} items left`}
```
emilionila, line 22: shows all todos length, while it should only show remaining todos

src/App.tsx
```diff
@@ -1,11 +1,70 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

src/App.tsx
```diff
@@ -1,11 +1,70 @@
...
+    switch (status) {
+      case Status.ACTIVE:
+        return todosFromServer.filter(todo => !todo.completed);
+
+      case Status.COMPLETED:
+        return todosFromServer.filter(todo => todo.completed);
+
+      default:
+        return todosFromServer;
+    }
```
roman-mirzoian, line 54: Filtering logic should be brought out as a separate helper outside of components to improve code readability

src/App.tsx
```diff
@@ -14,191 +73,25 @@ export const App: React.FC = () => {
...
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-        </section>
+        {todosFromServer.length > 0 && <TodoList todos={filteredTodos} />}
 
         {/* Hide the footer if there are no todos */}
```
roman-mirzoian, line 80: Do not leave unnecessary comments.
Fix it in all the code.

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,69 @@
+import { ErrorSpec } from '../../types/ErrorSpec';
+
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 3: Try to rewrite the code so that it is not necessary to disable the linter

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,69 @@
...
+            return (
+              <span>
+                Unable to update a todo
+                <br />
+              </span>
+            );
+          }
+
+          default:
+            return null;
```
roman-mirzoian, line 64:
```suggestion
            return <></>;
```

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const Errors: React.FC<Props> = ({ error }) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className="notification is-danger is-light has-text-weight-normal"
+    >
+      <button data-cy="HideErrorButton" type="button" className="delete" />
+      {/* show only one message at a time */}
+      {(() => {
```
roman-mirzoian, line 16: Calling such a function is hard to read and hard to maintain. It is better to make it as a separate helper that will return the desired tag and this tag is ready to be inserted into the code.
Also, each case part differs only in text, so there is no need to repeat the same tags every time, it is enough to calculate the required text and wrap it in the required tag

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

ArtemTeslenko (approved):
Great 🔥

PR #1034, opened 2023-12-17
solution

YehorHeneralov:
[DEMO LINK](https://YehorHeneralov.github.io/react_todo-app-loading-todos/)


mickhakk (approved):

src/App.tsx
```diff
@@ -1,204 +1,62 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+/* eslint-disable react-hooks/rules-of-hooks */
```
ArtemTeslenko, line 1: try to rewrite without disabling the linter

src/App.tsx
```diff
@@ -1,204 +1,62 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<TypeOfFilter>(TypeOfFilter.All);
+  const [error, setError] = useState<Error | null>(null);
+
+  const filteredTodos = useMemo(() => {
+    return filterTodos(todos, filter);
+  }, [todos, filter]);
+
+  useEffect(() => {
+    setError(null);
```
ArtemTeslenko, line 30: error has `null` as initial value, no need to set it again here, because useEffect has an empty array of deps so component will have state `error` `null` before mounting

src/components/Errors/Errors.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const Errors: React.FC<Props> = ({
+  error,
+  setError,
+}) => {
+  useEffect(() => {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }, [error, setError]);
```
ArtemTeslenko, line 18: you need to add state into array of deps, not a function, `setError` should not be here

src/components/header/Header.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+export const TodoHeader: React.FC = () => {
+  return (
+    <header className="todoapp__header">
+      {/* this buttons is active only if there are some active todos */}
```
ArtemTeslenko, line 6: remove unnecessary comments

src/components/main/TodoItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+          className="todo__status"
+          checked={completed}
+        />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {title}
+      </span>
+
+      {/* Remove button appears only on hover */}
```
ArtemTeslenko, line 27: the same, unnecessary comments

src/helpers/helper.ts
```diff
@@ -0,0 +1,19 @@
+import { Todo } from '../types/Todo';
+import { TypeOfFilter } from '../types/TypeOfFilters';
+
+export const filterTodos = (
+  todos: Todo[],
+  filter: TypeOfFilter,
+) => {
+  let copy = [...todos];
```
ArtemTeslenko, line 8: `copy` is very common name, also filter does not mutate array, you can filter your todos directly

ArtemTeslenko (changes requested):
Good job 👍 See comments and add more todos (active/completed), to have ability to check the filter's work)

src/App.tsx
YehorHeneralov: i tried to add USER_ID in dependancy array

  useEffect(() => {
    getTodos(USER_ID)
      .then(setTodos)
      .catch(() => setError(Error.LoadTodos));
  }, [USER_ID]);
  
  but it didn't help, i don't know how to fix it

ArtemTeslenko: because it should be state or prop in array of deps

src/App.tsx
```diff
@@ -1,204 +1,60 @@
...
+  const [error, setError] = useState<Error | null>(null);
+
+  const filteredTodos = useMemo(() => {
+    return filterTodos(todos, filter);
+  }, [todos, filter]);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => setError(Error.LoadTodos));
```
ArtemTeslenko, line 32: also there should be loader, when you load data from server

ArtemTeslenko (approved):
Nice 👌 I am approving, but see comments)

PR #1042, opened 2023-12-21
decompose project, implement todos loading, errors handling, filtering,

Softjey:
[DEMO LINK](https://softjey.github.io/react_todo-app-loading-todos/)


src/contexts/TodosContext.tsx
```diff
@@ -0,0 +1,90 @@
...
+}) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState(TodosFilterQuery.all);
+  const [error, setError] = useState('');
+
+  useEffect(() => {
+    getTodos()
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+      })
```
etojeDenys, line 42:
```suggestion
      .then(setTodos)
```
you can simplify it

etojeDenys (changes requested):
looks like redundant number :)
image

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,74 @@
...
+          Completed
+        </a>
+      </nav>
+
+        <button
+          type="button"
+          className="todoapp__clear-completed"
+          data-cy="ClearCompletedButton"
+        >
+          {completedTodosAmount > 0 &&'Clear completed'}
```
TarasHoliuk, line 70: now it seems a bit shifted to the right side:
image

it's better to use `visibility: hidden` CSS property to hide the button but keep space for it (in your case when button will appear the filter will "jump" left)

But you can fix it in the next stage task :)

TarasHoliuk (approved):
Approved with one comment (read it and apply in the next task)

PR #1045, opened 2023-12-22
Solution

Patrykmclaren14:
[Solution](https://Patrykmclaren14.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoCard/TodoCard.tsx
+++ b/src/components/TodoCard/TodoCard.tsx
@@ -1,0 +1,37 @@
+import { Todo } from '../../types/Todo';
+
+interface TodoCardProps {
+  todo: Todo
+}
+
+export const TodoCard: React.FC<TodoCardProps> = ({ todo }) => {
+  return (
+    <div data-cy="Todo" className={`todo ${todo.completed ? 'completed' : ''} `}>
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={!!todo.completed}
+        />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
+      </span>
+
+      <button
+        type="button"
+        className="todo__remove"
+        data-cy="TodoDelete"
+      >
+        ×
+      </button>
+
+      <div data-cy="TodoLoader" className="modal overlay">
+        <div className="modal-background has-background-white-ter" />
+        <div className="loader" />
+      </div>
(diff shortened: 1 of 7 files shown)


src/components/TodoCard/TodoCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const TodoCard: React.FC<TodoCardProps> = ({ todo }) => {
+  return (
+    <div data-cy="Todo" className={`todo ${todo.completed ? 'completed' : ''} `}>
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={!!todo.completed}
```
jacekkosiba, line 15: Maybe I am missing something but why not just
`checked={todo.completed}` ?

jacekkosiba (approved):

src/components/TodoCard/TodoCard.tsx
JoyCoffeeAddict: correct

src/context/TodosProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+      case SortType.active:
+        setFilteredTodos(todos.filter(todo => !todo.completed));
+        break;
+      case SortType.completed:
+        setFilteredTodos(todos.filter(todo => todo.completed));
+        break;
+      default:
+        break;
+    }
+  }, [filterType, todos, setFilteredTodos]);
```
JoyCoffeeAddict, line 79: this should be `useMemo` not `useEffect`, because you are making calculations based on another state to get a value.

JoyCoffeeAddict (changes requested):

mbruhler (approved):
Good job, all works as expected and code is well written!

src/context/TodosProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const handleSubmitSent = useCallback(
+    (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+      setMessageError('');
+      if (!query) {
+        setMessageError(ERROR_MESSAGES[1]);
```
mbruhler, line 86: This should be enum 100%!

mbruhler (commented):
One thing to think about for the future

src/context/TodosProvider.tsx
mbruhler: I mean the error

PR #1049, opened 2023-12-27
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_todo-app-loading-todos/)


MarcinLigmann (approved):
In my opinion code looks good, the link to the demo version doesn't work.

src/App.tsx
```diff
@@ -1,204 +1,112 @@
...
+        filteredTodos = filteredTodos.filter(todo => !todo.completed);
+        break;
+      case 'Completed':
+        filteredTodos = filteredTodos.filter(todo => todo.completed);
+        break;
+      default:
+        break;
+    }
+  }
+
```
JoyCoffeeAddict, line 81: I would already put all of that into a Context

src/App.tsx
```diff
@@ -1,204 +1,112 @@
...
+import { Todo, Filter } from './types';
+import {
+  Header, TodoList, Footer, Errors, UserWarning,
+} from './components';
+import { getTodos } from './api/todos';
+
+const USER_ID = 12107;
+
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
```
JoyCoffeeAddict, line 13: I don't think you need additional `null` state here.

```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/components/Errors.tsx
```diff
@@ -0,0 +1,34 @@
+import { FC } from 'react';
+import cn from 'classnames';
+
+type Props = {
+  showError: boolean,
+  setShowError: (value: boolean) => void,
+  errorMessage: string,
```
JoyCoffeeAddict, line 7: I don't think that you need both `showError` and `errorMessage` . 
If there is an `errorMessge` then show it. 
if you want to hide the error, just  clear the `errorMessage`

src/App.tsx
```diff
@@ -1,204 +1,112 @@
...
-          {/* This todo is not completed */}
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
+        <Header />
+        {
+          todos && !!todos.length && (
```
JoyCoffeeAddict, line 89: In my opinion that is the most readable
```suggestion
          todos.length > 0  (
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+          selected: selectedFilter === 'Completed',
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={handleFilterChange}
+      >
+        Completed
+      </a>
+    </nav>
+
+    {/* don't show this button if there are no completed todos */}
```
JoyCoffeeAddict, line 59: Implement

src/components/TodoList.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Todo } from '../types/Todo';
+import { SingleTodo } from './SingleTodo';
+
+type Props = {
+  filteredTodos: Todo[] | undefined,
+};
+
+export const TodoList: FC<Props> = ({ filteredTodos }) => (
+  <section className="todoapp__main" data-cy="TodoList">
+    {/* DISPLAYING TODOS */}
```
JoyCoffeeAddict, line 12: I think it's pretty obvious

src/temp/error-messages.tsx
JoyCoffeeAddict, line 1: I don't think you use this file

src/temp/todo-template.tsx
JoyCoffeeAddict, line 1: same here

JoyCoffeeAddict (changes requested):

adrian200002 (approved):

src/components/Errors.tsx
jacekkosiba: The thing is, if I set the errorMessage to an empty string, the error message disappears abruptly, and I believe it should disappear gradually

src/context/AppContext.tsx
mpiskadlo, line 1: Provide initial values for the context, then the type of it wouldn't have to be `AppContextType | undefined`. Then you wouldn't have to check this:
```javascript
  if (!context) {
    return <p>Context not provided.</p>;
  }
```
I would also create a AppContextProvider component (keep all context state & functions there) instead of adding `AppContext.Provider` inside of App.tsx.
Additionaly create a hook useAppContext, it will make you write a bit less code ;)
```javascript
export const useAppContext = () => useContext(ContextGoesHere);
```

mpiskadlo (changes requested):

mbruhler (approved):

PR #1050, opened 2023-12-27
Develop

szymanski-marcin:
[DEMO LINK](https://szymanski-marcin.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/ErrorNotification/ErrorNotification.tsx
+++ b/src/components/ErrorNotification/ErrorNotification.tsx
@@ -1,0 +1,29 @@
+import { FC } from 'react';
+
+interface Props {
+  hideErrorMessage: () => void,
+  errorMessage: string | null,
+}
+
+export const ErrorNotification: FC<Props> = (props) => {
+  const {
+    hideErrorMessage,
+    errorMessage,
+  } = props;
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className="notification is-danger is-light has-text-weight-normal"
+    >
+      {errorMessage}
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={hideErrorMessage}
+        aria-label="Hide Error Message"
+      />
+    </div>
+  );
+};
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,6 +1,7 @@
 import { FC } from 'react';
 import classnames from 'classnames';
 import { Todo } from '../../types/Todo';
+import { FilterValue } from '../../types/FilterValue';
 
 interface Props {
   todos: Todo[],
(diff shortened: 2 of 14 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const Footer: FC<Props> = (props) => {
+  const {
+    todos,
+    setFilterValue,
+    filterValue,
+  } = props;
+
+  const remainingTodosLength = todos.filter(todo => !todo.completed).length;
+  const completedTodosLength = todos.filter(todo => todo.completed).length;
```
zaura333, line 20: Remaining todos should be taken from all the todos before filtering, not just currently rendered ones. Now while filtering by 'completed' the message shows 0

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const ErrorNotification: FC<Props> = (props) => {
+  const {
+    hideErrorMessage,
+    errorMessage,
+  } = props;
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className="notification is-danger is-light has-text-weight-normal"
```
zaura333, line 17: I think that error component should be rendered regardless of occurrence of any error, and if there isn't any, it should be hidden with class 'hidden' (as in README). Also, notification should be automatically hidden after 3 seconds

zaura333 (approved):
Nice! I've included some suggestions

src/components/Footer/Footer.tsx
mpiskadlo: Agreed, 0 items left looks weird 🤔

mpiskadlo (changes requested):

src/components/Footer/Footer.tsx
szymanski-marcin: done

src/components/ErrorNotification/ErrorNotification.tsx
szymanski-marcin: done

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,41 @@
...
+      setIsVisible(true);
+
+      const timeout = setTimeout(() => {
+        hideErrorMessage();
+      }, 3000);
+
+      return () => clearTimeout(timeout);
+    }
+
+    return () => {};
```
JoyCoffeeAddict, line 23: use a [guard clause](https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/) instead of doing this empty return later 
```suggestion
    if (!errorMessage) {
    return;
    }
    
      setIsVisible(true);

      const timeout = setTimeout(() => {
        hideErrorMessage();
      }, 3000);

      return () => clearTimeout(timeout);
```

JoyCoffeeAddict (changes requested):

mpiskadlo (approved):
Looks good!

PR #1051, opened 2023-12-28
Solution

MarcinLigmann:
[DEMO LINK](https://MarcinLigmann.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/todoerr/todoerr.tsx
+++ b/src/components/todoerr/todoerr.tsx
@@ -1,33 +1,45 @@
+import classNames from 'classnames';
+import { useEffect } from 'react';
 import { useTodos } from '../../context/todoProvider';
 
 export const TodoErr = () => {
-  const { error } = useTodos();
+  const { error, setError } = useTodos();
+
+  useEffect(() => {
+    setTimeout(() => {
+      setError(null);
+    }, 3000);
+  }, [error, setError]);
 
   return (
-  // Notification is shown in case of any error
-  // Add the 'hidden' class to hide the message smoothly
-    <>
-      {error
-      && (
-        <div
-          data-cy="ErrorNotification"
-          className="notification is-danger is-light has-text-weight-normal"
-        >
-          {/* <button data-cy="HideErrorButton" type="button" className="delete" /> */}
-          {/* show only one message at a time */}
-          {`"${error}"`}
-          <br />
-          Unable to load todos
-          <br />
-          Title should not be empty
-          <br />
-          Unable to add a todo
-          <br />
-          Unable to delete a todo
-          <br />
-          Unable to update a todo
-        </div>
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
+          hidden: !error,
+        },
       )}
-    </>
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        aria-label="hide error button"
+        onClick={() => setError(null)}
+      />
+
+      {/* show only one message at a time */}
+      {error}
+      <br />
+      {/* Unable to load todos
+        <br />
+        Title should not be empty
+        <br />
+        Unable to add a todo
+        <br />
+        Unable to delete a todo
+        <br />
+        Unable to update a todo */}
+    </div>
   );
 };
--- a/src/components/todoerr/todoerr.tsx
+++ b/src/components/todoerr/todoerr.tsx
@@ -1,0 +1,33 @@
+import { useTodos } from '../../context/todoProvider';
+
+export const TodoErr = () => {
+  const { error } = useTodos();
+
+  return (
+  // Notification is shown in case of any error
+  // Add the 'hidden' class to hide the message smoothly
+    <>
+      {error
+      && (
+        <div
+          data-cy="ErrorNotification"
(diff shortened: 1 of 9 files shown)


src/components/todoerr/todoerr.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }, [error, setError]);
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
+          hidden: !error,
+        },
+      )}
```
kate-prakofyeva, line 21: // {/* Add the 'hidden' class to hide the message smoothly */}
  it's mean, than you need to add transition for styled as smoothly

src/context/todoProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+    error,
+    setError,
+  };
+
+  useEffect(() => {
+    try {
+      getTodos(12075)
+        .then(data => setTodos(dataFilter(data, filterBy)));
+    } catch (err) {
+      setError('Unable to load todos');
```
kate-prakofyeva, line 67: add enum type for erorr messages, and show only one of them in your case setErrors(ErrorType.Load);
```export enum ErrorType {
  Load = 'Unable to load todos',
  Title = 'Title should not be empty',
  Add = 'Unable to add a todo',
  Delete = 'Unable to delete a todo',
  Update = 'Unable to update a todo',
}```

File: src/context/todoProvider.tsx [hunk 3]
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    const counter = todos.filter(task => !task.completed).length;
```
kate-prakofyeva, line 70: FYI you can use your previous function for get active todos number
const counter= dataFilter(todos, FilterType.Active);

src/context/todoProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+      return data.filter(task => !task.completed);
+    case 'completed':
+      return data.filter(task => task.completed);
+    default:
+      return data;
+  }
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
```
kate-prakofyeva, line 45: I prefer to separate todos and todos after applying filter, making one more state, for save previous todos state untouchable

but I will ask mentors about this

kate-prakofyeva (approved):
maybe some comments are unnecessary, this is just a recommendation
Good job!

src/context/todoProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+      getTodos(USER_ID)
+        .then(data => setTodos(dataFilter(data, filterBy)));
+    } catch (err) {
+      setError(ErrorTypes.Load);
+    }
+
+    const counter = dataFilter(todos, 'active').length;
+
+    setCount(counter);
+  }, [todos, filterBy, count]);
```
mpiskadlo, line 76: Open Network Tab in the devtools and see how many requests are being send.
It's because of dependencies array here. You definitely don't want `todos` to be part of it, because you update this array here, which leads to endless requests. Examine also the rest of dependencies.
Maybe you shouldn't fetch todos each time filter changes and just simply filter them?

mpiskadlo (changes requested):

src/components/todofooter/todofooter.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const TodoFooter = () => {
+  const { filterBy, setFilterBy, count } = useTodos();
+
+  return (
+    // Hide the footer if there are no todos
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {count}
+        {' '}
+        items left
```
JoyCoffeeAddict, line 13: check if `count > 0` I wouldn't display `0 items left`

src/context/todoProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }, []);
+
+  useEffect(() => {
+    const visible = dataFilter(todos, filterBy);
+
+    setVisibleTodos(visible);
+    const counter = dataFilter(todos, 'active').length;
+
+    setCount(counter);
+  }, [filterBy, todos]);
```
JoyCoffeeAddict, line 88: this should be `useMemo` not `useEffect`, because you are making calculations based on another state to get a value.

JoyCoffeeAddict (changes requested):

src/context/todoProvider.tsx
```diff
@@ -0,0 +1,105 @@
...
+  //   setCount(counter);
+  // }, [filterBy, todos]);
+  useMemo(() => {
+    const visible = dataFilter(todos, filterBy);
+    const counter = dataFilter(todos, 'active').length;
+
+    setVisibleTodos(visible);
+    setCount(counter);
+    // return { visibleTodos, counter };
+  }, [filterBy, todos]);
```
JoyCoffeeAddict, line 96: read about [useMemo](https://react.dev/reference/react/useMemo) one more time, you should always return a value here, and never do the side effects (setting state)

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #1053, opened 2023-12-28
Solution

zaura333:
[DEMO LINK](https://zaura333.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/api/todos.ts
+++ b/src/api/todos.ts
@@ -6,3 +6,6 @@ export const getTodos = (userId: number) => {
 };
 
 // Add more methods here
+export const postTodo = (userId: number, todo: Omit<Todo, 'id'>) => {
+  return client.post<Todo>(`/todos?userId=${userId}`, todo);
+};
--- a/src/components/Error.tsx
+++ b/src/components/Error.tsx
@@ -1,0 +1,51 @@
+import { useEffect, useState } from 'react';
+// import { ErrorType } from '../types/ErrorType';
+import classNames from 'classnames';
+import { useTodo } from '../providers/TodoProvider';
+
+export const Error = () => {
+  const { error, setError } = useTodo();
+  const [isHidden, setIsHidden] = useState<boolean>(true);
+
+  enum Errors {
+    load = 'Unable to load todos',
+    title = 'Title should not be empty',
+    add = 'Unable to add a todo',
+    delete = 'Unable to delete a todo',
+    update = 'Unable to update a todo',
+  }
+
+  useEffect(() => {
+    setIsHidden(!error);
+
+    setTimeout(() => {
+      setIsHidden(true);
+      setError(null);
+    }, 3000);
+  }, [error, setError]);
+
+  const handleClick = () => {
+    setError(null);
+    setIsHidden(true);
+  };
+
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal', {
+          hidden: isHidden,
+        },
+      )}
+    >
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={handleClick}
+      />
+      {error && Errors[error]}
+    </div>
+  );
+};
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -1,0 +1,67 @@
+import classNames from 'classnames';
+import { useTodo } from '../providers/TodoProvider';
+import { Filter } from '../types/Filter';
+
+export const TodoFooter = () => {
+  const {
+    todos, activeFilter, setActiveFilter, todosLeft,
+  } = useTodo();
+
+  const handleClick = (filter: Filter) => () => {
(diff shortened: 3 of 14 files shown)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <section className="todoapp__main" data-cy="TodoList">
+      {visibleTodos.map(todo => {
+        return (
+          <div
+            key={todo.id}
+            data-cy="Todo"
+            className={classNames('todo', {
+              completed: todo.completed,
+            })}
+            onClick={handleClick(todo.id)}
```
Patrykmclaren14, line 46: It's unnecessary to use a click event and check event.detail === 2. You can simply use a doubleClick event

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: activeFilter === 'all',
+          })}
+          data-cy="FilterLinkAll"
+          onClick={handleClick('all')}
+        >
+          All
+        </a>
```
Patrykmclaren14, line 31: You can move the filter to another component and create an array to map filter values. In this scenario, if a value is repeated three times, you can express this more clearly

For Example
`const FILTER_VALUES = ['All', 'Active', 'Completed'];
        {FILTER_VALUES.map(filter => (
          <Filter filter={filter} key={filter} />
        ))}
`

src/providers/TodoProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }, []);
+
+  useEffect(() => {
+    let uncompleted = 0;
+
+    todos.forEach(todo => {
+      if (!todo.completed) {
+        uncompleted += 1;
+      }
+    });
```
Patrykmclaren14, line 53: In my opinion, it would be better to use `reduce` for counting

src/api/todos.ts
```diff
@@ -6,3 +6,6 @@ export const getTodos = (userId: number) => {
 };
 
 // Add more methods here
+export const postTodo = (userId: number, todo: Omit<Todo, 'id'>) => {
+  return client.post<Todo>(`/todos?userId=${userId}`, todo);
```
JoyCoffeeAddict, line 10: You don't need a query param in your `post` it's defined in the API (check readme) 
```suggestion
  return client.post<Todo>(`/todos`, todo);
```

src/components/Error.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { error, setError } = useTodo();
+  const [isHidden, setIsHidden] = useState<boolean>(true);
+
+  enum Errors {
+    load = 'Unable to load todos',
+    title = 'Title should not be empty',
+    add = 'Unable to add a todo',
+    delete = 'Unable to delete a todo',
+    update = 'Unable to update a todo',
+  }
```
JoyCoffeeAddict, line 16: create enum outside of the component.

src/components/TodoFooter.tsx
JoyCoffeeAddict: That is a great idea.

src/components/TodoList.tsx
JoyCoffeeAddict: true.

src/providers/TodoProvider.tsx
JoyCoffeeAddict: technically yes, but this is equaly readable and performant

src/providers/TodoProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+      switch (activeFilter) {
+        case 'active':
+          return !todo.completed;
+        case 'completed':
+          return todo.completed;
+        default:
+          return true;
+      }
+    }));
+  }, [todos, activeFilter]);
```
JoyCoffeeAddict, line 70: this should be `useMemo` not `useEffect`, because you are making calculations based on another state to get a value.

JoyCoffeeAddict (changes requested):

mbruhler (changes requested):
I don't understand why on adding new todo or clicking enter precisely, the page refreshes? Please tell me, overall it looks good and works closely as expected

zaura333:
> I don't understand why on adding new todo or clicking enter precisely, the page refreshes? Please tell me, overall it looks good and works closely as expected

In this task I've only implemented loading todos, filtering and errors. Adding todos and functioning form is in the next task, so there's no submit handler yet. I can add some temporary handler if needed

mbruhler:
No it's ok, I approve then because the provided functionality works ok

src/types/ErrorType.ts
```diff
@@ -0,0 +1,9 @@
+export type ErrorType = '' | 'load' | 'title' | 'add' | 'delete' | 'update';
+
+export enum Errors {
+  load = 'Unable to load todos',
+  title = 'Title should not be empty',
+  add = 'Unable to add a todo',
+  delete = 'Unable to delete a todo',
+  update = 'Unable to update a todo',
```
mbruhler, line 8: Please remember for the future to write enums UPPERCASE

mbruhler (approved):

PR #1056, opened 2023-12-28
Implemented todos loading, filter and errors.

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,43 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
-const USER_ID = 0;
+const USER_ID = 12113;
 
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[] | null>(null);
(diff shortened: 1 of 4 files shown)


mbruhler (changes requested):
Your solution is not working or you didn't deploy last changes.

WiktoriaBielec (approved):
Great code:)

src/components/TodoList.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [allTodos, inputRef]);
+
+  return (
+    <>
+      {visibleTodos?.map(todo => (
```
mpiskadlo, line 84: I think you should create a component for a todo, maybe you could use it for tempTodo either. This would make `TodoList` way cleaner

src/App.tsx
```diff
@@ -16,188 +98,56 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <header className="todoapp__header">
```
mpiskadlo, line 99: I think it would be good to move the header to a separate component

mpiskadlo (changes requested):
You were supposed to implement ONLY todos loading, errors, and filtering. But I guess it wouldn't make much sense to force you to remove adding & deleting functionality 😅 
I got some small improvements though

mbruhler (approved):

PR #1066, opened 2024-01-02
add task solution

7yrak34:
[DEMO LINK](https://7yrak34.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { Status } from '../../types/Status';
+import { Filter } from '../Filter';
+
+interface Props {
+  todos: Todo[];
+  filterBy: Status;
+  onAllclick: () => void;
+  onActiveClick: () => void;
+  onCompletedclick: () => void;
+}
+
+export const Footer: React.FC<Props> = React.memo(({
+  todos,
+  filterBy,
+  onAllclick,
+  onActiveClick,
+  onCompletedclick,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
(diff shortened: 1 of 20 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { TodoItem } from '../TodoItem/TodoItem';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = React.memo(({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => <TodoItem todo={todo} />)}
```
l4st1m0za, line 12: Each child in a list should have a unique 'key' prop. Add a unique key prop to the TodoItem component inside the map function.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onActiveClick: () => void;
+  onCompletedclick: () => void;
+}
+
+export const Footer: React.FC<Props> = React.memo(({
+  todos,
+  filterBy,
+  onAllclick,
+  onActiveClick,
+  onCompletedclick,
```
l4st1m0za, line 19: The method name 'onCompletedclick' should follow camel case convention. Consider renaming to 'onCompletedClick' for consistency and readability.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(todo => !todo.completed).length} items left`}
```
l4st1m0za, line 24: Filter operation is directly used within the return statement. For better performance and readability, consider performing this operation before the return statement and storing the result in a variable.

l4st1m0za (changes requested):
Good job, but don't forget to deploy

PR #1074, opened 2024-01-04
New pr, previous deleted by accident

Timi2424:
[DEMO LINK](https://Timi2424.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import { FC, useEffect, useState } from 'react';
 import { Todo } from './types/Todo';
 import { getTodos } from './api/todos';
 import { Footer } from './components/Footer/Footer';
@@ -12,7 +12,7 @@ import { Error } from './components/Error/Error';
 
 const USER_ID = 12106;
 
-export const App: React.FC = () => {
+export const App: FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
   const [filterBy, setFilterBy] = useState(FilterBy.All);
@@ -27,7 +27,7 @@ export const App: React.FC = () => {
     () => {
       getTodos(USER_ID)
         .then(setTodos)
-        .catch(() => displayError(ErrorType.UnableToLoadTodo));
+        .catch(() => displayError(ErrorType.LOAD));
     }, [],
   );
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,45 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
+import { Footer } from './components/Footer/Footer';
+import { FilterBy } from './types/Filter';
+import { TodoList } from './components/TodoList/TodoList';
+import { filterTodo } from './components/TodoFilter';
 import { UserWarning } from './UserWarning';
+import { Header } from './components/Header/Header';
+import { ErrorType } from './types/Errors';
+import { Error } from './components/Error/Error';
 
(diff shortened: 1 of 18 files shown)


mbruhler (changes requested):
sync

src/App.tsx
```diff
@@ -1,10 +1,45 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
danon321, line 1: You dont need import React here. For example u can write 'FC' instead of 'React.FC'

src/App.tsx
```diff
@@ -1,10 +1,45 @@
...
 
-const USER_ID = 0;
+const USER_ID = 12106;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<ErrorType | null>(null);
+  const [filterBy, setFilterBy] = useState(FilterBy.All);
+
+  const displayError = (error: ErrorType) => {
```
danon321, line 20: This should be implemented in Error component. App should doesnt care about it

src/types/Errors.ts
```diff
@@ -0,0 +1,9 @@
+export enum ErrorType {
+  UnableToLoadTodo = 'Unable to load todos',
+  TitleIsEmpty = 'Title should not be empty',
+  UnableToAddTodo = 'Unable to add a todo',
+  UnableToDeleteTodo = 'Unable to delete a todo',
+  UnableToUpdateTodo = 'Unable to update a todo',
```
danon321, line 6: Names in enums have to be in upercase. I think 'DELETE' instead of 'UnableToDeleteTodo' should be better

src/App.tsx
```diff
@@ -14,191 +49,20 @@ export const App: React.FC = () => {
...
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
+        {todos.length > 0 && (
+          <Footer filterBy={filterBy} todos={todos} setFilterBy={setFilterBy} />
```
danon321, line 57: If u have cases when u have to set in props whole state ([filterBy, setFilterBy]) its time to think of some other solutions. Maybe context?

danon321 (changes requested):
Im not sure if this should be done in this part of the task. But i cannot add single todo. :(

src/App.tsx
Timi2424: yeah, I figured it out in the next part of the project :D

src/App.tsx
Timi2424: I know how to do it with context, but how do you suggest it to work without it. If I got what you meant of course

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+type Props = {
+  filterBy: FilterBy,
+  todos: Todo[],
+  setFilterBy: React.Dispatch<React.SetStateAction<FilterBy>>,
+};
+
+export const Footer: React.FC<Props> = ((props) => {
+  const { filterBy, todos, setFilterBy } = props;
+
+  const countUncompletedTodo = todos.filter(todo => !todo.completed).length;
```
MarcinLigmann, line 15: you can also use useMemo() hook to prevent overcomputing, something like:
```
const countUncompletedTodo = useMemo(() => {
  return todos.filter(todo => !todo.completed).length;
}, [todos]);
```

MarcinLigmann (approved):
It looks very good, the comment below is not necessary for this part of the application, but it is worth considering

src/components/Footer/Footer.tsx
Timi2424: thank you for your insight, I'm gonna apply it if necessary on later stages of this project

mbruhler (approved):
good job, approved

PR #1076, opened 2024-01-04
todo-app-loading-todos first commit

baraban2003:
[DEMO LINK](https://baraban2003.github.io/react_todo-app-loading-todos/)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+              case Filter.completed:
+                return todo.completed;
+
+              case Filter.all:
+                return todo;
+
+              default:
+                return true;
+            }
+          })
```
etojeDenys, line 30: filter todos before jsx and just render them in jsx.

etojeDenys (approved):
only one small comment, good job

src/components/TodoList.tsx
baraban2003: Yes, right. Tnx!

PR #1078, opened 2024-01-07
add task solution

denlubn:
https://denlubn.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,46 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,10 +1,46 @@
...
+      case StatusFilter.ACTIVE:
+        return !todo.completed;
+
+      case StatusFilter.COMPLETED:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  }), [todos, statusFilter]);
```
SanyaBratashchuk, line 42: use filter inside switch case not vice versa to avoid extra iteration in default case

SanyaBratashchuk (approved):

PR #1094, opened 2024-01-24
Develop

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_todo-app-loading-todos/)


src/types/ErrorMessage.ts
```diff
@@ -0,0 +1,8 @@
+export enum ErrorMessage {
+  NULL = '',
```
sTorba24, line 2: Why do you need this value? Just use empty string
Comparing with empty string could be replaced with !string or just string( in conditions)

src/Components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+        {todos.length > 0 && (
+          <div>
+            <Main />
+            <Footer />
+          </div>
+        )}
+      </div>
+
+      {errorMessage !== ErrorMessage.NULL && <Notification />}
```
sTorba24, line 29:
```suggestion
      {errorMessage.length > 0 && <Notification />}
```

sTorba24 (changes requested):
Almost good!

selected tab is `completed`
image

sTorba24 (approved):

PR #1108, opened 2024-01-31
add solution

nabokok:
[DEMO LINK](https://nabokok.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,54 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { getFilteredTodos } from './services/getFilteredTodos';
 
-const USER_ID = 0;
+const USER_ID = 87;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<Status>(Status.All);
+
+  const filteredTodos = getFilteredTodos(filter, todos);
+  const activeTodosCount = todos.reduce((acc, cur) => {
+    if (!cur.completed) {
+      return acc + 1;
+    }
+
+    return acc;
+  }, 0);
+
+  useEffect(() => {
+    let interval: NodeJS.Timeout;
+
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Error');
+        interval = setInterval(() => {
+          setErrorMessage('');
+          clearInterval(interval);
+        }, 3000);
+      });
+
+    return () => {
+      clearInterval(interval);
+    };
+  }, []);
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+  const [filter, setFilter] = useState<Status>(Status.All);
+
+  const filteredTodos = getFilteredTodos(filter, todos);
+  const activeTodosCount = todos.reduce((acc, cur) => {
+    if (!cur.completed) {
+      return acc + 1;
+    }
+
+    return acc;
+  }, 0);
```
Xilston, line 24: In this case you can use filter with todo.completed & lenght property. Not necessary.

src/App.tsx
```diff
@@ -33,171 +77,124 @@ export const App: React.FC = () => {
...
       >
-        <button data-cy="HideErrorButton" type="button" className="delete" />
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={handleCloseError}
+        />
         {/* show only one message at a time */}
         Unable to load todos
```
Xilston, line 189: Try to show `{errorMessage}` there, in next step of this task.

Xilston (approved):
Love your code! Nice one!

src/App.tsx
```diff
@@ -1,10 +1,54 @@
...
+    let timeout: NodeJS.Timeout;
+
+    getTodos(USER_ID)
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Error');
+        timeout = setTimeout(() => {
+          setErrorMessage('');
+          clearInterval(timeout);
+        }, 3000);
```
olya-shyian, line 36: why do you use clearInterval if you did't use setInterval?

I think we can do it in more simple way, try to use `wait` from `utils/fetchClient.ts`

for example,
```
.catch(() => {
  setErrorMessage('Error');
  return wait(3000).then(() =>  setErrorMessage(''));
});
 ```
 
 but the best option to create a separate component for ErrorMessage, and this setTimeout and clearing of this Timeout should be there

File: src/App.tsx [hunk 4]
```diff
@@ -33,171 +77,124 @@ export const App: React.FC = () => {
...
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
 
-            <a
-              href="#/completed"
-              className="filter__link"
-              data-cy="FilterLinkCompleted"
+            {/* Active filter should have a 'selected' class */}
```
olya-shyian, line 127: you can delete this commented code

src/App.tsx
```diff
@@ -33,171 +77,124 @@ export const App: React.FC = () => {
...
+                className={cn('filter__link',
+                  { selected: filter === Status.Completed })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => handleFilter(Status.Completed)}
+              >
+                {Status.Completed}
+              </a>
+            </nav>
+
+            {/* don't show this button if there are no completed todos */}
```
olya-shyian, line 160: and here

plese check everywhere

src/App.tsx
olya-shyian: agree with `Xilston` student

olya-shyian (changes requested):
Great!🙂

YegorKorenkov (approved):
Nice job! Well done!

PR #1113, opened 2024-01-31
add task solution

artem8746:
[DEMO LINK](https://artem8746.github.io/react_todo-app-loading-todos/)


src/components/TodoApp/TodoApp.tsx
```diff
@@ -0,0 +1,168 @@
...
+                onClick={handleCleanCompletedButtonClicked}
+                type="button"
+                className="todoapp__clear-completed"
+                data-cy="ClearCompletedButton"
+              >
+                Clear completed
+              </button>
+            </footer>
+          )}
+      </div>
```
olya-shyian, line 158: please create components for `header` and `footer`

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { TodoItem } from '../TodoItem';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todos: Todo[]
+}
+
+export const TodoList: React.FC<Props> = React.memo(
+  ({ todos }) => {
+    return (
```
olya-shyian, line 11: if there is no logic in component and you have only jsx, then you can return without using word `return`

`React.memo(({ todos }) => (`

src/types/TodoAction.ts
```diff
@@ -0,0 +1,9 @@
+export enum TodoAction {
+  setError = 'setError',
+  SetTodos = 'setTodos',
+  SetFilterOptions = 'setFilterOptions',
```
olya-shyian, line 4: please update `setError` to `SetError`

Upper case first letter

olya-shyian (approved):
Thanks for the updates!✨

PR #1116, opened 2024-02-01
add solution

OleksandrRezanov:
[DEMO LINK](https://OleksandrRezanov.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/services/getVisibleTodos.ts
+++ b/src/services/getVisibleTodos.ts
@@ -1,0 +1,18 @@
+import { FilterBy } from '../types/FilterBy';
+import { Todo } from '../types/Todo';
+
+export function getVisibleTodos(todos: Todo[], query: FilterBy) {
+  return [...todos].filter(todo => {
+    switch (query) {
+      case 'active':
+        return !todo.completed;
+
+      case 'completed':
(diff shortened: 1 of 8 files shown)


kostrubitskii (approved):
Good job! Enable your test and check your code)

Pls read README and enable your test !
## If you want to enable tests
- open `cypress/integration/page.spec.js`
- replace `describe.skip` with `describe` for the root `describe`

src/services/getVisibleTodos.ts
```diff
@@ -0,0 +1,18 @@
+import { FilterBy } from '../types/FilterBy';
+import { Todo } from '../types/Todo';
+
+export function getVisibleTodos(todos: Todo[], query: FilterBy) {
+  return [...todos].filter(todo => {
```
olya-shyian, line 5: there is no need to do a copy of todos, because filter method does not mutate

olya-shyian (approved):
Good!🙂
There is some bug, open the demo link and do refresh of the page, during some seconds you will see the error meassage. It should not work like that

I'll approve. In order to save your time you can do updates according to my comments in the next task

PR #1121, opened 2024-02-01
add solution

ValeriaMyronets:
[DEMO LINK](https://ValeriaMyronets.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,0 +1,98 @@
+/* eslint-disable @typescript-eslint/indent */
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useContext, useState } from 'react';
+import { TodosContext } from '../TodosContext';
+import { USER_ID } from '../constants';
+import { createTodo } from '../api/todos';
+import { Todo } from '../types/Todo';
+
+export const getNextId = (todos: Todo[]) => {
+  const ids = todos.map((todo) => todo.id);
+
+  return Math.max(...ids) + 1;
+};
+
+export const Header:React.FC = () => {
+  const [title, setTitle] = useState('');
+
+  const {
+    todos,
+    setTodos,
+    setErrorMessage,
+  } = useContext(TodosContext);
+
+  const uncompletedTodos = todos.filter(todo => !todo.completed).map(t => t.id);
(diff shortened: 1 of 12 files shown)


src/components/Header.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { TodosContext } from '../TodosContext';
+import { USER_ID } from '../constants';
+import { createTodo } from '../api/todos';
+import { Todo } from '../types/Todo';
+
+export const getNextId = (todos: Todo[]) => {
+  const ids = todos.map((todo) => todo.id);
+
+  return Math.max(...ids) + 1;
+};
```
AntonDuchenko, line 13: You can remove this function already, bc in this task we don't need to create an id for todo

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+    setTodos((prev) => prev.filter((t) => t.id !== todoId));
+  };
+
+  return (
+    <div
+      data-cy="Todo"
+      className={classNames(
+        'todo',
+        // eslint-disable-next-line quote-props
+        { 'completed': completed },
```
AntonDuchenko, line 37: better write it like in the example below, and remove eslint-disable, it means the same as what you write

```suggestion
      className={classNames(
        'todo',
        { completed },
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,29 @@
...
+    switch (currentFiltering) {
+      case Filtering.ALL:
+        return todos;
+      case Filtering.ACTIVE:
+        return todos.filter((t) => !t.completed);
+      case Filtering.COMPLETED:
+        return todos.filter((t) => t.completed);
+      default:
+        return todos;
+    }
```
AntonDuchenko, line 19: You can remove `case Filtering.ALL` from this switch, bc in default you have a duplicate, and it will work the same

AntonDuchenko (approved):
Awesome Job! Keep going! :thumbsup: :muscle: :fire:
But check a few recommendations

olya-shyian (approved):
I will approve without code review )
It will be done in the third part of the task

PR #1125, opened 2024-02-02
solution

pivkopa:
[DEMO LINK](https://pivkopa.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,66 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
+import * as api from './api/todos';
+import { Todo } from './types/Todo';
 
-const USER_ID = 0;
+const USER_ID = 104;
 
 export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,28 +1,79 @@
...
+          return todo.completed === false;
+        });
+      case 'completed':
+        return todos.filter((todo) => {
+          return todo.completed === true;
+        });
+      case 'all':
+        return todos;
+      default:
+        return todos;
```
Manankin, line 50: same result for both cases. I think you can remove case 'all'

src/App.tsx
```diff
@@ -1,28 +1,79 @@
...
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState('');
+
+  const existingCompeted = todos.some((value) => {
+    return value.completed;
+  });
```
Manankin, line 17: where do you use this variable?

Manankin (approved):
Good job!!
Please fix this two cases before mentors checking

PR #1127, opened 2024-02-03
task done

Manankin:
[DEMO LINK](https://Manankin.github.io/react_todo-app-loading-todos/)


src/components/TodoList.tsx
```diff
@@ -0,0 +1,18 @@
...
+// import classNames from 'classnames';
+import { Todo } from '../types/Todo';
+import { TodoItem } from './TodoItem';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
```
olya-shyian, line 11: when you have no logic in component and just return jsx you can delete the word `return ` and just return jsx in round brackets

`}) => (`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+// import classNames from 'classnames';
```
olya-shyian, line 2: commented code

src/types/Errors.ts
```diff
@@ -0,0 +1,7 @@
+export enum Errors {
+  LoadError = 'Unable to load todos',
+  EmptyTitle = 'Title should not be empty',
+  AddingError = 'Unable to add a todo',
+  DeletingError = 'Unable to delete a todo',
+  UpdatingError = 'Unable to update a todo',
```
olya-shyian, line 6: great solution!)

olya-shyian (approved):
Nicely done!💥

I'll approve. In order to save your time you can do updates according to my comments in the next task

src/components/TodoList.tsx
Manankin: Yes, you are right. I always write return because have some problem with syntax and this way help me.

src/types/Errors.ts
Manankin: Thank you.

Manankin:
I have fixed all comments. Thank you a lot ))

PR #1136, opened 2024-02-06
Add task solution

PodvAx:
[DEMO LINK](https://PodvAx.github.io/react_todo-app-loading-todos/) 
.
.
.
There are problem with tests in filtering part, but on page all work correct. Also in task says that we need only complete the error processing, filtering and loading, so I didn't pay attention on other aspects except tests.

### DIFF
--- a/src/components/ErrorNotification.tsx
+++ b/src/components/ErrorNotification.tsx
@@ -1,0 +1,62 @@
+import React, { useContext, useEffect, useState } from 'react';
+import { ErrorContext } from '../providers/TodosProvider';
+
+type Props = {};
+
+export const ErrorNotification: React.FC<Props> = () => {
+  const { errorMessage, setErrorMessage } = useContext(ErrorContext);
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    let timerId: NodeJS.Timeout | null;
+
+    if (errorMessage) {
+      setIsVisible(true);
+      timerId = setTimeout(() => {
+        setIsVisible(false);
+      }, 3000);
+    }
+
+    return () => {
+      if (timerId) {
+        clearTimeout(timerId);
+      }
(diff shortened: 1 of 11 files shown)


src/components/TodoItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+      </label>
+
+      {/* This form is shown instead of the title and remove button */}
+      {/* form should be visible on doubleClick */}
+      {false ? (
```
volodymyr-soltys97, line 28: image

src/providers/TodosProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then((data) => {
+        setTodos(data);
+      })
```
etojeDenys, line 38:
```suggestion
      .then(setTodos)
```
we can simplfy it

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useContext, useEffect, useState } from 'react';
+import classNames from 'classnames';
+import { ErrorContext } from '../providers/TodosProvider';
+
+type Props = {};
+
+export const ErrorNotification: React.FC<Props> = () => {
```
etojeDenys, line 7:
```suggestion
export const ErrorNotification: React.FC = () => {
```

etojeDenys (approved):
good job

PR #1145, opened 2024-02-12
Solution

VKonoplianyi:
[DEMO LINK](https://VKonoplianyi.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,50 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,9 +1,50 @@
...
+    switch (filterStatus) {
+      case Status.Active:
+        return !todo.completed;
+      case Status.All:
+        return true;
+      case Status.Completed:
+        return todo.completed;
+      default:
+        return true;
+    }
```
volodymyr-soltys97, line 38:
```suggestion
    switch (filterStatus) {
      case Status.Active:
        return !todo.completed;

      case Status.Completed:
        return todo.completed;
        
      default:
        return true;
    }
```

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1146, opened 2024-02-13
add solution

ptbit:
[DEMO LINK](https://ptbit.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -1,0 +1,40 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+import { TodosFilter } from './TodosFilter';
+
+type Props = {
+  unCompletedTodos: number,
+  filter: Status,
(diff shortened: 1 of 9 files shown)


src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,40 @@
...
+  unCompletedTodos,
+  filter,
+  setFilter,
+  todos,
+  removeCompletedTodos,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${unCompletedTodos} items left`}
```
l4st1m0za, line 23: It's better to keep the UI text related to pluralization grammatically correct. For instance, '1 item left' instead of '1 items left'. You can include a condition to check if 'unCompletedTodos' is 1 and accordingly display 'item' or 'items'.

l4st1m0za (approved):
Good job!

PR #1148, opened 2024-02-15
react_todo-app-loading-todos

AnoriFrell:
[DEMO LINK](https://AnoriFrell.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,7 +65,7 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <Header />
         <TodoList todos={filteredTodos} />
-        {todos.length > 0 && (
+        {!!todos.length && (
           <Footer
             filterTodos={setFilter}
             currentFilter={filter}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,50 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { Error } from './types/Error';
+import { ErrorMessage } from './components/ErrorMessage';
+import { getTodos } from './api/todos';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import './styles/index.scss';
+
+const USER_ID = 27;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [error, setError] = useState<Error | null>(null);
+
+  useEffect(() => {
+    getTodos(USER_ID)
+      .then(res => {
+        setTodos(res);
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,204 +1,68 @@
...
+          return todo;
+        case Filter.Completed:
+          return todo.completed;
+        case Filter.Active:
+          return !todo.completed;
+        default:
+          return todo;
+      }
+    });
+  }, []);
```
DarkMistyRoom, line 42: looks like we can move this function out of the component

src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  isCompletedTodos: boolean,
+}
+
+export const Footer: React.FC<Props> = (
+  {
+    filterTodos,
+    currentFilter,
+    isCompletedTodos,
+  },
+) => (
```
DarkMistyRoom, line 16:
```suggestion
export const Footer: React.FC<Props> = ({
  filterTodos,
  currentFilter,
  isCompletedTodos,
}) => (
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+  <footer className="todoapp__footer" data-cy="Footer">
+    <span className="todo-count" data-cy="TodosCounter">
+      3 items left
+    </span>
```
DarkMistyRoom, line 20: I think you can count the real number of todos on this stage

src/components/TodoList.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+          <button type="button" className="todo__remove" data-cy="TodoDelete">
+            ×
+          </button>
+
+          <div data-cy="TodoLoader" className="modal overlay">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div>
+        </div>
```
DarkMistyRoom, line 38: Would be better to create `TodoItem` component

DarkMistyRoom (changes requested):
Good job! Let's improve it a bit.
Footer should be hidden when there are no todos

src/App.tsx
```diff
@@ -1,204 +1,81 @@
...
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
+        <Header />
+        <TodoList todos={filteredTodos} />
+        {todos.length > 0 && (
```
lerastarynets, line 68:
```suggestion
        {!!todos.length && (
```

src/components/TodoList.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(({ title, completed, id }) => (
+        <TodoItem
+          title={title}
+          completed={completed}
+          id={id}
```
lerastarynets, line 15: where is `key` prop?

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1160, opened 2024-02-29
add task solution p1

YevhenProtasov:
[DEMO LINK](https://YevhenProtasov.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,55 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import classNames from 'classnames';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import { wait } from './utils/fetchClient';
+import { Status } from './types/Status';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [search, setSearch] = useState<string>('');
+  const [sortBy, setSortBy] = useState<Status>(Status.All);
+
+  const titleInput = useRef<HTMLInputElement>(null);
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,9 +1,55 @@
...
+
+  const titleInput = useRef<HTMLInputElement>(null);
+
+  const isAllCompleted = todos.every(todo => todo.completed);
+
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+
+  let visibleTodos = todos.filter(todo =>
+    todo.title.toLowerCase().includes(search.toLowerCase()),
+  );
```
TarasHoliuk, line 25: input shouldn't work as a search
in the next stage, we will use it as input to add new todo

src/App.tsx
```diff
@@ -1,9 +1,55 @@
...
+  useEffect(() => {
+    if (titleInput.current) {
+      titleInput.current.focus();
+    }
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'))
+      .finally(() => {
+        wait(3000).then(() => setErrorMessage(''));
```
TarasHoliuk, line 49: it's unnecessary action in every `finally`

better to create useEffect with timeout (and check - if errorMessage is not empty - set it empty)
One more variant is to create a function that sets some error **and** launch timeout where error will be cleared. And call this function in `catch`

src/App.tsx
TarasHoliuk: I know that you know it) but it looks a bit confusing

TarasHoliuk (changes requested):
GJ :+1: 

Added 2 small comments :slightly_smiling_face:

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <input
+              data-cy="TodoStatus"
+              type="checkbox"
+              className="todo__status"
+              checked={todo.completed}
+            />
+          </label>
+
+          <span data-cy="TodoTitle" className="todo__title">
+            {todo.title}
```
volodymyr-soltys97, line 27: You can use destruction for `todo`

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1161, opened 2024-03-01
react_todo-app-loading-todos

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,119 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useRef } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { wait } from './utils/fetchClient';
+import { ErrorMessages } from './types/ErrorMessages';
+import { Todo } from './types/Todo';
+import { Status } from './types/FilterStatus';
+import * as postService from './api/todos';
+import { getFilteredTodos } from './services/getFilteredTodos';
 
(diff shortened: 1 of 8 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
It is a bad practice to write all the logic in one component, try to split the logic into several different components
It is also important to follow the same principle of creating functions everywhere (function expression or function declaration), now it is different

src/services/TodoList.tsx
```diff
@@ -0,0 +1,66 @@
...
+              <div className="loader" />
+            </div>
+          </div>
+        ))}
+
+        <div data-cy="TodoLoader" className="modal overlay">
+          <div className="modal-background has-background-white-ter" />
+          <div className="loader" />
+        </div>
+      </div>
```
vadiimvooo, line 48: This can be also used as another component - TodoItem

src/App.tsx
```diff
@@ -15,188 +63,102 @@ export const App: React.FC = () => {
...
+                href="#/completed"
+                className={classNames('filter__link', {
+                  selected: sortBy === Status.Completed,
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setSortBy(Status.Completed)}
+              >
+                Completed
+              </a>
+            </nav>
```
vadiimvooo, line 130: Nav can be also used as another component - TodoFilter

vadiimvooo (changes requested):
Almost done, a few changes required.

vadiimvooo (approved):
Good job!)

PR #1170, opened 2024-03-11
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_todo-app-loading-todos/)

Don't mind editing and toggle parts as I'm still working on them. I wanted to do everything in the first half, so there's more functionality than needed. Consider only loading data.


lerastarynets (approved):
You're on the right track!

PR #1176, opened 2024-03-15
finished 1 task

HolyRaveOS:
[DEMO LINK](https://HolyRaveOS.github.io/react_todo-app-loading-todos/)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,33 @@
...
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={handleRemoveError}
+      />
+      {todosError}
+      {/* Unable to load todos
+      <br />
+      Title should not be empty
+      <br />
```
polosanya, line 25: Don't forget to remove the comments

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }, []);
+
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      <button
+        aria-label="toggle all active todos"
+        type="button"
+        className="todoapp__toggle-all active"
```
polosanya, line 18: Would be great to implement this comment

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
+import { useTodos } from '../../context/TodosContext';
+import { TodoItem } from '../TodoItem';
+
+export const TodoList: React.FC = () => {
+  const { todos, filterStatus } = useTodos();
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterStatus) {
+      case 'Active':
```
polosanya, line 9: Better use Enum instead of the hardcoded string

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments

src/components/TodoList/TodoList.tsx
HolyRaveOS: I Using Enum here. My switch get filterStatus as parameter and this  filterStatus is type FilterStatus = keyof typeof Status;

So if I type in my Switch case key that doesn't exist in my Status keys I will get an error.
```
export enum Status {
  All = '#/',
  Active = '#/active',
  Completed = '#/completed',
}

PR #1178, opened 2024-03-17
task todosServer-1

Ihor-Prodan:
[DEMO LINK](https://Ihor-Prodan.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Hi, your demo link doesn't work, you need to make a deploy
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
When I choose completed, the entire footer hides and I can't switch to another, this needs to be fixed
image

Need to hide this text when I write only spaces to the input.
Also, there is no way to write text in the input, it needs to be fixed
image

src/components/TodoList.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+      {/* This todo is being edited */}
+      {/* <div data-cy="Todo" className="todo">
+        <label className="todo__status-label">
+          <input
+            data-cy="TodoStatus"
+            type="checkbox"
+            className="todo__status"
+          />
+        </label>
```
DarkMistyRoom, line 47: remove the comments

src/components/TodosFilter.tsx
```diff
@@ -0,0 +1,22 @@
...
+    case Status.All:
+      return todos;
+    case Status.Active:
+      return allTodos.filter(todo => !todo.completed);
+    case Status.Completed:
+      return allTodos.filter(todo => todo.completed);
+    default:
+      return allTodos;
+  }
+};
```
DarkMistyRoom, line 22: it is not a component. it is a util function - move it to a separate folder

DarkMistyRoom (changes requested):

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
image

l4st1m0za (approved):
Good job!

PR #1202, opened 2024-03-28
solution

ChrisHryts:
https://chrishryts.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,203 +1,57 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useState, useEffect, useContext } from 'react';
+import { getTodos } from './api/todos';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Main } from './components/Main';
+import { ErrorContext, TodosContext } from './components/TodoContext';
+import { Status } from './types/Status';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const { list, setList } = useContext(TodosContext);
+  const { errorMessage, setErrorMessage } = useContext(ErrorContext);
+  const [query, setQuery] = useState<Status>(Status.All);
+
+  useEffect(() => {
+    getTodos()
+      .then(setList)
+      .catch(() => setErrorMessage('Unable to load todos'))
+      .finally(() => {
+        setTimeout(() => setErrorMessage(''), 3000);
+      });
+  }, [setErrorMessage, setList]);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
-
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
-
-        <section className="todoapp__main" data-cy="TodoList">
-          {/* This is a completed todo */}
-          <div data-cy="Todo" className="todo completed">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-                checked
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Completed Todo
-            </span>
-
-            {/* Remove button appears only on hover */}
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
-
-            {/* overlay will cover the todo while it is being deleted or updated */}
-            <div data-cy="TodoLoader" className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is an active todo */}
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Not Completed Todo
-            </span>
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
-
-            <div data-cy="TodoLoader" className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is being edited */}
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            {/* This form is shown instead of the title and remove button */}
-            <form>
-              <input
-                data-cy="TodoTitleField"
-                type="text"
-                className="todo__title-field"
-                placeholder="Empty todo will be deleted"
-                value="Todo is being edited now"
-              />
-            </form>
-
-            <div data-cy="TodoLoader" className="modal overlay">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-
-          {/* This todo is in loadind state */}
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Todo is being saved now
-            </span>
-
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-        </section>
-
-        {/* Hide the footer if there are no todos */}
-        <footer className="todoapp__footer" data-cy="Footer">
-          <span className="todo-count" data-cy="TodosCounter">
-            3 items left
-          </span>
-
-          {/* Active link should have the 'selected' class */}
-          <nav className="filter" data-cy="Filter">
-            <a
-              href="#/"
-              className="filter__link selected"
-              data-cy="FilterLinkAll"
-            >
-              All
-            </a>
-
-            <a
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
-
-            <a
-              href="#/completed"
-              className="filter__link"
-              data-cy="FilterLinkCompleted"
-            >
-              Completed
-            </a>
-          </nav>
-
-          {/* this button should be disabled if there are no completed todos */}
-          <button
-            type="button"
-            className="todoapp__clear-completed"
-            data-cy="ClearCompletedButton"
-          >
-            Clear completed
-          </button>
-        </footer>
+        <Header />
+        <Main query={query} />
+        {list.length !== 0 && <Footer query={query} setQuery={setQuery} />}
       </div>
 
       {/* DON'T use conditional rendering to hide the notification */}
       {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
-        className="notification is-danger is-light has-text-weight-normal"
+        className={classNames(
+          'notification is-danger is-light has-text-weight-normal',
+          { hidden: !errorMessage },
+        )}
       >
-        <button data-cy="HideErrorButton" type="button" className="delete" />
-        {/* show only one message at a time */}
-        Unable to load todos
-        <br />
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setErrorMessage('')}
+        />
+        {errorMessage}
+        {/* Unable to load todos
         Title should not be empty
-        <br />
         Unable to add a todo
-        <br />
         Unable to delete a todo
-        <br />
-        Unable to update a todo
+        Unable to update a todo */}
       </div>
     </div>
   );
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,0 +1,82 @@
+import React, { useContext } from 'react';
+import classNames from 'classnames';
+import { Status } from '../types/Status';
+import { TodosContext } from './TodoContext';
+import { Todo } from '../types/Todo';
+import { deleteCompletedTodos } from '../api/todos';
+
+type Props = {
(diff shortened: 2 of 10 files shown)


src/App.tsx
```diff
@@ -1,203 +1,57 @@
...
+  const { list, setList } = useContext(TodosContext);
+  const { errorMessage, setErrorMessage } = useContext(ErrorContext);
+  const [query, setQuery] = useState<Status>(Status.All);
+
+  useEffect(() => {
+    getTodos()
+      .then(setList)
+      .catch(() => setErrorMessage('Unable to load todos'))
+      .finally(() => {
+        setTimeout(() => setErrorMessage(''), 3000);
```
Vasya564, line 20: Just use ``setErrorMessage('')`` because your TodosContext already handle errors

src/App.tsx
```diff
@@ -1,203 +1,57 @@
...
-        Unable to load todos
-        <br />
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => setErrorMessage('')}
+        />
+        {errorMessage}
+        {/* Unable to load todos
```
Vasya564, line 50: Create ``ErrorNotification`` component

src/types/Todo.ts
```diff
@@ -1,6 +1,6 @@
 export interface Todo {
   id: number;
-  userId: number;
+  userId?: number;
```
Vasya564, line 3: Todo should always have an ``userId``

src/components/Main.tsx
```diff
@@ -0,0 +1,150 @@
...
+              />
+            </form>
+          )}
+
+          {/* This todo is in loadind state */}
+          {/* 'is-active' class puts this modal on top of the todo */}
+          {/* <div data-cy="TodoLoader" className="modal overlay is-active">
+            <div className="modal-background has-background-white-ter" />
+            <div className="loader" />
+          </div> */}
```
Vasya564, line 139: Remove comments

Vasya564 (changes requested):
Good job! Let's improve code a little bit!

_Note: There is no need to implement other tasks that are not related to this task (adding, editing, updating, deleting todos). Because in future you need to change a lot of your code to pass the tests_ 😉

src/components/Main.tsx
ChrisHryts: need this for now

src/App.tsx
ChrisHryts: tests fail without it

Vasya564: Then it's better to create an error handler with error message as parameter which can show this error for 3s and close it. In future you can reuse this handler to display different errors in different components

src/App.tsx
```diff
@@ -1,204 +1,45 @@
...
+  TodosContext,
+} from './components/TodoContext';
+import { Status } from './types/Status';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const { list, setList } = useContext(TodosContext);
```
BudnikOleksii, line 15: - What `list` is this?
- Create a hooks to avoid passing context every time: `const useTodos = () => useContext(TodosContext);`

src/App.tsx
BudnikOleksii: > Just use `setErrorMessage('')` because your TodosContext already handle errors

Good point, create a handler that will set an error and a set timeout that will remove it in 3 seconds, because you should do the same in other cases in the future tasks

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+import React, { useContext } from 'react';
+import { ErrorContext } from './TodoContext';
+import classNames from 'classnames';
+
+type Props = {};
+
+export const ErrorNotification: React.FC<Props> = () => {
+  const { errorMessage, setErrorMessage } = useContext(ErrorContext);
```
BudnikOleksii, line 16:
```suggestion
export const ErrorNotification = () => {
  const { errorMessage, setErrorMessage } = useError();
```
- If you don't have props you don't need to specify the type
- Create hoooks for contexts as I mentioned previously

src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setErrorMessage('')}
+      />
+      {errorMessage}
+      {/* Unable to load todos
+        Title should not be empty
+        Unable to add a todo
+        Unable to delete a todo
```
BudnikOleksii, line 36: Remove all of them in all places

src/components/Footer.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+        <a
+          href="#/active"
+          className={classNames('filter__link', {
+            selected: query === Status.Active,
+          })}
+          data-cy="FilterLinkActive"
+          onClick={() => setQuery(Status.Active)}
+        >
+          Active
```
BudnikOleksii, line 56: You can create an array from your enum and map through it to create links

src/components/Main.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const handleTodoDelete = useCallback(
+    (todoId: number) => {
+      deleteTodos(todoId).catch(() =>
+        setErrorMessage('Unable to delete a todo'),
+      );
+
+      setList(currentTodos => {
+        return currentTodos.filter(currentTodo => currentTodo.id !== todoId);
+      });
```
BudnikOleksii, line 31: - You should remove todo from the todo list only if the request was successful 
- And you should show the loader to a user so they know that something happening. You can store todo ids that currently loading in an array, and depending on it show corresponding loaders for todos

The same for other handlers

src/components/Main.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const handleCheckbox = (todo: Todo) => {
+    const updatedList = list.map((item: Todo) => {
+      if (item.id === todo.id) {
+        return { ...item, completed: !item.completed };
+      }
+
+      return item;
+    });
+
+    setList(updatedList);
```
BudnikOleksii, line 68: When you need to change state depending on the previous one use functional setState

src/components/Main.tsx
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {listToGo.map((todo: Todo) => (
+        <div
+          data-cy="Todo"
```
BudnikOleksii, line 75: Create a todo component

src/components/Main.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <div data-cy="TodoLoader" className="modal overlay is-active">
+              <div className="modal-background has-background-white-ter" />
+              <div className={classNames({ loader: isLoading })} />
+            </div>
+          ) : (
+            <div data-cy="TodoLoader" className="modal overlay">
+              <div className="modal-background has-background-white-ter" />
+              <div className="loader" />
+            </div>
+          )}
```
BudnikOleksii, line 148: They are almost the same, just set classes that are different conditionally

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }, [errorMessage]);
+
+  return (
+    <LoadingContext.Provider value={loadingState}>
+      <ErrorContext.Provider value={errorState}>
+        <TodosContext.Provider value={todosState}>
+          {children}
+        </TodosContext.Provider>
+      </ErrorContext.Provider>
+    </LoadingContext.Provider>
```
BudnikOleksii, line 71: If you use them like this, there is not too much sense to create 3 contexts

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions, check the comments

src/App.tsx
ChrisHryts: todo list

src/App.tsx
BudnikOleksii: > > Just use `setErrorMessage('')` because your TodosContext already handle errors
> 
> Good point, create a handler that will set an error and a set timeout that will remove it in 3 seconds, because you should do the same in other cases in the future tasks

Not fixed

src/components/Footer.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { list, setList, setLoadingTodosIds, handleError } = useTodosContext();
+  const completedIds = list
+    .filter(item => item.completed === true)
+    .map(item => item.id);
+  const notCompletedQuantity = list.filter(
+    (todo: Todo) => !todo.completed,
+  ).length;
+
+  const handleClearCompleted = () => {
+    setLoadingTodosIds(completedIds);
```
BudnikOleksii, line 23: Add new loading todos to the previous one, use functional `setState` in case you have other todos loading at the same time.
And do the same in all other places where you set or remove loading todo ids

src/components/Footer.tsx
```diff
@@ -0,0 +1,90 @@
...
+      });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {notCompletedQuantity} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
```
BudnikOleksii, line 43: Remove comments

src/components/Footer.tsx
BudnikOleksii: > You can create an array from your enum and map through it to create links

Not fixed

src/components/Header.tsx
```diff
@@ -0,0 +1,115 @@
...
+    setTitle(event.target.value);
+    handleError('');
+  };
+
+  const toggleAllTodos = useCallback(() => {
+    const updatedList = list.map((todo: Todo) => ({
+      ...todo,
+      completed: !allCompleted,
+    }));
+
```
BudnikOleksii, line 84: Use functional `setState` to set todos depending on the previous one

src/App.tsx
BudnikOleksii: > todo list

If it's list of todos so name it like this or just `todos` so everyone can understand

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setList: (() => {}) as Dispatch<React.SetStateAction<Todo[]>>,
+  errorMessage: '',
+  setErrorMessage: (() => {}) as Dispatch<React.SetStateAction<string>>,
+  tempTodo: {} as Todo | null,
+  setTempTodo: (() => {}) as Dispatch<React.SetStateAction<Todo | null>>,
+  editingTodo: {} as Todo | null,
+  setEditingTodo: (() => {}) as Dispatch<React.SetStateAction<Todo | null>>,
+  handleError: (_message: string) => undefined,
+  loadingTodosIds: [] as number[],
+  setLoadingTodosIds: (() => {}) as Dispatch<React.SetStateAction<number[]>>,
```
BudnikOleksii, line 21: Don't use `as`, you can define types for the context
https://blog.logrocket.com/how-to-use-react-context-typescript/###

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,63 @@
...
+});
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const TodosProvider: React.FC<Props> = ({ children }) => {
+  const [list, setList] = useState<Todo[]>([]);
```
BudnikOleksii, line 29: Rename so everyone can understand what list is it

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const handleCheckbox = (checkedTodo: Todo) => {
+    const updatedList = list.map((item: Todo) => {
+      if (item.id === checkedTodo.id) {
+        return { ...item, completed: !item.completed };
+      }
+
+      return item;
+    });
+
+    setList(updatedList);
```
BudnikOleksii, line 33: Use functional `setState`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+          setList(currentTodos => {
+            return currentTodos.filter(
+              currentTodo => currentTodo.id !== todoId,
+            );
+          });
+          setLoadingTodosIds([]);
+        })
+        .catch(() => {
+          setErrorMessage('Unable to delete a todo');
+          setLoadingTodosIds([]);
```
BudnikOleksii, line 49: - Remove loaders in the finally block as you do it in both cases
- Use `handleError` function

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+        return { ...item, completed: !item.completed };
+      }
+
+      return item;
+    });
+
+    setList(updatedList);
+  };
+
+  const handleTodoDelete = useCallback(
```
BudnikOleksii, line 36: You don't need wrap with `useCallback` if you don't pass your function as in the child component

src/services/getFilteredItems.ts
```diff
@@ -0,0 +1,19 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const getFilteredItems = (list: Todo[], query: Status) => {
+  let filteredItems = [...list];
```
BudnikOleksii, line 5: `filter` doesn't mutate an array so you don't need to create a copy

BudnikOleksii (changes requested):
Good work 👍 
As I see you implemented some features from the next tasks. But a few more fixes are required, check the comments

src/components/TodoItem.tsx
ChrisHryts: dont have finally here at all

src/components/TodoItem.tsx
ChrisHryts: have this error without callback

image

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    deleteCompletedTodos(completedIds)
+      .then(() => {
+        setTodos(currentTodos => {
+          return currentTodos.filter(todo => !todo.completed);
+        });
+      })
+      .catch(() => handleError('Unable to delete completed todos'))
+      .finally(() => {
+        setLoadingTodosIds([]);
```
BudnikOleksii, line 39: You should remove only needed ids too, the same when you set them. Fix it in all places

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,155 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
BudnikOleksii, line 1: Don't disable linter

src/components/TodoItem.tsx
BudnikOleksii: > have this error without callback
> 
> image

I don't see what you have at line 77, but it doesn't make sense to wrap all with `useCallback`

src/components/TodoContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+      todos,
+      errorMessage,
+      handleError,
+      tempTodo,
+      editingTodo,
+      loadingTodosIds,
+      isInputFocused,
+      setIsInputFocused,
+    ],
+  );
```
BudnikOleksii, line 55: The same situation with `useMemo` here. on any change of values that you included in an array react do so many useless operations.
Wrap with `useMemo/useCallback` only what you actually need. Also, then you need to wrap components where you pass cached callback with `React.memo`
https://react.dev/reference/react/useCallback

BudnikOleksii (changes requested):
Nicely done 🚀 
Approved, but please, fix memoization in the next tasks, read how to use `useMemo/useCallback` properly

src/components/TodoItem.tsx
ChrisHryts: its used in the dependency array of another function

BudnikOleksii (approved):
Oh, I thought I approved this PR, sorry. Please, move to the next task, looks good 🚀

PR #1206, opened 2024-03-29
add task solution

kshepetska:
[DEMO LINK](https://kshepetska.github.io/react_todo-app-loading-todos/)

I accidentally started implementing tasks from the next part, hope it’s ok🙃

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,123 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import Header from './components/Header';
+import Footer from './components/Footer';
+import TodoList from './components/TodoList';
+import { Status } from './types/Status';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { wait } from './utils/fetchClient';
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        return wait(3000).then(() => setErrorMessage(''));
+      });
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const itemsLeft = todos.filter(({ completed }) => !completed).length;
+
+  const haveCompletedTodos = todos.some(({ completed }) => completed);
+
+  const allCompleted = todos.every(({ completed }) => completed);
+
+  const filteredTodos = todos.filter(task => {
+    switch (status) {
+      case Status.Active:
+        return !task.completed;
+      case Status.Completed:
+        return task.completed;
+      case Status.All:
+      default:
+        return true;
+    }
+  });
+
+  const addTodo = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (todo.trim() !== '') {
+      const newTodo: Todo = {
+        id: todos.length + 1,
+        userId: USER_ID,
+        title: todo,
+        completed: false,
+        status: status,
+      };
+
+      setTodos([...todos, newTodo]);
+      setTodo('');
+    }
+  };
+
+  const deleteTodo = (id: number) => {
+    const newTodos = todos.filter(todoItem => todoItem.id !== id);
+
+    setTodos(newTodos);
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setTodo(e.target.value);
+  };
+
+  const clearCompleted = () => {
+    const newTodos = todos.filter(todoItem => !todoItem.completed);
+
+    setTodos(newTodos);
+  };
+
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
-
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
-
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
+        <Header
+          isAllCompleted={allCompleted}
+          onAddTodo={addTodo}
+          handleInputChange={handleInputChange}
+          todosLength={todos.length}
+        />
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+  const [todo, setTodo] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
RusLan2186, line 23: you can create enum Errors and use it

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
-          </button>
-        </footer>
+          </>
+        )}
       </div>
-
-      {/* DON'T use conditional rendering to hide the notification */}
-      {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
```
RusLan2186, line 108: Create an ErrorNotification component

src/components/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {itemsLeft} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
```
RusLan2186, line 39: You can create a TodoFilter component

RusLan2186 (approved):
Good work 👍
Left some comments and suggestions, check them

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+        return task.completed;
+      case Status.All:
+      default:
+        return true;
+    }
+  });
+
+  const addTodo = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (todo.trim() !== '') {
```
StasSokolov1, line 53:
```suggestion
    if (!!todo.trim()) {
```

can be a bit simplify

src/components/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+    });
+  };
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {itemsLeft} items left
+      </span>
+
+      {/* Active link should have the 'selected' class */}
```
StasSokolov1, line 38: remove comments please

src/components/Footer.tsx
StasSokolov1: @kshepetska check this point please

src/components/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+      )}
+
+      {/* Add a todo on form submit */}
+      <form
+        onSubmit={e => {
+          e.preventDefault();
+          onAddTodo(e);
+          handleInputChange({
+            target: { value: '' },
+          } as React.ChangeEvent<HTMLInputElement>);
```
StasSokolov1, line 34: let's create separated function for this logic

StasSokolov1 (changes requested):
Almost done, let's improve your code a bit more🔥

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        return wait(3000).then(() => setErrorMessage(''));
```
BudnikOleksii, line 25: Create an error handler that will set an error message and then remove it in 3 seconds, as you have to do the same in future tasks

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+    if (!!todo.trim()) {
+      const newTodo: Todo = {
+        id: todos.length + 1,
+        userId: USER_ID,
+        title: todo,
+        completed: false,
+        status: status,
+      };
+
+      setTodos([...todos, newTodo]);
```
BudnikOleksii, line 62: When you need to set state depending on the previous one use functional `setState`

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+      setTodo('');
+    }
+  };
+
+  const deleteTodo = (id: number) => {
+    const newTodos = todos.filter(todoItem => todoItem.id !== id);
+
+    setTodos(newTodos);
```
BudnikOleksii, line 70: The same, and in all other places fix it

src/App.tsx
BudnikOleksii: > Create an ErrorNotification component

Good point

src/App.tsx
```diff
@@ -1,204 +1,123 @@
...
+          </>
+        )}
       </div>
-
-      {/* DON'T use conditional rendering to hide the notification */}
-      {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
-        className="notification is-danger is-light has-text-weight-normal"
+        className={`notification is-danger is-light has-text-weight-normal ${errorMessage ? '' : 'hidden'}`}
```
BudnikOleksii, line 109: Use classnames library to set classes conditionally

src/components/Filter.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+const Filter: React.FC<FilterProps> = ({ status, onStatusChange }) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      {Object.values(Status).map(value => (
+        <a
+          key={value}
+          href={`#${value}`}
+          className={`filter__link ${status === value ? 'selected' : ''}`}
```
BudnikOleksii, line 16: The same about classes, fix it in all places

src/components/Header.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onAddTodo,
+  isAllCompleted,
+  handleInputChange,
+  todosLength,
+}) => {
+  const handleClearInput = () => {
+    handleInputChange({
+      target: { value: '' },
+    } as React.ChangeEvent<HTMLInputElement>);
+  };
```
BudnikOleksii, line 20: I don't get what is the reason to pass an empty string as `target.value`, you need to change types to avoid it or this function do something else instead of clearing input

BudnikOleksii (changes requested):
Good progress, but some more fixes are required, check the comments

src/App.tsx
BudnikOleksii: > > Create an ErrorNotification component
> 
> Good point

Not fixed

src/components/Header.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isAllCompleted,
+  handleInputChange,
+  todosLength,
+}) => {
+  const handleClearInput = () => {
+    const event = {
+      target: { value: '' },
+    } as React.ChangeEvent<HTMLInputElement>;
+
+    handleInputChange(event);
```
BudnikOleksii, line 22: What is the point of doing it?

BudnikOleksii (approved):
Approved, but create an error notification component and fix the `handleClearInput` function in the next task

PR #1228, opened 2024-04-01
Develop

Dmitrel:
https://Dmitrel.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,50 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import classNames from 'classnames';
+import React, { useEffect, useState } from 'react';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,204 +1,75 @@
...
+    }
 
-            <span data-cy="TodoTitle" className="todo__title">
-              Not Completed Todo
-            </span>
-            <button type="button" className="todo__remove" data-cy="TodoDelete">
-              ×
-            </button>
+    return acc;
+  }, 0);
```
dffUqp, line 26: I think we can improve this logic. Let's filter out array first and then just `.length`

src/components/ErrorNotification/index.ts
```diff
@@ -0,0 +1 @@
+export * from '.';
```
dffUqp, line 1:
```suggestion
export * from './ErrorNotification';
```

dffUqp (approved):
LGTM 🐲

PR #1229, opened 2024-04-01
add first task solution

Graf6612:
https://graf6612.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,51 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { useState } from 'react';
+import { Todo } from './types/Todo';
+import Header from './components/header';
+import Main from './components/main';
+import Error from './components/Error';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [error, setError] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  const getFilteredTodos = () => {
+    switch (filter) {
+      case 'active':
+        return todos.filter(todo => !todo.completed);
+      case 'completed':
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  };
+
+  const tod = getFilteredTodos();
+
+  useEffect(() => {
+    setError(false);
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError(true);
+        setErrorMessage('Unable to load todos');
+        setTimeout(() => {
+          setError(false);
+        }, 3000);
+      });
+  }, []);
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+
+  const completedTodos = todos.filter(todo => todo.completed);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+      case 'active':
+        return todos.filter(todo => !todo.completed);
+      case 'completed':
+        return todos.filter(todo => todo.completed);
+      default:
+        return todos;
+    }
+  };
+
+  const tod = getFilteredTodos();
```
Vlobdam, line 30: rename it

Vlobdam (approved):
Looks good

src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [error, setError] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  const getFilteredTodos = () => {
+    switch (filter) {
+      case 'active':
+        return todos.filter(todo => !todo.completed);
+      case 'completed':
+        return todos.filter(todo => todo.completed);
```
StasSokolov1, line 24: let's use single `.filter` method and only change completed by status `active` or `completed`

src/App.tsx
StasSokolov1, line 30: let's use more informative name

src/App.tsx
```diff
@@ -14,191 +55,78 @@ export const App: React.FC = () => {
...
+                })}
+                data-cy="FilterLinkActive"
+                onClick={event => {
+                  event.preventDefault();
+                  setFilter('active');
+                }}
+              >
+                Active
+              </a>
+
```
StasSokolov1, line 98: use `.map` to display these elements, seems very similar

src/components/main.tsx
```diff
@@ -0,0 +1,120 @@
...
+import classNames from 'classnames';
+import React from 'react';
+
+export default function Main({ todos }) {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {/* This is a completed todo */}
+
+      {todos.map(todo => {
+        return (
```
StasSokolov1, line 11:
```suggestion
      {todos.map(todo => (
```

can be without return

StasSokolov1, line 11:
```suggestion
      {todos.map(todo => (
```

can be without return

StasSokolov1 (changes requested):
Good job, let's fix a few minor comments🔥

src/App.tsx
```diff
@@ -1,10 +1,52 @@
...
+    return todos.filter(todo => {
+      if (status === 'active') {
+        return !todo.completed;
+      } else if (status === 'completed') {
+        return todo.completed;
+      } else {
+        return true;
+      }
+    });
+  };
```
dffUqp, line 29: Let's create separate helpers file for this

src/App.tsx
```diff
@@ -1,10 +1,52 @@
...
+import { useState } from 'react';
+import { Todo } from './types/Todo';
+import Header from './components/header';
+import Main from './components/main';
+import Error from './components/Error';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
dffUqp, line 15: We can create enum or separate type for filter fields

src/App.tsx
dffUqp: Not fixed

src/App.tsx
```diff
@@ -14,191 +56,78 @@ export const App: React.FC = () => {
...
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
-        </section>
+        <Main todos={filteredTodos} />
```
dffUqp, line 61: I'm not sure that Main is a good name. 

Let's name it TodoList or something like this.

src/App.tsx
```diff
@@ -1,10 +1,52 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
dffUqp, line 2: I think we can fix these eslint errors

dffUqp (changes requested):
Almost done🙂

BudnikOleksii (approved):
Ok, if you fixed all comments in the next task I approving this one as well.

P.S. If you've already fixed all things, why you haven't added the code to this task?

PR #1249, opened 2024-04-11
solution 1 part

yurovych:
https://yurovych.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/components/todoFeilter.tsx
+++ b/src/components/todoFeilter.tsx
@@ -1,0 +1,64 @@
+import { useContext } from 'react';
+import { ActiveContext, AllContext, CompletedContext } from './filterContext';
+
+export const TodosFilter: React.FC = () => {
+  enum Status {
+    All = 'All',
+    Active = 'Active',
+    Completed = 'Completed',
+  }
+
+  const { isAllSelected, setIsAllSelected } = useContext(AllContext);
+  const { isActiveSelected, setIsActiveSelected } = useContext(ActiveContext);
+  const { isCompletedSelected, setIsCompletedSelected } =
+    useContext(CompletedContext);
+
+  const handleAllfilter = () => {
+    setIsAllSelected(true);
+    setIsActiveSelected(false);
+    setIsCompletedSelected(false);
+  };
+
+  const handleActivefilter = () => {
+    setIsActiveSelected(true);
+    setIsAllSelected(false);
+    setIsCompletedSelected(false);
+  };
+
+  const handleCompletedfilter = () => {
+    setIsCompletedSelected(true);
+    setIsAllSelected(false);
+    setIsActiveSelected(false);
+  };
+
+  return (
+    <>
+      <a
+        href="#/"
+        className={`filter__link ${isAllSelected === true ? 'selected' : ''}`}
+        data-cy="FilterLinkAll"
+        onClick={handleAllfilter}
+      >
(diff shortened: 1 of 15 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/components/todosList.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const TodosList: React.FC<ListProps> = ({ items }) => {
+  const { todos } = useContext(TodosContext);
+
+  const isListShown = () => {
+    return todos.length > 0;
+  };
+
+  return (
+    isListShown() === true && (
```
etojeDenys, line 18:
```suggestion
  return (
    todos.length > 0 && (
```
we can use it directly

etojeDenys: or like that
```suggestion
if(!todos.length){
 return null
}

  return (
```

src/components/todosItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  const isTodoCompletedClass = classNames({
+    todo: true,
+    completed: item.completed,
+  });
+
+  return (
+    <>
+      {toRender() === true && (
```
etojeDenys, line 60: this check looks redundant, you always render the TodoItem

src/components/todosItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const { isAllSelected } = useContext(AllContext);
+  const { isActiveSelected } = useContext(ActiveContext);
+  const { isCompletedSelected } = useContext(CompletedContext);
+  const { setIsChecked } = useContext(ManageCheckboxContext);
+
+  if (todos.every(element => element.completed === true)) {
+    setIsChecked(true);
+  } else {
+    setIsChecked(false);
+  }
```
etojeDenys, line 26: why do you need it?

src/components/todoFeilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+  enum Status {
+    All = 'All',
+    Active = 'Active',
+    Completed = 'Completed',
+  }
+
+  const { isAllSelected, setIsAllSelected } = useContext(AllContext);
+  const { isActiveSelected, setIsActiveSelected } = useContext(ActiveContext);
+  const { isCompletedSelected, setIsCompletedSelected } =
+    useContext(CompletedContext);
```
etojeDenys, line 14: it can be only one StatusContext or just one status state that will save the enum Status values

etojeDenys (changes requested):
your todos don't have the border, try to fix that
image



don't hesitate to ask for some help in the chat with that

src/components/todosItem.tsx
yurovych: No, there are fiters , and depends on it we decide which 'item' shoud be rendered or not

src/components/todosItem.tsx
yurovych: for button 'toggle all'

src/components/todoFeilter.tsx
yurovych: there is no any commont bond between those Contexts and enum
contexts just toggle (true or false) for out fiilters.

SanyaBratashchuk (approved):

PR #1261, opened 2024-04-18
solution loading todos

Anothar:
[DEMO LINK](https://Anothar.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,116 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useContext, useEffect, useRef } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos, postTodos } from './api/todos';
+import { Action, DispatchContext, StateContext } from './store/store';
+import cn from 'classnames';
+import classNames from 'classnames';
+
+const downloadAllTodos = async (dispatch: React.Dispatch<Action>) => {
+  try {
+    const allTodos = await getTodos();
+
+    dispatch({ type: 'setAllTodos', todos: allTodos });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    dispatch({ type: 'setError', error: 'Unable to load todos' });
+  }
+};
 
 export const App: React.FC = () => {
+  // #region
+  const dispatch = useContext(DispatchContext);
+  const {
+    todos,
+    newTodo,
+    focusNewTodo,
+    useTodos,
+    changerId,
+    idTodoSubmitting,
+    errorsInTodo,
+  } = useContext(StateContext);
+
+  useEffect(() => {
+    downloadAllTodos(dispatch);
+  }, [dispatch]);
+
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+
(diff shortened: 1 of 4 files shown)


Anothar:
I rushed and completed a large portion of the app that I was supposed to work on in the upcoming parts, and realized it very late.

anastasiia-tilikina (changes requested):
GJ!

things we should improve:
1. divide your code into smaller components, it is not a good idea to make one huge component, besides in this case global context doesn't make any sense.
2. In the next parts you should save all the changes on server such as completing, editing and deleting todos.
3. No need to keep everything in context. You should store there only values you need globally, such as todos, filter status etc. On the other hand - new title is not a global thing, you can store it locally. for focusing on input you can avoid to create some variable at all.
4. some of your action names are not clear. for instance I'm not familiar with your code and such action as 'add' is not clear . what does it add? or 'changed', Some of them have similar names such as 'changed' and 'setChanged' it's not clear what is the difference. Please read it one more time and try to use more self descriptive names. For example 'addTodo' 'removeTodo' 'completeAllTodos'.

src/App.tsx
```diff
@@ -1,203 +1,324 @@
...
+    if (useTodos === 'Completed') {
+      return todo.completed;
+    }
+
+    return true;
+  });
+
+  const addNewTodo = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    inputRef.current?.blur();
```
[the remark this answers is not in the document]
Anothar: I don't know why this works this way, but when this line is absent and there is no internet connection, pressing Enter doesn't focus.

src/component/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const Footer = () => {
+  const dispatch = useContext(DispatchContext);
+  const { todos, useTodos } = useContext(StateContext);
+
+  const todosLeft = todos.filter(todo => !todo.completed).length;
+
+  return (
+    <>
+      {todos.length !== 0 && (
```
volodymyr-soltys97, line 13:
```suggestion
      {!todos.length && (
```

src/component/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+              data-cy="FilterLinkAll"
+              onClick={() => dispatch({ type: 'setUseTodos', name: 'All' })}
+            >
+              All
+            </a>
+
+            <a
+              href="#/active"
+              className={cn('filter__link', {
+                selected: useTodos === 'Active',
```
volodymyr-soltys97, line 33: Create a enum for `'Active', 'All', 'Completed'` and use it everywhere

src/component/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,111 @@
...
+        return true;
+    }
+  });
+
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todosFilter.map(todo => (
+        <div
+          data-cy="Todo"
+          className={cn('todo', { completed: todo.completed })}
```
volodymyr-soltys97, line 28: Use destruction for `todo`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #1281, opened 2024-05-03
Solution

Aksahir:
[DEMO LINK](https://Aksahir.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Filter.tsx
+++ b/src/components/Filter.tsx
@@ -1,0 +1,90 @@
+import React from 'react';
+import { useContext } from 'react';
+import cn from 'classnames';
+import { TodoContext } from './TodoContext';
+import { FilterAction } from '../types/Actions';
+import { deleteTodo } from '../api/todos';
+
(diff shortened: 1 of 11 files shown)


l4st1m0za (changes requested):
Good job, but there is something wrong with the styles

Here is your version
image

Here is how it should be 
image

src/components/Filter.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const handleDeletCompleteTodos = () => {
+    for (const item of hasCompleteTodo) {
+      deleteTodo(item.id);
+    }
+  };
+
+  return (
+    <>
+      {todos.length > 0 && (
```
lerastarynets, line 30:
```suggestion
      {!!todos.length && (
```
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1289, opened 2024-05-07
Solution #1

kubsen:
[DEMO LINK](https://kubsen.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,6 +26,14 @@ export const App: React.FC = () => {
       });
   }, []);
 
+  const handleStatusChange = (newStatus: Status) => {
+    setStatus(newStatus);
+  };
+
+  const handleSetError = (errorMessage: string) => {
+    setError(errorMessage);
+  };
+
   const visibleTodos = getVisibleTodos(todos, status);
 
   if (!todosApi.USER_ID) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,191 +37,17 @@ export const App: React.FC = () => {
...
-          >
-            Clear completed
-          </button>
-        </footer>
+        <Header />
+
+        {!error && !!todos.length && (
+          <>
+            <TodoList visibleTodos={visibleTodos} />
+            <Footer todos={todos} status={status} setStatus={setStatus} />
```
nataliaklonowska, line 45: don't pass setStatus directly and instead create function to handle it. [here you can read more about why you shouldn’t pass react’s setState as a prop](https://medium.com/@christopherthai/why-you-shouldnt-pass-react-s-setstate-as-a-prop-a-deep-dive-8a3dcd74bec8)

src/App.tsx
```diff
@@ -14,191 +37,17 @@ export const App: React.FC = () => {
...
-        <br />
-        Title should not be empty
-        <br />
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <ErrorNotification error={error} setError={setError} />
```
nataliaklonowska, line 50: same problem

nataliaklonowska (changes requested):

src/App.tsx
```diff
@@ -1,11 +1,42 @@
...
+
+  const handleStatusChange = (newStatus: Status) => {
+    setStatus(newStatus);
+  };
+
+  const handleSetError = (errorMessage: string) => {
+    setError(errorMessage);
+  };
+
+  const visibleTodos = getVisibleTodos(todos, status);
```
nataliaklonowska, line 37: getVisibleTodos only if todos or status changes

nataliaklonowska (changes requested):

PR #1292, opened 2024-05-07
task solution

kinqbert:
[DEMO LINK](https://kinqbert.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,58 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+
+// TYPES
+import { Todo } from './types/Todo';
+import { FilterField } from './types/FilterField';
+
+import { TodoList } from './components/TodoList/TodoList';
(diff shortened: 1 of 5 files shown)


ArtemZakhar (approved):
Great job!)

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
undeletable, line 1: [Not critical]
Disabling eslint rules is not a big deal for a homework. But for a real work, I would not recommend to disable eslint rules if errors are produced on your code. These rules were most likely added for some reason, so it's better to fix your code to align it with rules. If this is not possible or you feel that a rule doesn't give enough profit, you can discuss changing its configuration or even removal with the team.

src/App.tsx
```diff
@@ -1,10 +1,39 @@
...
+    if (inputRef.current) {
+      inputRef.current.focus();
+    }
+
+    getTodos()
+      .then(result => dispatch({ type: ActionType.SetTodos, payload: result }))
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        setTimeout(() => {
```
undeletable, line 31: [Not critical]
It also makes sense to clear this timeout when effect is cleared:
```
useEffect(() => {
  // set timeout
  return () => {
    // clear timeour
  };
}, []);
```

undeletable (approved):
The solution looks fine! I just have a couple of notes on how it can be improved - please take a look.

PR #1297, opened 2024-05-09
Develop

KatsuboDmytro:
https://KatsuboDmytro.github.io/react_todo-app-loading-todos/


src/types/Todo.ts
```diff
@@ -1,4 +1,4 @@
-export interface Todo {
+export interface typeTodo {
```
TarasHoliuk, line 1:
```suggestion
export interface TypeTodo {
```
Check the linter error (do not ignore them - if you have questions it's better to ask in the chat) - types should be named using PascalCase

src/App.tsx
```diff
@@ -1,10 +1,42 @@
...
-import { USER_ID } from './api/todos';
+import { USER_ID, filterByStatus } from './api/todos';
+import { typeTodo } from './types/Todo';
+import { Todo } from './components/Todo/Todo';
+import { Loader } from './components/Loader/Loader';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterType, setFilterType] = useState('all');
```
TarasHoliuk, line 13: it may have 3 defined values - let's create enum for it

src/App.tsx
```diff
@@ -1,10 +1,42 @@
...
+    if (filter === 'all') {
+      status = 'all';
+    } else if (filter === 'active') {
+      status = false;
+    } else {
+      status = true;
+    }
+
+    setIsLoading(true);
+    filterByStatus(status)
```
TarasHoliuk, line 30: no need to fetch all todos every time you set filter
Fetch all todos once and save them in state. Then create a common const variable with filtered todos. This variable will be recalculated every time a component rerenders - it means after you change some state (including `filterStatus`)

src/App.tsx
```diff
@@ -1,10 +1,42 @@
...
+import { Loader } from './components/Loader/Loader';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterType, setFilterType] = useState('all');
+  const [todos, setTodos] = useState<typeTodo[]>([]);
+
+  const filterStatus = (filter: string) => {
+    let status: string | boolean = '';
```
TarasHoliuk, line 17: no need to set a boolean value, just handle different string values. But again - it's better to create enum for it

src/App.tsx
```diff
@@ -147,24 +86,33 @@ export const App: React.FC = () => {
           <nav className="filter" data-cy="Filter">
             <a
               href="#/"
-              className="filter__link selected"
+              className={`filter__link ${
+                'all' === filterType ? 'selected' : ''
```
TarasHoliuk, line 90: you can use classnames library for such cases

src/App.tsx
```diff
@@ -34,107 +66,14 @@ export const App: React.FC = () => {
...
+            ? <Loader />
+            : (
+              todos.map(todo => (
+                <Todo key={todo.id} todo={todo} />
+              ))
+            )
+          }
         </section>
 
         {/* Hide the footer if there are no todos */}
```
TarasHoliuk, line 79: As comment says:
image

You shouldn't display footer when there are no todos

Also, move Footer to a separate component

src/App.tsx
```diff
@@ -183,22 +131,24 @@ export const App: React.FC = () => {
...
-        <br />
-        Title should not be empty
-        <br />
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      {errorMessage && (
```
TarasHoliuk, line 134: image

src/App.tsx
```diff
@@ -147,24 +86,33 @@ export const App: React.FC = () => {
           <nav className="filter" data-cy="Filter">
```
TarasHoliuk, line 86: You have hardcoded value here:

image


Use actual value

src/App.tsx
```diff
@@ -34,107 +66,14 @@ export const App: React.FC = () => {
         </header>
```
TarasHoliuk, line 66: Let's move header to a separate component

TarasHoliuk (changes requested):
Pay attention to tests and linter errors. Feel free to ask for help in the chat if you need

anastasiiavorobiova (approved):
Great work!

PR #1300, opened 2024-05-10
add 1/3 solution

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,57 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { SortType } from './types/SortType';
+
+function getFilter(todos: Todo[], sortField: SortType) {
+  switch (sortField) {
+    case SortType.Active:
+      return todos.filter(todo => !todo.completed);
+    case SortType.Completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortField, setSortField] = useState(SortType.All);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const activeInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
+
+    setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+
+    if (activeInput.current) {
+      activeInput.current.focus();
+    }
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const sortedTodos = getFilter(todos, sortField);
+
+  const everyTodosCompleted = todos.every(todo => todo.completed);
+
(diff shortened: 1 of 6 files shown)


ArtemZakhar (approved):
Good job) 
But I believe that the header block should be moved to a separate component in further)
And "Clear completed" button should work differently. When there is nothing to clear, filter buttons have to be in the middle of the footer.

src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+function getFilter(todos: Todo[], sortField: SortType) {
+  switch (sortField) {
+    case SortType.Active:
+      return todos.filter(todo => !todo.completed);
+    case SortType.Completed:
+      return todos.filter(todo => todo.completed);
+    default:
+      return todos;
+  }
+}
```
DanilWeda, line 20: Move it outside component file

src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000);
```
DanilWeda, line 37: You can write handler for catch case

src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+      });
+
+    if (activeInput.current) {
+      activeInput.current.focus();
+    }
+  }, []);
+
+  const sortedTodos = getFilter(todos, sortField);
```
DanilWeda, line 45: Change naming ( gerFilter mean get filter elements for me), think better will be getFiltredTodos or something like that.

src/App.tsx
```diff
@@ -33,171 +74,24 @@ export const App: React.FC = () => {
...
-            type="button"
-            className="todoapp__clear-completed"
-            data-cy="ClearCompletedButton"
-          >
-            Clear completed
-          </button>
-        </footer>
+        <Todos todos={sortedTodos} />
+
+        {todos.length !== 0 && (
```
DanilWeda, line 79:
```suggestion
        {!!todos.length && (
```

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,64 @@
...
+  todos: Todo[];
+};
+
+export const Footer: React.FC<Props> = ({ sortField, setSortField, todos }) => {
+  const oneTodoCompleted = todos.find(todo => todo.completed);
+
+  const activeTodos = todos.reduce(
+    (acc, todo) => (todo.completed ? acc : acc + 1),
+    0,
+  );
```
DanilWeda, line 16: You can use one iteration in reduce and return object with completed and uncompleted fields. ( not critical)

DanilWeda (changes requested):
Good job. 
Check some comments for improvements. 
Many thanks)

yanamomot:
hello!
I wrote it like that and corrected it in the next task

src/App.tsx
```diff
@@ -1,10 +1,51 @@
...
+
+  const activeInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
+
+        setTimeout(() => {
```
undeletable, line 35: [Not critical]
It also makes sense to clear this timeout when effect is cleared:
```
useEffect(() => {
  // set timeout
  return () => {
    // clear timeout
  };
}, []);
```

undeletable (changes requested):
Your solution looks good. Just please take care of the comments from previous review - that should not be a big deal if you already have handled that in scope of other task.
Also I have another couple of non-critical comments - I recommend to take them into account during your further work.

DanilWeda (approved):
Approved! 
Thanks)

PR #1303, opened 2024-05-13
Solution

Allanpt:
[DEMO LINK](https://Allanpt.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,147 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, createTodo, getTodos, updateTodo } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+
+type SortSelectOptions = 'all' | 'active' | 'completed';
+export type ErrorMessages =
+  | ''
+  | 'Unable to load todos'
+  | 'Title should not be empty'
+  | 'Unable to add a todo'
+  | 'Unable to delete a todo'
+  | 'Unable to update a todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectSort, setSelectSort] = useState<SortSelectOptions>('all');
+  const [error, setError] = useState<ErrorMessages>('');
+  const [inputTodo, setInputTodo] = useState('');
+  const [loading, setLoading] = useState(false);
+
+  const todoList = todos.filter(todo => {
+    if (selectSort === 'active') {
+      return !todo.completed;
+    } else if (selectSort === 'completed') {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,149 @@
...
+
+  const todoList = todos.filter(todo => {
+    if (selectSort === 'active') {
+      return !todo.completed;
+    } else if (selectSort === 'completed') {
+      return todo.completed;
+    }
+
+    return true;
+  });
```
EdPirro, line 33: Wrap `todos.filter(...)` in a `useMemo` to avoid unecessary reprocessing

src/App.tsx
```diff
@@ -1,14 +1,149 @@
...
+      .finally(() => setLoading(false));
+  }, []);
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setError('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [error]);
```
EdPirro, line 52: Add conditional `if (error !== '')` to avoid creating timeout when error is already empty (will happen once every time setError is reset)

src/App.tsx
```diff
@@ -1,14 +1,149 @@
...
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
+    setError('');
+    if (event.key === 'Enter') {
```
EdPirro, line 60: Use early return instead of having your entire logic inside a conditional:

Instead of
```js
if(neededCondition) { 
    ...
}
```

Do:
```js
if(!neededCondition) return;
...
```

src/App.tsx
```diff
@@ -18,186 +153,101 @@ export const App: React.FC = () => {
...
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todos.length !== 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span className="todo-count" data-cy="TodosCounter">
+              {todos.filter(el => !el.completed).length} items left
```
EdPirro, line 188: `.reduce` would be a more meaningful array function to use here; this value could also be moved to a `useMemo` to prevent unnecessary reprocessing

src/components/TodoList.tsx
```diff
@@ -0,0 +1,195 @@
...
+        setTodos(
+          [...todos].map(todo => {
+            if (todo.id === todoSelect.id) {
+              return { ...todo, completed: updatedTodo.completed };
+            }
+
+            return todo;
+          }),
+        );
+      })
```
EdPirro, line 58: Use a callback to update the state `setTodos(last => [...last]...)`; this will prevent issues with asynchronous requests;
Try toggling two items as completed quickly (setting the second one before the first request is complete) and you'll see the issue: the first one will be shown as incomplete again even though it is completed in the backend

EdPirro (approved):
Great job, @Allanpt! Keep it up! 🚀

I pointed out a few issues but most of them are related to creation and update of items which is not in the scope of this exercise, so I'll approve it.

Do have a look at them before submitting the next parts of the project, though.

PR #1311, opened 2024-05-17
Develop

AYupko:
[DEMO LINK](https://AYupko.github.io/react_todo-app-loading-todos/)

Didn`t understand right away what exactly need to do now, so started from things, which should be done later. Sorry

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -1,0 +1,72 @@
+import cn from 'classnames';
+
+import { SortFields } from '../../types/sortFields';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  setSelectedField: (string: SortFields) => void;
+  selectedField: string;
+  todos: Todo[];
+};
+
+export const Footer: React.FC<Props> = ({
+  setSelectedField,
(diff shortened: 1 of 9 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  selectedField: string;
+  todos: Todo[];
+};
+
+export const Footer: React.FC<Props> = ({
+  setSelectedField,
+  selectedField,
+  todos,
+}) => {
+  const isAnyCompleted = todos.every(todo => !todo.completed);
```
alexander-ignatow, line 17: It is probably not a perfect naming here, but it is not critical

Better: `hasCompletedTodos`, etc

alexander-ignatow: And a slightly better way is :

```suggestion
  const hasCompletedTodos = !!todos.find(todo => todo.completed);
```

Your version needs to check each todo, while the other just searches for the first completed todo, so potentially it would not need to iterate through all todos

alexander-ignatow (approved):
Good job 👍 
Lets move on - see my comments for the future

PR #1321, opened 2024-05-20
add task solution

SlawikH:
[DEMO LINK](https://SlawikH.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  error: string;
+  setError: (errorMesage: string) => void;
+};
+
+export const Error: React.FC<Props> = ({ error, setError }) => {
+  return (
+    <div
+      data-cy="ErrorNotification"
+      className={classNames(
+        'notification is-danger is-light has-text-weight-normal',
+        {
+          hidden: !error,
+        },
+      )}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setError('')}
+      />
+      {error}
+    </div>
+  );
+};
--- a/src/components/ToDoItem/ToDoItem.tsx
+++ b/src/components/ToDoItem/ToDoItem.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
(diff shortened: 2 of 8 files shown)


dyakym (approved):
Good job)

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
```
undeletable, line 4: [Not critical]
It's better to name this type something like `ErrorProps`, so that the name describes this type the most precisely.

src/components/ToDoItem/ToDoItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+  todo: Todo;
+};
+
+export const ToDoItem: React.FC<Props> = ({ todo }) => {
+  return (
+    <div
+      data-cy="Todo"
+      className={classNames('todo', { completed: todo.completed })}
+    >
+      {/* eslint-disable jsx-a11y/label-has-associated-control */}
```
undeletable, line 15: [Not critical]
Disabling eslint rules is not a big deal for a homework. But for a real work, I would not recommend to disable eslint rules if errors are produced on your code. These rules were most likely added for some reason, so it's better to fix your code to align it with rules. If this is not possible or you feel that a rule doesn't give enough profit, you can discuss changing its configuration or even removal with the team.

src/components/ToDoItem/ToDoItem.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
```
undeletable, line 5: [Not critical]
It's better to name this type something like `ToDoItemProps`, so that the name describes this type the most precisely.

src/components/ToDoList/ToDoList.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import { ToDoItem } from '../ToDoItem/ToDoItem';
+
+type Props = {
```
undeletable, line 5: [Not critical]
It's better to name this type something like `ToDoListProps`, so that the name describes this type the most precisely.

undeletable (approved):
The solution looks fine, so I'm approving. I just have several advices for improvements - please check that and consider during the future work.

PR #1332, opened 2024-05-21
add solution

D-Kondratiev:
[DEMO LINK](https://D-Kondratiev.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,63 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './TodoList';
+import { Footer } from './Footer';
+import { Errors } from './types/Errors';
+import { FilterBy } from './types/FiilterBy';
+import { ErrorNotification } from './ErrorNotification';
+
+type FilterTheTodos = (todos: Todo[], filterBy: FilterBy) => Todo[];
+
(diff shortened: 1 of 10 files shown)


src/TodoList.tsx
```diff
@@ -0,0 +1,20 @@
...
+    {todos.map(todo => (
+      <TodoComp
+        key={todo.id}
+        id={todo.id}
+        completed={todo.completed}
+        title={todo.title}
+      />
+    ))}
+  </section>
+);
```
kinqbert, line 20: I believe it would be better if you'd pass whole todo instead of separate properties.

`<TodoComp key={todo.id} todo={todo} />`

kinqbert (approved):
**Great job!** 👍 Code looks clean and nice to me, approved. Check for small improvements.

_P.S._ Had to add some todos with my code to check if everything works fine.

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+        case FilterBy.Completed:
+          return todo.completed;
+        default:
+          throw new Error('Unknown filter type');
+      }
+    });
+  }
+
+  return filteredTodos;
+};
```
DanilWeda, line 31: Move to a separate file. In this case, you can use const without using let.

src/App.tsx
```diff
@@ -1,10 +1,43 @@
...
+import { FilterBy } from './types/FiilterBy';
+import { ErrorNotification } from './ErrorNotification';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.All);
+
+  useEffect(() => {
```
mrkovaliv, line 18: move on this logic to custom hook

mrkovaliv (approved):
👍 👍 👍

PR #1338, opened 2024-05-22
partial task solution(somewhere went a bit further)

mykhailonl:
[DEMO LINK](https://mykhailonl.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/context/AppContext.tsx
+++ b/src/context/AppContext.tsx
@@ -1,0 +1,128 @@
+import {
+  ReactNode,
+  createContext,
+  useEffect,
+  useMemo,
+  useReducer,
+} from 'react';
+import { State } from '../types/State';
+import { Action } from '../types/Action';
+import { AppContextType } from '../types/AppContextType';
+import { getTodos } from '../api/todos';
+import { Error } from '../types/Error';
+import { updateErrorState } from '../utils/errorHandler';
+
+const initialErrorState: Error[] = [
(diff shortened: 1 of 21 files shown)


src/context/AppContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+    case 'SET_TARGET_TODO':
+      return {
+        ...state,
+        targetTodo: action.payload,
+      };
+    case 'SET_FILTER':
+      const filteredTodos = state.todos.filter(todo => {
+        switch (action.payload) {
+          case 'All':
+            return true;
```
volodymyr-soltys97, line 80: This case is redundant, because you return it in default case
```suggestion
```

src/context/AppContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+          case 'Active':
+            return !todo.completed;
+          case 'Completed':
+            return todo.completed;
```
volodymyr-soltys97, line 84: Create a enum for `'Active', 'Completed', 'All'` and use it everywhere
```suggestion
          case 'Active':
            return !todo.completed;
            
          case 'Completed':
            return todo.completed;
            
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

volodymyr-soltys97 (approved):
Great work 🔥

PR #1352, opened 2024-06-02
react-todoApp-1

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,50 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [edit, setEdit] = useState(false);
```
DarkMistyRoom, line 11:
```suggestion
  const [isEditing, setIsEditing] = useState(false);
```

DarkMistyRoom: Read this article about bool variables naming https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+  const [status, setStaus] = useState('all');
```
DarkMistyRoom, line 12: would be good to create an Enum for statuses

src/App.tsx
DarkMistyRoom: But the main issue here is that when we double-click on one todo, all of them became editable

DarkMistyRoom (changes requested):
Good job! But let's fix the editing

PR #1357, opened 2024-06-06
Part 1 of 3

dmytro-popovych:
[DEMO LINK](https://dmytro-popovych.github.io/react_todo-app-loading-todos/)


src/components/TodoError.tsx
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useContext } from 'react';
+import classNames from 'classnames';
+
+import { TodoContext } from './TodoContext';
+
+export const TodoError = () => {
+  const { state, dispatch } = useContext(TodoContext);
+
+  const hiddenError = useCallback(() => {
```
TarasHoliuk, line 9:
```suggestion
  const hideError = useCallback(() => {
```
function name should be with verb
hiddenError sounds like some error (string/DOM element, etc., which is not visible)

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {state.todos.filter(todo => !todo.completed).length} items left
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {(Object.values(TodoStatus) as TodoStatus[]).map(status => (
```
TarasHoliuk, line 34: Do you really need type casting here?

src/components/ClearError.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useEffect, useCallback } from 'react';
+
+import { TodoErrors } from '../types/TodoErrors';
+import { Action } from './TodoContext';
+
+export const useAutoClearError = (
```
TarasHoliuk, line 6: 👍 
But it's a hook, not component, so the filename shouldn't be in PascalCase

I'd recommend name with `use` at the beginning to highlight there is hook

TarasHoliuk (approved):
Nicely done!

Just a few small comments. Also, remove redundant "background paper" in the footer:
image

PR #1393, opened 2024-07-05
Develop

MariCodec:
[DEMO LINK](https://MariCodec.github.io/react_todo-app-loading-todos/)


HumenVAsya (approved):
The code looks working, but it is better to separate everything into components. Also, the demo doesn't work. Try to fix it )))

MariCodec:
is divided into components in the following part

PR #1415, opened 2024-07-09
[Feature] added load todos

Arthur-Tolkachov:
[DEMO LINK](https://arthur-tolkachov.github.io/react_todo-app-loading-todos/)


src/components/todoApp/useTodoApp.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
+        setTodos(response);
+      })
+      .catch(() => {
+        onError('Unable to load todos');
+      });
+  }, []);
```
volodymyr-soltys97, line 49: Fix it 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix this failed test 
image
It is a bad practice to write so much code in one component, you need to split the code into different components, e.g TodoItem, TodoList, Footer

etojeDenys (approved):
good job

PR #1416, opened 2024-07-09
solution

ikocherovets:
[DEMO LINK](https://ikocherovets.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/List.tsx
+++ b/src/components/List.tsx
@@ -1,0 +1,51 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { FC } from 'react';
+import { Todo } from '../types/Todo';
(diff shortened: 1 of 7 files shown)


VenomMB (changes requested):
image
Your example doesn't work correctly, I can't test your solution

src/components/List.tsx
```diff
@@ -0,0 +1,51 @@
...
+import cn from 'classnames';
+
+interface Props {
+  todos: Todo[];
+}
+
+export const List: FC<Props> = ({ todos }) => {
+  if (todos.length === 0) {
+    return null;
+  }
```
VenomMB, line 14:
```suggestion
  if (!todos.length) {
    return null;
  }
```

SerhiiKirik (approved):
Well done 🔥

PR #1417, opened 2024-07-09
Solution part 1

DimaK88:
[DEMO LINK](https://DimaK88.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,203 +1,213 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
-
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { getTodos, loadTodos } from './api/todos';
+import classNames from 'classnames';
+export enum Status {
+  all = 'all',
+  active = 'active',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,203 +1,213 @@
...
-
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import { getTodos, loadTodos } from './api/todos';
+import classNames from 'classnames';
+export enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+}
```
volodymyr-soltys97, line 9: Move enum to separate file

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It's bad practice to write so much code in one place, you need to separate it into different components. For example TodoItem, TodoList, Footer

src/App.tsx
DimaK88: can i do all three parts at once?

src/App.tsx
```diff
@@ -1,203 +1,132 @@
...
+  useEffect(() => {
+    if (error) {
+      const timer = setTimeout(() => {
+        setError(null);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return () => {};
```
etojeDenys, line 34: looks redundant

etojeDenys (approved):
lgtm

PR #1431, opened 2024-07-17
loading

shelyaa:
[DEMO LINK](https://shelyaa.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,14 +1,39 @@
...
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const fileredTodos = todos.filter(todo => {
+    if (filter === 'active') {
+      return !todo.completed;
+    }
+
+    if (filter === 'completed') {
```
volodymyr-soltys97, line 30: Create a enum for 'active', 'completed' and use it here

src/App.tsx
```diff
@@ -35,157 +60,102 @@ export const App: React.FC = () => {
...
-            </div>
-          </div>
+          {loading ? (
+            <div>Loading...</div>
+          ) : (
+            fileredTodos.map(todo => (
+              <div
+                key={todo.id}
+                data-cy="Todo"
+                className={`todo ${todo.completed ? 'completed' : ''}`}
```
volodymyr-soltys97, line 70: Use the classnames library rules for add classes with condition

src/App.tsx
```diff
@@ -35,157 +60,102 @@ export const App: React.FC = () => {
...
+                  <input
+                    data-cy="TodoStatus"
+                    type="checkbox"
+                    className="todo__status"
+                    checked={todo.completed}
+                  />
+                </label>
+
+                <span data-cy="TodoTitle" className="todo__title">
+                  {todo.title}
```
volodymyr-soltys97, line 82: Use destructuring for `todo`

src/App.tsx
```diff
@@ -35,157 +60,102 @@ export const App: React.FC = () => {
...
+              </a>
+
+              <a
+                href="#/completed"
+                className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
```
volodymyr-soltys97, line 138: Use `Object.values()` and `.map()` method for render options 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix error message for pass tests 
image

It is a bad practice to write so much code in one component, you need to separate the logic into different components, for example `TodoItem, TodoList, Header, Footer`

src/components/TodoList.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (filter === Filter.active) {
+      return !todo.completed;
+    }
+
+    if (filter === Filter.completed) {
+      return todo.completed;
+    }
+
+    return true;
+  });
```
SanyaBratashchuk, line 26: move filtering to higher level. TodoList should receive list of todos and render it shouldn't filter them

src/components/TodoList.tsx
```diff
@@ -0,0 +1,38 @@
...
+  loading,
+}) => {
+  const fileredTodos = todos.filter(todo => {
+    if (filter === Filter.active) {
+      return !todo.completed;
+    }
+
+    if (filter === Filter.completed) {
+      return todo.completed;
+    }
```
SanyaBratashchuk, line 23: use switch case

src/components/TodoList.tsx
SanyaBratashchuk, line 26: move filter insdide conditions to avoid extra iteration when filter value is `All `

SanyaBratashchuk (changes requested):

VitaliyBondarenko1982 (approved):
GJ

PR #1439, opened 2024-07-22
feat: add solution

mfcastilho:
[DEMO LINK](https://mfcastilho.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,44 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList';
+import { Error } from './components/Error/Error';
+import { Footer } from './components/Footer/Footer';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,10 +1,44 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [status, setStatus] = useState(Status.All);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
+        setTimeout(() => setError(''), 3000);
```
pedro-ruas, line 23: It will be much better to move this `setTimeout` to another `useEffect` targeting "error".
Then it will track all errors and not only this one

src/App.tsx
```diff
@@ -1,10 +1,44 @@
...
+      });
+  }, []);
+
+  function getVisibleToDos(newTodos: Todo[], newStatus: Status) {
```
pedro-ruas, line 27: This is a pure function, and it does not check on the state by itself, it needs to receive the state.
If that is the case, you should probably declare this function outside the component so it doesn't get re-declared on every re-render

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { useEffect, useRef } from 'react';
+
+export const Header: React.FC = () => {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(() => {
+    if (inputRef.current) {
```
pedro-ruas, line 8: `inputRef.current?.focus()` gets the same effect

pedro-ruas (changes requested):
Minimal improvements. Good job!

PR #1442, opened 2024-07-23
Solution to task

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_todo-app-loading-todos/)


src/context/AppContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type Props = {
+  children: ReactNode;
+};
+
+const AppContextContainer = createContext({} as AppContextContainerProps);
+
+export const useAppContextContainer = () => {
+  return useContext(AppContextContainer);
+};
```
Zibi95, line 30: It would be also good to throw error in case when useContext returns undefined which means that the component that tries to use this hook is not wrapped inside context. Its then easier to debug errors.

Zibi95 (approved):
Really well done. Code looks solid keep it up 💪

PR #1449, opened 2024-07-27
loading todo

IrynaZahorodnia:
[DEMO LINK](https://IrynaZahorodnia.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -29,175 +62,117 @@ export const App: React.FC = () => {
...
-            </a>
-
-            <a
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todos.length > 0 && (
```
volodymyr-soltys97, line 108:
```suggestion
        {!!todos.length && (
```

src/App.tsx
```diff
@@ -29,175 +62,117 @@ export const App: React.FC = () => {
...
+              <a
+                href="#/completed"
+                className={classNames('filter__link', {
+                  selected: filter === Filter.completed,
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter(Filter.completed)}
+              >
+                Completed
+              </a>
```
volodymyr-soltys97, line 147: Use `Object.values()` and render these options with `map` method

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It's bad practice to write so much code in one component, you need to split the logic into different components. For example TodoItem, TodoList, Header, Footer

GUSILLUS (approved):
GJ!

PR #1450, opened 2024-07-27
Develop

Hy-tapa-kot:
[DEMO LINK](https://Hy-tapa-kot.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. You need to fix all linter errors

2. Add data attribute for tests 
image

3. It is bad practice to write so much code in one component, you need to split it into different components. For example TodoItem, TodoList, Footer, Header

anastasiiavorobiova (approved):
Great work!

PR #1457, opened 2024-08-01
The task is not fully completed

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_todo-app-loading-todos/)

The task is not fully completed, there was not enough time to complete it, I will finish either later or tomorrow


BoGoBooM:
I just don't quite understand what the problem is right now, I want to solve it myself, but if I have problems with it, I'll ask a question in the Mates chat

etojeDenys (approved):

PR #1468, opened 2024-08-19
success

vladisssi:
[DEMO LINK](https://vladisssi.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodos, setSelectedTodos] = useState<FilterTypes>(
+    FilterTypes.All,
+  );
+  const [isErrorHidden, setIsErrorHidden] = useState(true);
+
+  const areTodosExist = todos.length !== 0;
+  const notCompletedTodos = todos.filter(todo => !todo.completed).length;
+
+  let filteredTodos = [...todos];
```
TarasHoliuk, line 21: instead create `const` variable, use `useMemo` hook and write filtering logic inside the callback

src/components/TodoInput/TodoInput.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
+
+type Props = {
+  todo: Todo;
+};
+
+export const TodoInput: React.FC<Props> = ({ todo }) => (
```
TarasHoliuk, line 9: it's not input (there could be input during editing todo, but the entire component is not just input). It's one item of the TodoList, so let's reneme it to TodoItem

```suggestion
export const TodoItem: React.FC<Props> = ({ todo }) => (
```

src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { FilterTypes } from './types/FilterTypes';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodos, setSelectedTodos] = useState<FilterTypes>(
```
TarasHoliuk, line 13: No need to add `Types` word in enum, so just `Filter`

Also, let's state name match it - filter, setFilter

src/App.tsx
```diff
@@ -33,171 +70,95 @@ export const App: React.FC = () => {
...
-
-            <a
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {areTodosExist && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
TarasHoliuk, line 76: It's better to move the footer to a separate component
Same for the header

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #1478, opened 2024-08-22
Added solution

k-marchuk:
[DEMO LINK](https://k-marchuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,41 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useMemo, useState } from 'react';
+import classNames from 'classnames';
+import { getTodos } from './api/todos';
+import { TodoList } from './components/Todo';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { getFilteredTodos } from './utils/TodosFilter';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { Error } from './types/Error';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortedStatus, setSortedStatus] = useState<Filter>(Filter.ALL);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(error => {
+        setErrorMessage(Error.GET);
+        throw error;
+      });
+  }, []);
+
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
+
+  const preparedTodos = getFilteredTodos(todos, sortedStatus);
+
+  const activeTodosCount = useMemo(() => {
+    return todos?.filter(todo => !todo.completed).length;
+  }, [todos]);
 
   return (
     <div className="todoapp">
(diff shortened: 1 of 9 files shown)


Kateryna99 (approved):

src/App.tsx
```diff
@@ -1,28 +1,56 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [sortedStatus, setSortedStatus] = useState<Filter>(Filter.ALL);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(error => {
+        setErrorMessage(Error.GET);
+        throw error;
```
BudnikOleksii, line 22: You don't to throw error, you set it to the state, show to user and that's enough

src/App.tsx
```diff
@@ -1,28 +1,56 @@
...
+      });
+  }, []);
+
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+  }, [errorMessage]);
```
BudnikOleksii, line 32: - Move this logic to the error notification component
- And it's a good practice to remove timers/timeouts in case a component unmount

src/App.tsx
```diff
@@ -34,170 +62,34 @@ export const App: React.FC = () => {
...
-            <a
-              href="#/completed"
-              className="filter__link"
-              data-cy="FilterLinkCompleted"
-            >
-              Completed
-            </a>
-          </nav>
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
BudnikOleksii, line 69: Create a footer component

src/App.tsx
```diff
@@ -34,170 +62,34 @@ export const App: React.FC = () => {
...
-        </footer>
+            <TodoFilter status={sortedStatus} onSorted={setSortedStatus} />
+          </footer>
+        )}
       </div>
 
-      {/* DON'T use conditional rendering to hide the notification */}
-      {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
```
BudnikOleksii, line 80: Create an `ErrorNotification` component

src/components/Todo/TodoList.tsx
```diff
@@ -0,0 +1,41 @@
...
+import classNames from 'classnames';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return todos.map(todo => (
+    <div
+      data-cy="Todo"
```
BudnikOleksii, line 14: Create a todo component

src/components/Todo/TodoList.tsx
```diff
@@ -0,0 +1,41 @@
...
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          defaultChecked={todo.completed}
+        />
+      </label>
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
+      </span>
+      {/* Remove button appears only on hover */}
```
BudnikOleksii, line 30: Remove all comments

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { Filter } from '../../types/Filter';
+import classNames from 'classnames';
+
+type Props = {
+  status: Filter;
+  onSorted: (sortedField: Filter) => void;
```
BudnikOleksii, line 7:
```suggestion
  onFilterChange: (sortedField: Filter) => void;
```
It's not about sorting 😄

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <a
+        href="#/completed"
+        className={classNames('filter__link', {
+          selected: status === Filter.COMPLETED,
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={() => onSorted(Filter.COMPLETED)}
+      >
+        Completed
+      </a>
```
BudnikOleksii, line 44: You can generate a list of these links from your enum. `Object.keys/values/entries` - creates an array from enum

BudnikOleksii (changes requested):
Looks good 👍 
Let's improve your solution slightly

src/App.tsx
k-marchuk: Ok, but this component will never unmount because it changes its visibility by className, not by conditional rendering.

BudnikOleksii: > Ok, but this component will never unmount because it changes its visibility by className, not by conditional rendering.

In this case, it's correct, but this app could be a part of another app that renders it on some specific page.
Update and move this logic to the corresponding component in the next task

BudnikOleksii (approved):
👍

PR #1482, opened 2024-08-22
Solution

Kateryna99:
[DEMO LINK](https://Kateryna99.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,88 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useRef } from 'react';
+import { getTodos, updateTodos } from './api/todos';
+import { Todo } from './types/Todo';
+
+import classNames from 'classnames';
+import { TodoList } from './TodoList';
+import { getFilteredData } from './helpres/helpers';
+import { FilterTypes } from './enum/FilterTypes';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [todos, setTodos] = React.useState<Todo[]>([]);
+
+  const [todoLoadingStates, setTodoLoadingStates] = React.useState<{
+    [key: number]: boolean;
+  }>({});
+  const [isListLoading, setIsListLoading] = React.useState(true);
+  const [errorMessage, setErrorMessage] = React.useState('');
+
+  const [filter, setFilter] = React.useState<FilterTypes>(FilterTypes.All);
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const displayedTodos = getFilteredData(todos, filter);
+  const getNotCompletedTodos = todos.filter(todo => !todo.completed).length;
+
+  const setTodoLoading = (id: number, loading: boolean) => {
+    setTodoLoadingStates(prevState => ({ ...prevState, [id]: loading }));
+  };
+
+  const onSelectInputChange = async (id: number, completed: boolean) => {
+    try {
+      setTodoLoading(id, completed);
+      setErrorMessage('');
+
+      await updateTodos(id, true);
+
+      setTodos(todosList =>
+        todosList.map(todo => (todo.id === id ? { ...todo, completed } : todo)),
+      );
+    } catch (error) {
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,204 +1,94 @@
...
+    if (errorMessage) {
+      mistakeTimer = window.setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+    }
+
+    return () => {
+      window.clearTimeout(mistakeTimer);
+    };
+  }, [errorMessage]);
```
BudnikOleksii, line 60: Move this logic to error component

src/App.tsx
```diff
@@ -1,204 +1,94 @@
...
-          />
-
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
+    <Provider value={contextValue}>
```
BudnikOleksii, line 70: Create a separate component for your context and move it to context file

src/components/filterFooter/FilterFooter.tsx
```diff
@@ -0,0 +1,75 @@
...
+          {FilterTypes.Active}
+        </a>
+
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: isSelectedFilter(FilterTypes.Completed),
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilterType(FilterTypes.Completed)}
```
BudnikOleksii, line 59: You can render a list of links by converting your enum into an array with `Object.keys/values/entries` and then map through it

src/components/todoForm/TodoForm.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={classNames('todoapp__toggle-all', {
+          active: completedTodos === todos.length,
+        })}
+        data-cy="ToggleAllButton"
+      />
+
+      {/* Add a todo on form submit */}
```
BudnikOleksii, line 29: Remove all comments

src/components/todoItem/TodoItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+        } catch (error) {
+          context?.setErrorMessage('Something went wrong');
+        } finally {
+          context?.setTodoLoading(selectedTodoId, false);
+        }
+      };
+
+      updateTodo();
+    }
+  }, [selectedTodoId, todoStatus]);
```
BudnikOleksii, line 45: It shouldn't be in `useEffect`, just call the `updateTodo` function when you need it

src/context/context.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Todo } from '../types/Todo';
+
+interface ContextProps {
+  todoLoadingStates: { [key: number]: boolean };
+  setTodoLoading: (id: number, status: boolean) => void;
+  setErrorMessage: (message: string) => void;
+  setTodos: Dispatch<SetStateAction<Todo[]>>;
+}
+
+export const TodoContext = createContext<ContextProps | undefined>(undefined);
```
BudnikOleksii, line 11: You can create a custom hook for your context so you don't need to pass it in `useContext` and it always will be defined, something like this:
```typescript
const useTodosContext = () => {
  const context = useContext(TodoContext);

  if (!context) {
    throw new Error('You can use context only inside the provider');
  }

  return context;
};
```

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

src/components/todoForm/TodoForm.tsx
Kateryna99: I will not remove now this comment because I have not done this part of task.
I will remove it on the second part.

src/components/todoItem/TodoItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+      setTodoLoading(id, true);
+      setErrorMessage('');
+      await updateTodos(id, newStatus);
+      setTodos(prevTodos => {
+        return prevTodos.map(item =>
+          item.id === id ? { ...item, completed: newStatus } : item,
+        );
+      });
+    } catch (error) {
+      setErrorMessage('Something went wrong');
```
BudnikOleksii, line 30: The error message should describe what actually gone wrong

BudnikOleksii (approved):
Looks good👍

PR #1491, opened 2024-08-26
react_todo-app-loading-todos

Abdrakhmanov-Renat:
I completed the tasks from the README file, but not all the tests passed.
https://Abdrakhmanov-Renat.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,69 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
+
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoCard } from './components/TodoCard';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState<string>('');
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  const [allTodos, setAllTodos] = useState<boolean>(true);
+  const [activeTodos, setActiveTodos] = useState<boolean>(false);
+  const [completedeTodos, setCompletedTodos] = useState<boolean>(false);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const handleTitleSubmit = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setErrorMessage('');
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,204 +1,187 @@
...
+    if (!title) {
+      setErrorMessage('Title should not be empty');
+    }
+  };
+
+  const handleButtonClose = () => {
+    setErrorMessage('');
+  };
+
+  const filterTodoAll = () => {
```
TarasHoliuk, line 43: You shouldn't fetch todos from the server on every filter change. Instead fetch it once (as you did in useEffect). on filter click save the filter value to a separate state.

Then having all `todos` and `filter` status value, you can filter todos depending on status and save filtered todos to a common variable. On filter or todos state change rerender will happen and this common variable will be recomputed.

src/App.tsx
```diff
@@ -1,204 +1,187 @@
...
+import { TodoCard } from './components/TodoCard';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState<string>('');
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  const [allTodos, setAllTodos] = useState<boolean>(true);
+  const [activeTodos, setActiveTodos] = useState<boolean>(false);
+  const [completedeTodos, setCompletedTodos] = useState<boolean>(false);
```
TarasHoliuk, line 18: redundant states

src/App.tsx
TarasHoliuk: P.S. use enum for Filter values

src/App.tsx
```diff
@@ -1,204 +1,187 @@
...
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+          {todos.map(todo => (
+            <TodoCard key={todo.id} todo={todo} />
```
TarasHoliuk, line 99: it's not a card, but a list item, so TodoItem name would look better here

src/App.tsx
```diff
@@ -1,204 +1,187 @@
...
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todos.length !== 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span className="todo-count" data-cy="TodosCounter">
+              0 items left
```
TarasHoliuk, line 107: you should display the real amount of active (not completed) todos

TarasHoliuk (changes requested):

volodymyr-soltys97 (changes requested):
Hi, you need to fix the failed test, need to add the data-cy attribute, if you need help feel free in the fe_chat 
image

etojeDenys (approved):
lgtm

PR #1503, opened 2024-09-05
Solution

yuliia-nudyk:
DEMO-LINK - https://yuliia-nudyk.github.io/react_todo-app-loading-todos/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code It is a bad practice to write so much code in one component, you need to split the code into different components, for example `Header, TodoItem, TodoList, Footer`

anastasiiavorobiova (approved):
Great work!

PR #1513, opened 2024-09-16
Solution

meljaszuk:
[DEMO](https://meljaszuk.github.io/react_todo-app-loading-todos/)


natalia-klonowska (changes requested):
tests are not working :( I see that you forked this repo a while ago so maybe you need to update it

PR #1516, opened 2024-09-18
solution part 1

syavaYki:
[DEMO LINK](https://syavaYki.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,61 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+
 import { UserWarning } from './UserWarning';
+import { Todo } from './types/Todo';
+
 import { USER_ID } from './api/todos';
+import { todos as todosClient } from './api/todos';
+import { ErrorNotification } from './Components/ErrorNotification';
+import { TodoHead } from './Components/TodoHead';
+import { TodoList } from './Components/TodoList';
+import { TodoFilter } from './Components/TodoFilter';
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [filter, setFilter] = useState<Filters>(Filters.all);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    todosClient
+      .get()
+      .then(res => {
+        setTodos(res);
+        setFilteredTodos(res);
+      })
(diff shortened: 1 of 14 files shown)


src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onClick: (filter: Filters) => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({ todos, filter, onClick }) => {
+  return (
+    <>
+      {todos.length > 0 && (
+        <footer className="todoapp__footer" data-cy="Footer">
+          <span className="todo-count" data-cy="TodosCounter">
+            {`${todos.filter(todo => !todo.completed).length} items left`}
```
volodymyr-soltys97: Move this logic to the helper variable and use it here

src/Components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+            <a
+              href="#/completed"
+              className={classNames('filter__link', {
+                selected: filter === Filters.completed,
+              })}
+              data-cy="FilterLinkCompleted"
+              onClick={() => onClick(Filters.completed)}
+            >
+              Completed
```
volodymyr-soltys97: Use `Object.values(your enum)` and render these options with `map()` method

src/Components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+import { TodoList } from './Components/TodoList';
+import { TodoFilter } from './Components/TodoFilter';
+import { Errors } from './types/Errors';
+import { Filters } from './types/Filters';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<Errors | null>(null);
+  const [filter, setFilter] = useState<Filters>(Filters.all);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
volodymyr-soltys97, line 19: Do you need separate state for filtered todos? If you can use only one state `todos`
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/App.tsx
syavaYki: Used to store todos from server to avoid api call when filtering. Same as Example. Is there better way to do it?

volodymyr-soltys97 (approved):
GJ!

PR #1546, opened 2024-10-14
solved 1st part of Todo task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useState } from 'react';
+import { deleteTodos, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+import { FilterOptions } from './FilterOptions';
 
 export const App: React.FC = () => {
```
KrivorotkoYurii, line 13: It will be better if you create more components so that your code is more readable

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+        setErrorMessage('');
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [errorMessage]);
+
+  const handleTodoDelete = (todoId: number) => {
+    deleteTodos(todoId)
+      .then(() => {
```
KrivorotkoYurii, line 45: Since you`ve been already using async await syntax above, it might be better to use it here too.

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      try {
+        const loadedTodos = await getTodos();
+
+        setTodos(loadedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to load todos');
```
KrivorotkoYurii: It will be better to create an enum in different file . You could add the error`s string values to your future enum. And use the enum in the places these strings occur.

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+  };
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case FilterOptions.ACTIVE:
+        return !todo.completed;
+      case FilterOptions.COMPLETED:
+        return todo.completed;
+      default:
+        return true;
```
KrivorotkoYurii: Your solution means that you are going to use method 'filter' also in case of state filter === all.  It leads you to overiterating and  missing the optimisation. Imagine that your array will have too many items inside and it will take more time to render the page. So you can create a function getFilteredTodos and use your switch case and filter method in it with a default value of 'return todos'   . After that const filteredTodos = getFilteredTodos()

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
+                    type="checkbox"
+                    className="todo__status"
+                    checked={todo.completed}
+                  />
+                </label>
+
+                <span data-cy="TodoTitle" className="todo__title">
+                  {todo.title}
+                </span>
+
```
KrivorotkoYurii: You missed a form for todo which is being edited
```suggestion
                            {/* This form is shown instead of the title and remove button */}
            <form>
              <input
                data-cy="TodoTitleField"
                type="text"
                className="todo__title-field"
                placeholder="Empty todo will be deleted"
                value="Todo is being edited now"
              />
            </form>
```

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
+            <nav className="filter" data-cy="Filter">
+              <a
+                href="#/"
+                className={classNames('filter__link', {
+                  selected: filter === FilterOptions.ALL,
+                })}
+                data-cy="FilterLinkAll"
+                onClick={() => handleFilterChange(FilterOptions.ALL)}
+              >
+                All
```
KrivorotkoYurii: If you have already created an enum for these values ,  you schould use it here instead of hardcoded values

KrivorotkoYurii (approved):
Good job) Just check some commetns below

src/App.tsx
BudnikOleksii: > It will be better to create an enum in different file . You could add the error`s string values to your future enum. And use the enum in the places these strings occur.

Good point

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+
+  useEffect(() => {
+    if (errorMessage) {
+      const timer = setTimeout(() => {
+        setErrorMessage('');
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [errorMessage]);
```
BudnikOleksii: Move this logic to the error notification component

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+
+    setTodos(currentTodos =>
+      currentTodos.map(todo => ({
+        ...todo,
+        completed: hasIncompleteTodos,
+      })),
+    );
+  };
+
+  const filteredTodos = todos.filter(todo => {
```
BudnikOleksii: Create a function to filter todos and move it to utils/helpers folder/file

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
+      default:
+        return true;
+    }
+  });
+
+  const completedTodos = todos.filter(todo => todo.completed);
+
+  const handleFilterChange = (newFilter: FilterOptions) => {
+    setFilter(newFilter);
+  };
```
BudnikOleksii: You don't need this function as it just path an argument to the `setFilter`

src/App.tsx
```diff
@@ -1,26 +1,99 @@
...
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
```
BudnikOleksii: Create a header component

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
               type="text"
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              autoFocus
             />
           </form>
         </header>
 
         <section className="todoapp__main" data-cy="TodoList">
```
BudnikOleksii: Create a todo list component

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+          {filteredTodos.map(todo => {
+            return (
+              <div
+                data-cy="Todo"
```
BudnikOleksii: Create a todo component

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
+                <button
+                  type="button"
+                  className="todo__remove"
+                  data-cy="TodoDelete"
+                  onClick={() => handleTodoDelete(todo.id)}
+                >
+                  ×
+                </button>
+
+                {/* overlay will cover the todo while it is being deleted or updated */}
```
BudnikOleksii: Remove all comments

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
-            type="button"
-            className="todoapp__clear-completed"
-            data-cy="ClearCompletedButton"
-          >
-            Clear completed
-          </button>
-        </footer>
-      </div>
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
BudnikOleksii: Create a footer component

src/App.tsx
BudnikOleksii: > If you have already created an enum for these values , you schould use it here instead of hardcoded values

Good point

src/App.tsx
```diff
@@ -29,175 +102,121 @@ export const App: React.FC = () => {
...
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => handleFilterChange(FilterOptions.COMPLETED)}
+              >
+                Completed
+              </a>
+            </nav>
+
+            {/* this button should be disabled if there are no completed todos */}
+            {todos.some(todo => todo.completed) && (
```
BudnikOleksii: Create a variable for `todos.some(todo => todo.completed)`

BudnikOleksii (changes requested):
Good work 👍 
Left some comments with suggestion, check them

src/App.tsx
VolodymyrKirichenko: Good point 🔥 
Don't mix ur code! Use async/await in both cases

src/App.tsx
```diff
@@ -1,203 +1,113 @@
...
+
+    setTodos(currentTodos =>
+      currentTodos.map(todo => ({
+        ...todo,
+        completed: hasIncompleteTodos,
+      })),
+    );
+  };
+
+  const todosAfterFiltering = filteredTodos(todos, filter);
```
VolodymyrKirichenko: useMemo?

src/App.tsx
```diff
@@ -1,203 +1,113 @@
...
+          completedTodos={completedTodos}
+          onTodosToggle={handleTodosToggle}
+        />
+
+        <TodoList
+          todosAfterFiltering={todosAfterFiltering}
+          onTodoDelete={handleTodoDelete}
+        />
+
+        {todos.length > 0 && (
```
VolodymyrKirichenko: u can use the `!!` operator to convert it to a boolean value
 change for all such cases and use it in the future
```suggestion
        {!!todos.length && (
```

src/api/todos.ts
```diff
@@ -1,10 +1,12 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const USER_ID = 0;
+export const USER_ID = 1592;
```
VolodymyrKirichenko: CONSTANTS must be in a separate `utils/constants` file

src/App.tsx
```diff
@@ -1,203 +1,113 @@
...
+        setTodos(currentTodos =>
+          currentTodos.filter(todo => todo.id !== todoId),
+        );
+      })
+      .catch(() => {
+        setErrorMessage(ErrorMessages.DELETING_ERROR);
+      });
+  };
+
+  const handleCompletedTodosDeleted = () => {
```
VolodymyrKirichenko: 1) `useCallback`? use it everywhere u need it
2) sounds like all the COMPLETED todos will be removed. Maybe - `handleDeleteSelectedTodos`?
```suggestion
  const handleDeleteSelectedTodos = () => {
```

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's do some fixes

src/App.tsx
```diff
@@ -1,203 +1,120 @@
...
+        ...todo,
+        completed: hasIncompleteTodos,
+      })),
+    );
+  }, [todos]);
+
+  const todosAfterFiltering = useMemo(
+    () => filteredTodos(todos, filter),
+    [todos, filter],
+  );
```
VolodymyrKirichenko, line 73: bad syntax
```suggestion
  const todosAfterFiltering = useMemo(() => (
    filteredTodos(todos, filter)
  ), [todos, filter]);
```

src/App.tsx
VolodymyrKirichenko: I saw ur option with `async/await.` Why did u remove it?
I recommend u to forget about `then`

src/App.tsx
```diff
@@ -1,203 +1,120 @@
...
+
+  useEffect(() => {
+    if (errorMessage) {
+      const timer = setTimeout(() => {
+        setErrorMessage(ErrorMessages.NO_ERRORS);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [errorMessage]);
```
VolodymyrKirichenko, line 41: all `useEffects` must be at the bottom level before rendering

src/components/Header.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const Header: React.FC<HeaderProps> = ({
+  todos,
+  completedTodos,
+  onTodosToggle,
+}) => {
+  return (
+    <header className="todoapp__header">
+      {todos.length > 0 && (
```
VolodymyrKirichenko, line 17: change it in all the cases
```suggestion
      {...... && (
```

VolodymyrKirichenko (approved):
Good job 👍 
But review my comments, again)

PR #1549, opened 2024-10-14
add solution

YuriiMykhailenko:
[DEMO LINK](https://YuriiMykhailenko.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,13 +1,49 @@
...
-import { USER_ID } from './api/todos';
+import React, { useEffect, useState } from 'react';
+// import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { ToodoList } from './components/TodoList';
+import { FilterOptions } from './utils/FilterOptions';
+import cn from 'classnames';
+
+function getFilteredTodos(todos: Todo[], filterOption: FilterOptions): Todo[] {
```
Ann-Maksay: It is better to put the helper functions in a separate directory, all the more it already exists in our project

src/App.tsx
```diff
@@ -1,13 +1,49 @@
...
+    }
+  });
+}
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todoos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
```
Ann-Maksay: It would be nice to create a separate enum for the list of possible errors and use it. Here and below in useEffect

src/App.tsx
```diff
@@ -1,13 +1,49 @@
...
+      .then(setTodos)
+      .catch(() => setError('Unable to load todos'))
+      .finally(() =>
+        setTimeout(() => {
+          setError('');
+        }, 3000),
+      );
+  }, []);
+
+  const numberOfActiveTodos = todoos.filter(todo => !todo.completed).length;
```
Ann-Maksay: This is an example of an operation that could be wrapped in useMemo, but it's all up to you

src/App.tsx
```diff
@@ -34,170 +70,79 @@ export const App: React.FC = () => {
...
-            </a>
-
-            <a
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todoos.length && (
```
Ann-Maksay: Better make sure it's a boolean value. If you pay attention, your demo shows 0 before the data is loaded

src/App.tsx
```diff
@@ -34,170 +70,79 @@ export const App: React.FC = () => {
...
-              ×
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
-            </div>
-          </div>
+          <ToodoList todos={filteredTodos} />
```
Ann-Maksay: Perhaps we don't need to render the list if the todos list is empty?

src/App.tsx
```diff
@@ -34,170 +70,79 @@ export const App: React.FC = () => {
...
-        </footer>
+              Clear completed
+            </button>
+          </footer>
+        )}
       </div>
 
       {/* DON'T use conditional rendering to hide the notification */}
       {/* Add the 'hidden' class to hide the message smoothly */}
       <div
```
Ann-Maksay: I would recommend taking more components out separately, such as Header, Footer and ErrorNotification, as they are quite complex and in theory could be reused.

Ann-Maksay (approved):
Nice work 🌟
From myself added a few recommendations for improvement

src/App.tsx
```diff
@@ -1,13 +1,49 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useState } from 'react';
+// import { UserWarning } from './UserWarning';
```
VolodymyrKirichenko: remove all unused comments

src/App.tsx
VolodymyrKirichenko: Good point 🔥 
Move it to `utils` directory

src/App.tsx
VolodymyrKirichenko: Good idea 👍 
Use `enum` for ur error.
```suggestion
  const [error, setError] = useState<ErrorType>(ErrorType.Default);
```

src/App.tsx
VolodymyrKirichenko: yep, u need `useMemo` in this case.
With `useMemo`, u will avoid re-creating the variable at each rendering

src/App.tsx
VolodymyrKirichenko: 1) use operator `!!` to convert to boolean type
2) fix syntax error `todoos`
```suggestion
        {!!todos.length && (
```

src/App.tsx
VolodymyrKirichenko: yep, do we need it?what do u think?

src/App.tsx
```diff
@@ -34,170 +70,79 @@ export const App: React.FC = () => {
...
-
-            <a
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todoos.length && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
VolodymyrKirichenko: move ur `Footer` to the separate component

src/App.tsx
VolodymyrKirichenko: Create a new component - `ErrorNotification ` for it

VolodymyrKirichenko (changes requested):
Looks good, but let's improve ur code 🥹

BudnikOleksii (approved):

PR #1559, opened 2024-10-15
Solution

vikshiyan:
[DEMO LINK](https://vikshiyan.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
 
 import { getTodos } from './api/todos';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,6 +8,7 @@ import { Footer } from './components/Footer';
 import { Error } from './components/Error';
 import { TypeError } from './types/TypeError';
 import { SelectedType } from './types/SelectedType';
+import { Header } from './components/Header';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
@@ -18,35 +19,20 @@ export const App: React.FC = () => {
     SelectedType.ALL,
   );
 
+  const filteredTodos = getFilteredTodos(todos, selectedOption);
+
   useEffect(() => {
     getTodos()
       .then(setTodos)
       .catch(() => setErrorMessage(TypeError.LOADING));
   }, []);
 
-  const filteredTodos = getFilteredTodos(todos, selectedOption);
-
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
-
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
+        <Header />
 
         {todos.length > 0 && (
           <>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,32 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useState } from 'react';
+
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import { getFilteredTodos } from './utils/getFilterdTodos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Error } from './components/Error';
+import { TypeError } from './types/TypeError';
+import { SelectedType } from './types/SelectedType';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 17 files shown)


mbulchak (approved):
Nice job!

src/App.tsx
```diff
@@ -1,28 +1,45 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
BudnikOleksii: Avoid linter disabling

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,34 @@
...
+  errorMessage: TypeError;
+  setErrorMessage: React.Dispatch<React.SetStateAction<TypeError>>;
+};
+
+export const Error: React.FC<Props> = ({ errorMessage, setErrorMessage }) => {
+  useEffect(() => {
+    if (errorMessage) {
+      setTimeout(() => setErrorMessage(TypeError.DEFAULT), 3000);
+    }
+  }, [errorMessage]);
```
BudnikOleksii: It's a good practice to remove timers on component unmount

BudnikOleksii: Add setErrorMessage as a dependency

React Hook useEffect has a missing dependency: 'setErrorMessage'. Either include it or remove the dependency array. If 'setErrorMessage' changes too often, find the parent component that defines it and wrap that definition in useCallback

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => {
+        return (
+          <div
+            key={todo.id}
+            data-cy="Todo"
```
BudnikOleksii: Create a todo component

src/types/TypeError.ts
```diff
@@ -0,0 +1,6 @@
+export enum TypeError {
+  DEFAULT = '',
+  LOADING = 'Unable to load todos',
+}
+//
+// add
```
BudnikOleksii: Remove all comments

src/utils/getFilterdTodos.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Todo } from '../types/Todo';
+
+export const getFilteredTodos = (
+  todos: Todo[],
+  selectedOption: SelectedType,
+) => {
+  let filteredTodos = [...todos];
+
+  if (selectedOption !== SelectedType.ALL) {
+    filteredTodos = filteredTodos.filter(todo => {
```
BudnikOleksii:
```suggestion
  if (selectedOption !== SelectedType.ALL) {
    todos.filter(todo => {
```
- `filter` doesn't mutate an array, so you don't need a copy
- You don't need to assign to a variable, you can just return values from function

BudnikOleksii (changes requested):
Good work 👍
Left a few comments to improve your solution

src/App.tsx
vikshiyan: I check and use linter disable in two files TodoItem and in fetchClient, because there are mistake with label and TS mistake with any

src/App.tsx
```diff
@@ -1,28 +1,43 @@
...
+  );
+  const [selectedOption, setSelectedOption] = useState<SelectedType>(
+    SelectedType.ALL,
+  );
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage(TypeError.LOADING));
+  }, []);
```
igorkruz, line 28: It's better to move the logic inside the useEffect hook to a separate function, as it improves code readability, reusability, and makes testing easier. It also helps keep useEffect focused on side effects.

Move the `useEffect `to the end of code
Placing `useEffect `at the end improves readability by letting developers first grasp the main logic before seeing side effects. This promotes clearer structure and consistency in the codebase.

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  setSelectedOption: React.Dispatch<React.SetStateAction<SelectedType>>;
+};
+
+export const Footer: React.FC<Props> = ({
+  selectedOption,
+  setSelectedOption,
+  todos,
+}) => {
+  const filtersOptionName = Object.values(SelectedType);
+  const notCompleteedCount = todos.filter(todo => !todo.completed).length;
```
KosolapovAnatolii, line 19:
```suggestion
  const notCompletedCount = todos.filter(todo => !todo.completed).length;
```
:))

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import React from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+};
+
+export const Todoitem: React.FC<Props> = ({ todo }) => {
```
KosolapovAnatolii, line 12:
```suggestion
export const TodoItem: React.FC<Props> = ({ todo }) => {
```
Much better for readabillity call React components var with UpperCamelCase
:))

KosolapovAnatolii (approved):
LGTM 🔥 🔥 🚒

src/components/TodoItem/TodoItem.tsx
vikshiyan: Thank you, I didn't see these two mistakes. I'll fix it

PR #1561, opened 2024-10-16
1 task solution

Valeraom:
[DEMO LINK](https://Valeraom.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/mocks.ts
+++ b/src/mocks.ts
@@ -1,0 +1,3 @@
+import { FilterType } from './types/FilterType';
+
+export const FILTERS: FilterType[] = ['All', 'Active', 'Completed'];
--- a/src/utils/filterTodos.ts
+++ b/src/utils/filterTodos.ts
@@ -1,0 +1,23 @@
+import { FilterType } from '../types/FilterType';
+import { Todo } from '../types/Todo';
+
+type Queries = {
+  filter: FilterType;
+};
+
+export const filterTodos = (todos: Todo[], { filter }: Queries) => {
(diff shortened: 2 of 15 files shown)


Valeraom:
Adding isn't completed, I just started)

coldmelnyk (approved):
Good work!

src/mocks.ts
```diff
@@ -0,0 +1,3 @@
+import { FilterType } from './types/FilterType';
+
+export const FILTERS: FilterType[] = ['All', 'Active', 'Completed'];
```
BudnikOleksii, line 3: You can create type from an array
https://stackoverflow.com/questions/59857223/how-to-convert-typescript-types-of-strings-to-array-of-strings

src/utils/filterTodos.ts
```diff
@@ -0,0 +1,23 @@
...
+      switch (filter) {
+        case 'Active':
+          return !todo.completed;
+        case 'Completed':
+          return todo.completed;
+      }
+    });
+  }
+
+  return filteredTodos;
```
BudnikOleksii, line 22:
```suggestion
  if (filter !== 'All') {
    return todos.filter(todo => {
      switch (filter) {
        case 'Active':
          return !todo.completed;
        case 'Completed':
          return todo.completed;
      }
    });
  }

  return todos;
```
You don't need to create a copy as `filter` doesn't mutate an array

BudnikOleksii (approved):
Good work 👍 
Left a few suggestions for the future, check them

PR #1568, opened 2024-10-19
Dev

MaxF1996:
- [DEMO LINK](https://MaxF1996.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,10 +1,56 @@
...
+
+  useEffect(() => {
+    setFilteredTodos(todos);
+  }, [todos]);
+
+  const onFilterChange = (filter: Filters) => {
+    setCurrentFilter(filter);
+
+    switch (filter) {
+      case 'all':
```

src/App.tsx
```diff
@@ -34,170 +82,120 @@ export const App: React.FC = () => {
...
+              <a
+                href="#/completed"
+                className={classNames('filter__link', {
+                  selected: currentFilter === 'completed',
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => onFilterChange('completed')}
+              >
+                Completed
+              </a>
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
Fix the button position, check it on the working example 
image

It's bad practice to write so much code in one component, you need to split the logic into different components, for example `Header, Footer, TodoItem, TodoList`

MaxF1996:
I had to disable the linter for label>input because of false positives

volodymyr-soltys97 (approved):
Great! 🔥

PR #1570, opened 2024-10-22
task solution

olshum8:
[DEMO LINK](https://olshum8.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,61 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { addTodos, getTodos, USER_ID } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Todo } from './types/Todo';
+import { ErrorMessage } from './components/ErrorMessage';
+import { TodoForm } from './components/TodoForm';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [renderCount, setRenderCount] = useState(0);
+
+  const increaseCounter = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+          case 'completed':
+            filteredTodos = todosFromServer.filter(
+              todo => todo.completed === true,
+            );
+            break;
+          default:
+            filteredTodos = todosFromServer;
+            break;
+        }
+
```
etojeDenys: you don't need to filter the array here, just save the todosFromServer in todos state and create some var in component
```
const filteredTodos = getFilteredTodos()
```
move switch case into the getFilteredTodos function

src/components/TodoForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const resetForm = () => {
+    setTodo('');
+  };
+
+  const handleSubmit = async (evnt: React.FormEvent<HTMLFormElement>) => {
+    evnt.preventDefault();
+
+    if (!todo.trim()) {
+      handleError('Title should not be empty');
+      increaseCounter();
```
etojeDenys: I think we don't need this counter logic at all in this task

src/components/TodoList.tsx
```diff
@@ -0,0 +1,39 @@
...
+  todos: Todo[];
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return todos.map(todo => (
+    // eslint-disable-next-line react/jsx-key
+    <div
+      key={todo.id}
+      data-cy="Todo"
+      className={classNames('todo', todo.completed && 'completed')}
```
etojeDenys:
```suggestion
      className={classNames('todo', {'completed': todo.completed})}
```

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,204 +1,77 @@
...
+      setTodos(prevTodos => [...prevTodos, newTodo]);
+    } catch (error) {
+      setErrorMessage('Unable to add a todo');
+      throw error;
+    }
+  };
+
+  const getFilteredTodos = () => {
+    switch (filter) {
+      case 'active':
```
anastasiiavorobiova, line 30: It's better to use constants or enum for this

anastasiiavorobiova (approved):

PR #1579, opened 2024-11-04
Solution

anna-agerone:
[DEMO LINK](https://anna-agerone.github.io/react_todo-app-loading-todos/)


SanyaBratashchuk (changes requested):
1. image
handle plural it should be `1 item left` but `2 items left`

SanyaBratashchuk (approved):

PR #1584, opened 2024-11-13
Develop

svitlanak31:
- [DEMO LINK](https://svitlanak31.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,21 +34,16 @@ export const App: React.FC = () => {
     return undefined;
   }, [error]);
 
-  const filteredTodos = todos.filter(todo => {
-    if (statusFilter === 'All') {
-      return true;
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
+
+    if (statusFilter === 'Active') {
+      return !todo.completed;
+    }
+
+    if (statusFilter === 'Completed') {
+      return todo.completed;
+    }
+
+    return true;
```
SanyaBratashchuk: use switch case here and mov filter under cases to avoid extra iteration in 'All' case

SanyaBratashchuk (changes requested):

etojeDenys (approved):
lgtm

PR #1586, opened 2024-11-14
Finishing

italomagno:
[DEMO LINK](https://italomagno.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,78 @@
+/* eslint-disable max-len */
+/* eslint-disable padding-line-between-statements */
+/* eslint-disable @typescript-eslint/semi */
+/* eslint-disable @typescript-eslint/indent */
+/* eslint-disable react-hooks/rules-of-hooks */
+/* eslint-disable prettier/prettier */
+/* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { deleteTodo, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+
+  const [todos, setTodos] = useState<Todo[]>([
+    {
+      id: 1,
+      userId: 1,
+      title: 'Completed Todo',
+      completed: true,
+    },
+    {
+      id: 2,
+      userId: 1,
+      title: 'Not Completed Todo',
+      completed: false,
+    },
+    {
+      id: 4,
+      userId: 1,
+      title: 'Todo is being saved now',
+      completed: false,
+    },
+  ]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [error, setError] = useState<string | null>(null);
+  const [editingTodoId, setEditingTodoId] = useState<number[]>([]);
+
+  async function loadAllTodos() {
+    try {
+      const loadedTodos = await getTodos();
+      setTodos(loadedTodos);
+      setFilteredTodos(loadedTodos);
+    } catch (e) {
+      setError('Unable to load todos');
+    }
+
+  }
+
+
+
+  async function handleRemoveTodo(todoId: number) {
+    try {
+      setEditingTodoId([...editingTodoId, todoId]);
+      const deletedTodo = await deleteTodo(todoId);
+      setEditingTodoId(editingTodoId.filter((id) => id !== todoId));
+      loadAllTodos();
+    } catch (e) {
+      setError('Unable to delete a todo');
+    }
+  }
+
+  useEffect(() => {
+    loadAllTodos();
+  },[]);
+
+  useEffect(() => {
+      setTimeout(()=>{
+        setError(null)
+      },3000)
+  },[error])
+
   if (!USER_ID) {
     return <UserWarning />;
   }
@@ -34,6 +102,42 @@ export const App: React.FC = () => {
         </header>
 
         <section className="todoapp__main" data-cy="TodoList">
+          {
+            filteredTodos.map((todo) => (
+              <div key={todo.id} data-cy="Todo" className="todo completed">
+              <label className="todo__status-label">
+                <input
+                  data-cy="TodoStatus"
+                  type="checkbox"
+                  className="todo__status"
+                  checked
+                />
+              </label>
+
+              <span data-cy="TodoTitle" className="todo__title">
+                {todo.title}
+              </span>
+
+              {/* Remove button appears only on hover */}
+              <button onClick={() => handleRemoveTodo(todo.id)} type="button" className="todo__remove" data-cy="TodoDelete">
+                ×
+              </button>
+
+              {/* overlay will cover the todo while it is being deleted or updated */}
+            <div data-cy="TodoLoader" className={`modal overlay ${editingTodoId.includes(todo.id) ? "is-active" : ''}`}>
+                <div className="modal-background has-background-white-ter" />
+                <div className="loader" />
+              </div>
+            </div>
+
+            ))
+          }
+
+
+          <hr /> {/* remove this line after */}
+          Examples of different todo states {/* remove this line after */}
+          <hr /> {/* remove this line after */}
+
           {/* This is a completed todo */}
           <div data-cy="Todo" className="todo completed">
             <label className="todo__status-label">
@@ -55,7 +159,7 @@ export const App: React.FC = () => {
             </button>
 
             {/* overlay will cover the todo while it is being deleted or updated */}
-            <div data-cy="TodoLoader" className="modal overlay">
+          <div data-cy="TodoLoader" className="modal overlay is-active">
               <div className="modal-background has-background-white-ter" />
               <div className="loader" />
             </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,7 +9,7 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { deleteTodo, getTodos, USER_ID } from './api/todos';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [editingTodoId, setEditingTodoId] = useState<number[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [newTodoTitle, setNewTodoTitle] = useState('');
+
+  async function loadAllTodos() {
+    try {
+      const loadedTodos = await getTodos();
```
raulriato, line 20: It's not good practice to have this much code on your `App.jsx`

The `App` should only display your components. 

A better approach would be to have the general logic, like the `loadAllTodos()`, `handleToggleAllTodos()`, etc. in your `TodoList.jsx`

raulriato: Also, The main goal of using react is to have your code distributed inside smaller and reusable components, so you should have more than just the component `TodoList`. For exemplo, you could have a `Todo.jsx`: which would take care of each one of the items of the list of todos. You could also have a `TodoInput` that would contain the code related to the input on which the user will add a todo to the list, and so on.

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+} from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [editingTodoId, setEditingTodoId] = useState<number[]>([]);
```
raulriato: the `editingTodoId` variable is an array, right? So it's name should be plural.

`const [editingTodoIds, setEditingTodoIds] ...`

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+
+        await Promise.all(updatedTodos.map(t => updateTodo(t)));
+        setEditingTodoId([]);
+        loadAllTodos();
+
+        return;
+      } else {
+        const todosToEdit = todos.filter(t => !t.completed);
+
+        setEditingTodoId([...editingTodoId, ...todosToEdit.map(t => t.id)]);
```
raulriato: You can improve this function. It doesn't need to have all these verifications. instead of verifying if ALL the todos are completed, or if ALL of them are incomplete. You could just put a simple verification of your `todosToEdit`'s length in here.

For example, you could use in the lline 58:

`
if (todosToEdit.length === 0) {
  //write the code for the case where all todos are complete here
}
`

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+
+      setEditingTodoId([...editingTodoId, newTodoIdAvailable]);
+      const newTodo: Todo = {
+        id: newTodoIdAvailable,
+        title: newTodoTitle,
+        completed: false,
+        userId: USER_ID,
+      };
+
+      setFilteredTodos([...todos, newTodo]);
```
raulriato: This is not quite correct. When the user is using the `Completed` filter, setting the filtered todos this way will make the list of todos be shown to the user with the `Completed` filter on. You should only add the new todo to the filtered todos if the filter `Active` is on, otherwise, just add it to the list of todos.

src/components/TodoList.tsx
```diff
@@ -0,0 +1,102 @@
...
+                type="checkbox"
+                className="todo__status"
+                checked={todo.completed}
+                onClick={() => handleUpdateTodoStatus(todo.id)}
+              />
+            </label>
+
+            {!selectedTodo ? (
+              <span
+                onDoubleClick={() => setSelectedTodo(todo)}
```
raulriato: Here your code verifies if a `selectedTodo` exists, so if any todo is selected, it makes all todos selected, instead, you should verify if there is a `selectedTodo` AND if the `selectedTodo` has the same id as the `todo` you're rendering, for example

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+      const todo = todos.find(t => t.id === todoId);
+
+      setEditingTodoId([...editingTodoId, todoId]);
+      if (!todo) {
+        throw new Error('Todo not found');
+      }
+
+      const updatedTodo: Todo = { ...todo, completed: !todo.completed };
+      const updatedTodoFromServer = await updateTodo(updatedTodo);
+      const newTodos = todos.map(t =>
```
raulriato: a better name for this variable would be `updatedTodos`, the list of todos is not new, just updated

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+        return;
+      }
+    } catch (e) {
+      setError('Unable to update todo');
+      setEditingTodoId([]);
+      loadAllTodos();
+    }
+  }
+
+  function handleNewTodoTitle(e: React.ChangeEvent<HTMLInputElement>) {
```
raulriato: The name of this function does not sendo a clear message of what it does. Try using, for example `handleAddNewTodo`.

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
+      await deleteTodo(todoId);
+      setEditingTodoId(editingTodoId.filter(id => id !== todoId));
+      loadAllTodos();
+    } catch (e) {
+      setError('Unable to delete a todo');
+      setEditingTodoId(editingTodoId.filter(id => id !== todoId));
+    }
+  }
+
+  async function handleSubmitForm(
```
raulriato: This function name is ambiguous, it's not clear if it's meant to the input that adds a new todo, or if it's to update a todo.

src/App.tsx
```diff
@@ -1,203 +1,364 @@
...
       <div className="todoapp__content">
         <header className="todoapp__header">
           {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
+          {todos.length > 0 && (
+            <button
```
raulriato: as I said before, try distributing all elements into components separated from the `App`, and use `App` only to render those components.

src/components/TodoList.tsx
raulriato: Also, instead of using this `<span>` on this map, try isolating the code of one todo to a component `Todo` in a way that here you'll have:

image


Not exactly like the picture, it's just an idea for you to work on, create more components. The goal is to have components that do one thing, and just one thing. This component `TodoList` should handle the list, and only the list, not each todo individually

raulriato (changes requested):
The app is working almost perfectly, good job!

Just need some changes to have a better code.

italomagno:
Hey mr @raulriato thank you for you time and your improvements in my code, but it is an activity very large, I've tried to solve all code in one time! You'll see in the next exercises that I've "componentizated" all my code in small pieces. but of course there is a lot of ambiguities in it. Of course I can change this piece of code, but please, I would appreciate it if you could take this into account. My next exercises in this module is better than here.

src/App.tsx
italomagno: hey mr @raulriato, this function do exactly this. It change the state of newTodoTitle.

src/components/TodoList.tsx
italomagno: I've tried to change to your suggestion, but i've noticed that it is written in a way that make more sense, i "avoided" handle using "Ids" in this case, because the user only can  handle one edit per time.  Thus, adding the Todo in a selectedTodo when the user make a double click in this span, trigger the toggle to the form. Then, in the form, is dealing with the other cases that the user click outside of the form, etc..

raulriato: @italomagno I'm sorry, I sent you a direct message on our platform. That's all on me, I made a mistake while reviewing your code, and used the wrong parameters to review it. I'll correct it the next time.

src/App.tsx
```diff
@@ -1,204 +1,111 @@
...
+      setFilteredTodos(filteredTodos.filter(t => t.completed));
+    }
 
-            <span data-cy="TodoTitle" className="todo__title">
-              Todo is being saved now
-            </span>
+    if (filter === 'active') {
+      setFilteredTodos(filteredTodos.filter(t => !t.completed));
+    }
+  }, [filter, filteredTodos.length, todos.length]);
```
joaorpereira: instead of `filteredTodos.length`

use `filteredTodos`

src/App.tsx
```diff
@@ -1,204 +1,111 @@
...
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
+  useEffect(() => {
+    if (filter === 'completed') {
+      setFilteredTodos(filteredTodos.filter(t => t.completed));
```
joaorpereira: Use this approach:
`setFilteredTodos(prev => prev.filter(t => t.completed));`

src/App.tsx
```diff
@@ -1,204 +1,111 @@
...
+    }
 
-            <span data-cy="TodoTitle" className="todo__title">
-              Todo is being saved now
-            </span>
+    if (filter === 'active') {
+      setFilteredTodos(filteredTodos.filter(t => !t.completed));
```
joaorpereira: Use this approach:
`setFilteredTodos(prev => prev.filter(t => !t.completed));`

joaorpereira (changes requested):

src/App.tsx
italomagno: setFilteredTodos(prev => prev.filter(t => t.completed));

this causing the error below:

Argument of type '(prev: Todo[]) => Todo[]' is not assignable to parameter of type 'Todo[]'.ts(2345)
Footer.tsx(34, 26): Did you mean to call this expression?

so, i think that is not a good approach, did you try to use this in my code? or just a suggestion? @joaorpereira

julianFBarbosa:
Based on the comments of the mentor and your recent changes, I would approve it!

the only issue is that I can't approve it right away because I'm not a reviewer =/

Despite that, congrats =)

raulriato (approved):
Good job!

joaorpereira (approved):

PR #1590, opened 2024-11-20
add task solution

BohdanDymydiuk:
[DEMO LINK](https://BohdanDymydiuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,7 +18,6 @@ enum Filter {
 export const App: React.FC = () => {
   const [value, setValue] = useState<string>('');
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [counter, setCounter] = useState<number>(0);
   const [filter, setFilter] = useState<string>(Filter.all);
   const filteredTodos = useMemo(() => {
     return todos.filter(todo => {
@@ -49,19 +48,7 @@ export const App: React.FC = () => {
     }, 100);
   }, []);
 
-  useEffect(() => {
-    let counterOfActive: number = 0;
-
-    if (todos.length > 0) {
-      todos.forEach(todo => {
-        if (!todo.completed) {
-          counterOfActive++;
-        }
-      });
-    }
-
-    setCounter(counterOfActive);
-  }, [todos]);
+  const nrOfActiveTodos = todos.filter(t => !t.completed).length;
 
   function onSubmit(event: React.FormEvent) {
     event.preventDefault();
@@ -106,7 +93,7 @@ export const App: React.FC = () => {
         {/* Hide the footer if there are no todos */}
         {todos.length > 0 && (
           <Footer
-            counter={counter}
+            nrOfActiveTodos={nrOfActiveTodos}
             filter={filter}
             allSelected={allSelected}
             activeSelected={activeSelected}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -75,44 +75,20 @@ export const App: React.FC = () => {
 
   function allSelected(event: React.MouseEvent<HTMLAnchorElement>) {
     event.preventDefault();
-    const { target } = event;
 
     setFilter(Filter.all);
-
-    if (target instanceof HTMLAnchorElement) {
-      target.classList.add('selected');
-    }
   }
 
   function activeSelected(event: React.MouseEvent<HTMLAnchorElement>) {
     event.preventDefault();
-    const { target } = event;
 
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,10 +1,111 @@
...
+  const [value, setValue] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [counter, setCounter] = useState<number>(0);
+  const [filter, setFilter] = useState<string>('All');
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filter) {
+        case 'All':
+          return todo;
+        case 'Active':
```
volodymyr-soltys97: Create a enum for 'All', 'Active', 'Completed' and use it everywhere

src/App.tsx
```diff
@@ -1,10 +1,111 @@
...
+          return todo.completed === false;
```
volodymyr-soltys97:
```suggestion
          return !todo.completed;
```

src/App.tsx
```diff
@@ -1,10 +1,111 @@
...
+        .catch(() => {
+          setError('Unable to load todos');
+
+          setTimeout(() => setError(''), 3000);
+        });
+    }, 100);
+  }, []);
+
+  useEffect(() => {
+    let counter2: number = 0;
```
volodymyr-soltys97: Use a more descriptive name

src/App.tsx
```diff
@@ -1,10 +1,111 @@
...
+
+    if (todos.length > 0) {
+      todos.forEach(todo => {
+        if (todo.completed === false) {
```
volodymyr-soltys97:
```suggestion
        if (!todo.completed) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is a bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoList, TodoItem`

src/App.tsx
```diff
@@ -1,10 +1,120 @@
...
+    if (todos.length > 0) {
+      todos.forEach(todo => {
+        if (!todo.completed) {
+          counterOfActive++;
+        }
+      });
+    }
+
+    setCounter(counterOfActive);
+  }, [todos]);
```
etojeDenys: simplify it using the reduce or filter without the state

src/App.tsx
```diff
@@ -1,10 +1,120 @@
...
+  function allSelected(event: React.MouseEvent<HTMLAnchorElement>) {
+    event.preventDefault();
+    const { target } = event;
+
+    setFilter(Filter.all);
+
+    if (target instanceof HTMLAnchorElement) {
+      target.classList.add('selected');
+    }
+  }
```
etojeDenys: looks redundant

etojeDenys (changes requested):

src/App.tsx
BohdanDymydiuk: but I have to update the counter, don't I?

src/App.tsx
BohdanDymydiuk: no, it isn't redudant

VitaliyBondarenko1982: set selected class dynamically, based on filter state. 
no need to work with dom directly in this case.

src/App.tsx
```diff
@@ -1,10 +1,120 @@
...
+
+  function completedSelected(event: React.MouseEvent<HTMLAnchorElement>) {
+    event.preventDefault();
+    const { target } = event;
+
+    setFilter(Filter.completed);
+
+    if (target instanceof HTMLAnchorElement) {
+      target.classList.add('selected');
+    }
```
VitaliyBondarenko1982: same here. All such cases are redundant. Use state and other data for dynamic managing classNames.

VitaliyBondarenko1982 (changes requested):
Almost done.
Please check comments.

src/App.tsx
BohdanDymydiuk: oooh, thanks. I didn't notice it

src/App.tsx
etojeDenys:
```suggestion  
  const nrOfActiveTodos = todos.filter(t => !t.completed).length;
```

etojeDenys (approved):
looks good in general, but don't forget to fix that comment with counter in next part

PR #1594, opened 2024-11-27
add first part react todo

moskkat40:
[DEMO LINK](https://moskkat40.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,72 @@
+/* eslint-disable max-len */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import * as servisesTodos from './api/todos';
+import { Todo } from './types/Todo';
+import { AddTodo } from './components/AddTodo/AddTodo';
+import { TodoList } from './components/TodoList/TodoList';
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { FooterFiltering } from './components/FooterFiltering/FooterFiltering';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('All');
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    servisesTodos
+      .getTodos()
+      .then(setTodos)
+      .catch(error => {
+        setErrorMessage('Unable to load todos');
+        throw error;
+      })
+      .finally(() =>
+        setTimeout(() => {
+          setErrorMessage('');
+        }, 3000),
+      );
+  }, []);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filteredTodos = useMemo(() => {
+    if (filter === 'Active') {
+      return todos.filter(todo => !todo.completed);
+    } else if (filter === 'Completed') {
+      return todos.filter(todo => todo.completed);
+    }
+
+    return todos;
+  }, [filter, todos]);
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
+        <AddTodo />
+
+        <TodoList filteredTodos={filteredTodos} />
+        {todos.length > 0 && (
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,204 +1,72 @@
...
   }
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    servisesTodos
+      .getTodos()
+      .then(setTodos)
+      .catch(error => {
+        setErrorMessage('Unable to load todos');
+        throw error;
```
TarasHoliuk: do not throw an error here - you handled it already, so there is no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

src/App.tsx
```diff
@@ -1,204 +1,72 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('All');
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
TarasHoliuk: remove it
instead - fix the reason of this lint error (move hook above `if` block)

src/App.tsx
```diff
@@ -1,204 +1,72 @@
...
+import { Todo } from './types/Todo';
+import { AddTodo } from './components/AddTodo/AddTodo';
+import { TodoList } from './components/TodoList/TodoList';
+import { ErrorNotification } from './components/ErrorNotification/ErrorNotification';
+import { FooterFiltering } from './components/FooterFiltering/FooterFiltering';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('All');
```
TarasHoliuk: Create a union type for it and use it as generic here
```suggestion
  const [filter, setFilter] = useState<Filter>('All');
```

src/App.tsx
```diff
@@ -1,204 +1,72 @@
...
+          setErrorMessage('');
+        }, 3000),
+      );
+  }, []);
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const filteredTodos = useMemo(() => {
+    if (filter === 'Active') {
+      return todos.filter(todo => !todo.completed);
+    } else if (filter === 'Completed') {
```
TarasHoliuk:
```suggestion
    }
    
    if (filter === 'Completed') {
```
you return inside, so `else` is redundant here - it's easier to read code without it

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,51 @@
...
+  errorMessage,
+  setErrorMessage,
+}) => {
+  {
+    /* DON'T use conditional rendering to hide the notification */
+  }
+
+  {
+    /* Add the 'hidden'  to hide the message smoothly */
+  }
```
TarasHoliuk: You can delete these comments now - it was hints for you 🙂

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      <button
+        onClick={handleCloseNotification}
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+      />
+      {errorMessage}
+      {/* show only one message at a time */}
+      {/* Title should not be empty
```
TarasHoliuk: delete comments

src/components/FooterFiltering/FooterFiltering.tsx
```diff
@@ -0,0 +1,73 @@
...
+  todos: Todo[];
+};
+
+export const FooterFiltering: React.FC<Props> = ({
+  filter,
+  setFilter,
+  todos,
+}) => {
+  function handleFilter(event: React.MouseEvent<HTMLElement>) {
+    setFilter(event.currentTarget.textContent || '');
```
TarasHoliuk: in case you use the union type for `filter` state you should use typecasting here

src/components/FooterFiltering/FooterFiltering.tsx
```diff
@@ -0,0 +1,73 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  filter: string;
+  setFilter: (a: string) => void;
+  todos: Todo[];
+};
+
+export const FooterFiltering: React.FC<Props> = ({
```
TarasHoliuk: just` Footer` (preferable since the parent jsx element you return here is footer), or just `Filter`

src/components/AddTodo/AddTodo.tsx
```diff
@@ -0,0 +1,22 @@
+export function AddTodo() {
```
TarasHoliuk: AddTodo (with verb) sounds like a function and not a component. Rename it to TodoForm or TodoInput or even Header (you return header here)

TarasHoliuk (changes requested):
GJ! Let's improve it a bit

VitaliyBondarenko1982 (approved):
Looks good.

PR #1595, opened 2024-11-28
add solution (part 1)

associate2coder:
[DEMO LINK](https://associate2coder.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,56 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -1,10 +1,56 @@
...
+
+      default: {
+        return todos;
+      }
+    }
+  }, [filterQuery, todos]);
+
+  // COUNT TODOs
+  const activeTodoCount = useMemo(() => {
+    return [...todos].filter(todo => !todo.completed).length;
```
etojeDenys: it's redundant to copy the array, the filter method doesn't mutate an array

etojeDenys (approved):

src/App.tsx
associate2coder: Yeap. Thanks! My bad)

PR #1597, opened 2024-12-02
solution

anna-daryna:
[DEMO LINK](https://anna-daryna.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,24 @@
 import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 
 import { UserWarning } from './UserWarning';
 import { getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
+import Header from './components/Header';
+import TodoList from './components/TodoList';
+import Footer from './components/Footer';
+
+export enum TodoStatus {
+  All = 'All',
+  Active = 'Active',
+  Completed = 'Completed',
+}
 
 export const App: React.FC = () => {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,10 +1,53 @@
...
+    }
+
+    return undefined;
+  }, [errorMessage]);
+
+  const filteredTodos = (() => {
+    switch (statusFilter) {
+      case 'Active':
+        return todos.filter(todo => !todo.completed);
+      case 'Completed':
```
volodymyr-soltys97: Create a enum for 'All', 'Active', 'Completed' and use it everywhere

src/App.tsx
```diff
@@ -34,170 +75,103 @@ export const App: React.FC = () => {
...
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
+          {filteredTodos.map(todo => (
+            <div
+              data-cy="Todo"
+              className={`todo ${todo.completed ? 'completed' : ''}`}
```
volodymyr-soltys97: Use the classnameslibrary for add classes with condition, fix it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

etojeDenys (approved):
good job in general

PR #1603, opened 2024-12-05
Develop

VikaChereushenko:
[DEMO LINK](https://VikaChereushenko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,13 +6,7 @@ import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import { getTodos, addTodo, deleteTodo } from './api/todos';
-
-function findMaxId(todos: Todo[]) {
-  const allIds = todos.map(todo => todo.id);
-
-  return Math.max(...allIds) + 1;
-}
+import { getTodos } from './api/todos';
 
 function filterTodos(todos: Todo[], status?: string | null) {
   const todosCopy = [...todos];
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+    return todosCopy.filter(todo => todo.completed === true);
+  }
+
+  return todosCopy;
+}
 
 export const App: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
etojeDenys: you don't need the separate state for filtered todos

src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+
+    setTimeout(() => setErrorMessage(''), 3000);
+  }, []);
+
+  const handleFiltration = (
+    e: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    const status = e.currentTarget.textContent?.toLowerCase();
+
+    setFilteredTodos(filterTodos(todos, status));
```
etojeDenys: you can just write in component
```
const filteredTodos = filterTodos(todos, status)
```

etojeDenys (changes requested):

src/App.tsx
VikaChereushenko: Thanks. you're right, I remember doing it in the previous task, but again got confused. Will fix it

VikaChereushenko:
there is an issue with the lable 'A form label must be associated with a control', so I haven't passed some checks

src/App.tsx
```diff
@@ -1,10 +1,45 @@
...
+function filterTodos(todos: Todo[], status?: string | null) {
+  const todosCopy = [...todos];
+
+  if (status === 'active') {
+    return todosCopy.filter(todo => todo.completed === false);
+  }
+
+  if (status === 'completed') {
+    return todosCopy.filter(todo => todo.completed === true);
+  }
```
etojeDenys, line 23: it's better to rewrite it using the switch case

src/api/todos.ts
```diff
@@ -1,10 +1,18 @@
...
 
-export const USER_ID = 0;
+export const USER_ID = 2042;
 
 export const getTodos = () => {
   return client.get<Todo[]>(`/todos?userId=${USER_ID}`);
 };
 
+export const addTodo = (todoData: Todo) => {
+  return client.post<Todo[]>(`/todos`, todoData);
```
etojeDenys, line 11:
```suggestion
  return client.post<Todo[]>('/todos', todoData);
```

etojeDenys (approved):
good job

src/api/todos.ts
VikaChereushenko: Done) I've changed the next file with the same code

PR #1607, opened 2024-12-12
solution

VasylPylypchynets:
[DEMO LINK](https://VasylPylypchynets.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,12 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
-import { getTodos } from './api/todos';
+import { deleteTodo, getTodos } from './api/todos';
 import { Todo } from './types/Todo';
-// import { UserWarning } from './UserWarning';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Error } from './components/ErrorMessage/ErrorMessage';
 
 export const App: React.FC = () => {
   const [query, setQuery] = useState<string>('');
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,13 +1,71 @@
...
+  // const [isLoadingChange, setIsLoadingChange] = useState(false);
+
+  // if (!USER_ID) {
+  //   return <UserWarning />;
+  // }
+
+  // function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+  //   e.preventDefault();
+  //   setNewTask(query);
+  // }
```
volodymyr-soltys97: Remove comments
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoList, TodoItem`

src/App.tsx
```diff
@@ -1,204 +1,120 @@
...
+import { Todo } from './types/Todo';
+import { Header } from './components/Header/Header';
+import { TodoList } from './components/TodoList/TodoList';
+import { Footer } from './components/Footer/Footer';
+import { Error } from './components/ErrorMessage/ErrorMessage';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [query, setQuery] = useState<string>('');
```
etojeDenys, line 12: don't specify types for primitives, it's redundant

src/App.tsx
```diff
@@ -1,204 +1,120 @@
...
+        }
+      }
+    }
+
+    const cleanup = deleteTodoFromServer();
+
+    return () => {
+      if (cleanup instanceof Function) {
+        cleanup();
+      }
```
etojeDenys, line 48:
```suggestion
    const cleanup = deleteTodoFromServer();

    return () => {
      deleteTodoFromServer()
```

etojeDenys (approved):

PR #1612, opened 2024-12-14
Develop

bhoriunov:
[DEMO LINK](https://bhoriunov.github.io/react_todo-app-loading-todos/)

Maybe the code is a bit confusing, but I tried. While I "was lost" and couldn't solve the problem, I go a little forward (adding and deleting, check these also, please)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,11 +21,11 @@ export const App: React.FC = () => {
       try {
         const fetchedTodos = await getTodos();
 
-        setTodos(fetchedTodos);
+        setTimeout(() => setTodos(fetchedTodos), 300);
       } catch {
         setError('Unable to load todos');
       } finally {
-        setIsLoading(false);
+        setTimeout(() => setIsLoading(false), 300);
       }
     };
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,203 +1,255 @@
...
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      setIsLoading(true);
+      setError(null);
+
+      try {
+        const fetchedTodos = await getTodos();
+
+        setTimeout(() => setTodos(fetchedTodos), 300);
```
etojeDenys, line 25: timeout looks redundant here

src/App.tsx
```diff
@@ -1,203 +1,255 @@
...
+      } catch {
+        setError('Unable to load todos');
+      } finally {
+        setTimeout(() => setIsLoading(false), 300);
```
etojeDenys, line 29: .

etojeDenys (approved):
good job in general, but you will rewrite todo add and delete logic in next task

PR #1632, opened 2024-12-17
Develop

DmytroHoncharuk:
[DEMO LINK](https://DmytroHoncharuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
+/*eslint-disable*/
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
+
 import React, { useEffect } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
-import React from 'react';
+
+import React, { useEffect } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
-
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+import { Simulate } from 'react-dom/test-utils';
```
volodymyr-soltys97: Why did you import it?
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code

src/App.tsx
DmytroHoncharuk: didn't do that, maybe, some automatically import worked

src/App.tsx
```diff
@@ -1,14 +1,88 @@
...
+import { Footer } from './components/Footer/Footer';
+import { TodoList } from './components/TodoList/TodoList';
 
 export const App: React.FC = () => {
+
+  type Filters = 'All' | 'Active' | 'Completed';
+  type ErrorObject = '' | 'Load' | 'Add';
+
+  const [todosData, setTodosData] = React.useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = React.useState<Todo[]>([]);
```
etojeDenys, line 26: you don't need separate state for this

src/App.tsx
```diff
@@ -1,14 +1,88 @@
...
+            return !todo.completed;
+          case 'Completed':
+            return todo.completed;
+          default:
+            return true;
+        }
+      });
+
+      setFilteredTodos(temp);
+    };
```
etojeDenys, line 70:
```suggestion
    const filterTodos = () => {
      const temp = todosData.filter(todo => {
        switch (filter) {
          case 'All':
            return true;
          case 'Active':
            return !todo.completed;
          case 'Completed':
            return todo.completed;
          default:
            return true;
        }
      });

  const filteredTodos = filterTodos()
```

etojeDenys (approved):

PR #1633, opened 2024-12-17
add task solution

hrebinets:


src/App.tsx
```diff
@@ -1,10 +1,121 @@
...
+      return true;
+    }
+
+    if (selectedFilter === 'completed') {
+      return todo.completed;
+    }
+
+    if (selectedFilter === 'active') {
+      return !todo.completed;
+    }
```
etojeDenys, line 113: better to rewrite it using the switch case

etojeDenys (approved):
good job

PR #1643, opened 2024-12-18
add task solution

LevytskyiV:
[DEMO LINK](https://LevytskyiV.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,118 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos, addTodo, deleteTodo } from './api/todos';
+
+import { Todo } from './types/Todo';
+import { TodoList } from './сomponents/TodoList';
+import { Footer } from './сomponents/Footer';
+
+export enum Filter {
+  ALL = 'All',
+  ACT = 'Active',
+  COMP = 'Completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<Filter>(Filter.ALL);
+  const [newTitle, setNewTitle] = useState('');
+  const [loading, setLoading] = useState(false);
+
+  const checkComplete = todos.length > 0 && todos.every(todo => todo.completed);
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,10 +1,118 @@
...
+import { USER_ID, getTodos, addTodo, deleteTodo } from './api/todos';
+
+import { Todo } from './types/Todo';
+import { TodoList } from './сomponents/TodoList';
+import { Footer } from './сomponents/Footer';
+
+export enum Filter {
+  ALL = 'All',
+  ACT = 'Active',
+  COMP = 'Completed',
```
volodymyr-soltys97: Move enum for the separate file
```suggestion
export enum Filter {
  All = 'All',
  Active = 'Active',
  Completed = 'Completed',
```

src/сomponents/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+          Active
+        </a>
+        <a
+          href="#/completed"
+          className={`filter__link ${filter === Filter.COMP ? 'selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFilter(Filter.COMP)}
+        >
+          Completed
+        </a>
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

src/сomponents/TodoList.tsx
```diff
@@ -0,0 +1,58 @@
...
+              data-cy="TodoStatus"
+              type="checkbox"
+              className="todo__status"
+              checked={todo.completed}
+              onChange={() => onToggleTodo(todo.id)}
+            />
+          </label>
+
+          <span data-cy="TodoTitle" className="todo__title">
+            {todo.title}
```
volodymyr-soltys97: Use destructuring for `todo`

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write so much code in one component, you need to separate the logic into different components, for example `Header, TodoItem`

src/сomponents/Footer.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${activeTodos} items left`}
+      </span>
+
+      <nav className="filter" data-cy="Filter">
+        {filters.map(({ value, label, href, dataCy }) => (
+          <a
+            key={value}
+            href={href}
+            className={`filter__link ${filter === value ? 'selected' : ''}`}
```
volodymyr-soltys97, line 40: Don't forget to use classnames library everywhere for add classes with condition

volodymyr-soltys97 (approved):
Good job! 👍

PR #1661, opened 2024-12-30
solution

RUBIC0N5545:
https://RUBIC0N5545.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,17 +2,18 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { createTodo, deleteTodo, updateTodo, USER_ID } from './api/todos';
 import { getTodos } from './api/todos';
 import { Todo } from './types/Todo';
-import { client } from './utils/fetchClient';
+import Header from './Components/Header';
+import { FilterType } from './types/FilterType';
+import Footer from './Components/Footer';
+import TodoList from './Components/TodoList';
 
 interface OnDblClickParams {
   index: number;
   todo: Todo;
 }
-
-type FilterType = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
   const [todosDb, setTodosDb] = useState<Todo[]>([]);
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,203 +1,324 @@
...
+  const [newTodoTitle, setNewTodoTitle] = useState<string>('');
+
+  useEffect(() => {
+    setTodosToShow(
+      todosDb.filter(todo => {
+        if (selectedFilter === 'all') {
+          return true;
+        }
+
+        if (selectedFilter === 'active') {
```
volodymyr-soltys97: Use switch case here and move this logic from useEffect to variable

src/App.tsx
```diff
@@ -1,203 +1,324 @@
...
+          return !todo.completed;
+        }
+
+        return todo.completed;
+      }),
+    );
+  }, [todosDb, selectedFilter]);
+
+  useEffect(() => {
+    // setLoading(true);
```
volodymyr-soltys97: Remove all comments
```suggestion
```

src/App.tsx
```diff
@@ -1,203 +1,324 @@
...
+      setLoadingTodoId(null);
+    }
+  };
+
+  const completeTodo = async (id: number, completed: boolean) => {
+    setLoadingTodoId(id);
+    setError(null);
+    try {
+      await client.patch(`/todos/${id}`, { completed });
+      setTodosDb(
```
volodymyr-soltys97: Move methods to the api file 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
It is bad practice to write all the code in one component, it is hard to read now. You need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

src/App.tsx
```diff
@@ -1,203 +1,219 @@
...
+
+  useEffect(() => {
+    setTodosToShow(
+      todosDb.filter(todo => {
+        switch (selectedFilter) {
+          case 'all':
+            return true;
+          case 'active':
+            return !todo.completed;
+          case 'completed':
```

volodymyr-soltys97 (changes requested):
Hi, you need to push your new components for the review

src/App.tsx
```diff
@@ -1,203 +1,221 @@
...
+  }, [todosDb, selectedFilter]);
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      try {
+        setTodosDb(await getTodos());
+      } catch (err) {
+        setError('Unable to load todos');
+      } finally {
+      }
```
vadiimvooo, line 54: finally block can be removed

vadiimvooo (approved):
Well done

PR #1681, opened 2025-01-17
Develop

Shmyhelskiy:


src/servises/filterTodos.ts
```diff
@@ -0,0 +1,15 @@
+import { Todo } from '../types/Todo';
+
+export const filteredTodos = (todos: Todo[], filter: string) => {
+  return todos.filter(todo => {
+    if (filter === 'active') {
```
anastasiiavorobiova, line 5:
```suggestion
    if (filter === 'active') {
```
Consider creating an enum for statuses

anastasiiavorobiova (approved):
Great work!
Try to fix linter error 
Screenshot 2025-01-17 at 13 14 22

PR #1702, opened 2025-01-30
done

myuriy81:
[DEMO LINK](https://myuriy81.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,203 +1,182 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
+import * as todoService from './api/todos';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,203 +1,182 @@
...
+
+  useEffect(loadTodos, [userId]);
+
+  const handleFilter = (currentFilter: string) => {
+    setFilter(currentFilter);
+  };
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'all':
```
anastasiiavorobiova: Consider using an enum here

src/App.tsx
anastasiiavorobiova, line 1: Consider splitting this huge God component

anastasiiavorobiova (changes requested):
Great work! Added a few comments to fix

volodymyr-soltys97 (changes requested):
Hi, It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

etojeDenys (approved):

PR #1712, opened 2025-02-07
load todos

yzhyhaliuk:
[DEMO LINK](https://yzhyhaliuk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,102 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, updateTodo, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,103 @@
...
+        ),
+      );
+    } catch {
+      setErrorMessage('Unable to update a todo');
+    } finally {
+      setUpdatingTodoId(null);
+    }
+  };
+
+  function handleFilter(filterType: string) {
```
volodymyr-soltys97: It is important to follow the same way of creating functions everywhere
```suggestion
  const handleFilter = (filterType: string) => {
```

src/App.tsx
```diff
@@ -1,14 +1,103 @@
...
+    let filteredTodos;
+
+    setActivefilter(filterType);
+
+    switch (filterType) {
+      case 'all': {
```

src/App.tsx
```diff
@@ -1,14 +1,103 @@
...
+      }
+
+      case 'completed': {
+        filteredTodos = allTodos?.filter(todo => todo.completed);
+        break;
+      }
+
+      default: {
+        filteredTodos = allTodos?.filter(todo => !todo.completed);
+      }
```
volodymyr-soltys97:
```suggestion
      default:
        filteredTodos = allTodos;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoList, TodoItem`

anastasiiavorobiova (approved):
Awesome!

PR #1713, opened 2025-02-07
Update README.md

Radymyr:
added missing letter


anastasiiavorobiova (changes requested):
There are no files to check in this PR

PR #1719, opened 2025-02-11
part 1

Manuilenkoart:
[DEMO LINK](https://manuilenkoart.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,60 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useCallback, useEffect, useState } from 'react';
+import { getTodos } from './api/todos';
+import { ErrorNotification, Footer, Header, TodoList } from './components';
+import { ActiveFilter, Todo } from './types';
+// import { todosMock } from './utils/mock';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosFiltered, setTodosFiltered] = useState<Todo[]>([]);
+
+  const [error, setError] = useState('');
+  // const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError('Unable to load todos'));
+  }, []);
+
+  const handleHideError = useCallback(() => setError(''), []);
+
+  const handleFilter = useCallback(
+    (filter: ActiveFilter) => {
+      switch (filter) {
+        case 'active':
+          setTodosFiltered(todos.filter(({ completed }) => !completed));
+          break;
+
+        case 'completed':
+          setTodosFiltered(todos.filter(({ completed }) => completed));
+          break;
+
+        case 'all':
+          setTodosFiltered(todos);
+          break;
+
+        default:
+          break;
+      }
+    },
+    [todos],
+  );
 
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
-        <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
-          <button
-            type="button"
-            className="todoapp__toggle-all active"
-            data-cy="ToggleAllButton"
-          />
+        <Header todos={todos} />
 
-          {/* Add a todo on form submit */}
-          <form>
-            <input
-              data-cy="NewTodoField"
-              type="text"
-              className="todoapp__new-todo"
-              placeholder="What needs to be done?"
-            />
-          </form>
-        </header>
-
-        <section className="todoapp__main" data-cy="TodoList">
-          {/* This is a completed todo */}
-          <div data-cy="Todo" className="todo completed">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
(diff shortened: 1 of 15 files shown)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check after the demo 
image

Manuilenkoart:
> Hi, your demo link isn't working, you need to make a deploy and check after the demo 

fixed

src/App.tsx
```diff
@@ -1,204 +1,60 @@
...
+import { ErrorNotification, Footer, Header, TodoList } from './components';
+import { ActiveFilter, Todo } from './types';
+// import { todosMock } from './utils/mock';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosFiltered, setTodosFiltered] = useState<Todo[]>([]);
```
etojeDenys: its redundant to create state for filteredTodos

src/App.tsx
```diff
@@ -1,204 +1,60 @@
...
+        case 'all':
+          setTodosFiltered(todos);
+          break;
+
+        default:
+          break;
+      }
+    },
+    [todos],
+  );
```
etojeDenys:
```
```
and that function will return the array of todos

src/components/Footer.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FC, memo, useEffect, useMemo, useState } from 'react';
+import { ActiveFilter, Todo } from '../types';
+
+type Props = {
+  todos: Todo[];
+  onFilter: (filter: ActiveFilter) => void;
+};
+
+export const Footer: FC<Props> = memo(({ todos, onFilter }) => {
+  const [activeFilter, setActiveFilter] = useState<ActiveFilter>('all');
```
etojeDenys: we don't need it here, just pass the filter value from app.tsx to footer

etojeDenys (changes requested):

src/components/Footer.tsx
Manuilenkoart: I’ve refactored the code based on the review comments. However, I believe this version [pr link](https://github.com/mate-academy/react_todo-app-loading-todos/commit/23e80418e8df42fc8e9c7ca058fb1a1cb0ea69f4) is better because it keeps the parent component unaware of the child’s logic

src/App.tsx
Manuilenkoart: remove to makeFilterTodos

src/App.tsx
Manuilenkoart: deleted  const [todosFiltered, setTodosFiltered] = useState<Todo[]>([]);

src/types/index.ts
```diff
@@ -4,3 +4,5 @@ export interface Todo {
   title: string;
   completed: boolean;
 }
+
+export type ActiveFilter = 'all' | 'active' | 'completed';
```
volodymyr-soltys97, line 8: Use enum for these values

src/utils/filters.ts
```diff
@@ -0,0 +1,23 @@
...
+  filter: ActiveFilter,
+): Todo[] => {
+  switch (filter) {
+    case 'active':
+      return todos.filter(({ completed }) => !completed);
+
+    case 'completed':
+      return todos.filter(({ completed }) => completed);
+
+    case 'all':
```
volodymyr-soltys97, line 14: Use enum for these words

src/components/TodoList.tsx
```diff
@@ -0,0 +1,156 @@
...
+TodoList.displayName = 'TodoListMemo';
+
+// <section className="todoapp__main" data-cy="TodoList">
+//   {/* This is a completed todo */}
+//   <div data-cy="Todo" className="todo completed">
+//     <label className="todo__status-label">
+//       <input
+//         data-cy="TodoStatus"
+//         id="todoStatus"
+//         type="checkbox"
```
volodymyr-soltys97, line 60: Remove all comments

src/components/Footer.tsx
```diff
@@ -0,0 +1,69 @@
...
+          </a>
+
+          <a
+            href="#/active"
+            className={classNames('filter__link ', {
+              selected: activeFilter === 'active',
+            })}
+            data-cy="FilterLinkActive"
+            onClick={() => onFilterClick('active')}
+          >
```
volodymyr-soltys97: Use `Object.values(your created enum)` and render these options with `map()` method

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

src/types/index.ts
Manuilenkoart: Why is enum better than union type in this case?

src/components/TodoList.tsx
Manuilenkoart: I need it as a reference to the state of all components, and I delete it at the final stage when everything is implemented

src/components/Footer.tsx
Manuilenkoart: Add Navigation component and refactor Footer to use children prop

src/types/index.ts
etojeDenys: cause if you change some of this field, you will need to change it everywhere, with enum it's different(you need to change it only in one place)

etojeDenys (approved):

src/components/Footer.tsx
Manuilenkoart: @etojeDenys What do you think about this [PR](https://github.com/mate-academy/react_todo-app-loading-todos/commit/23e80418e8df42fc8e9c7ca058fb1a1cb0ea69f4) when components take initial data and return modified data?

PR #1729, opened 2025-02-17
Develop

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,203 +1,25 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
-import { UserWarning } from './UserWarning';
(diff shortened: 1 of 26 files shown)


Dominik-Dudek-code:
Basically, there are a lot of components and files here, as well as many methods that are not yet working because I copied the entire file from another task. Therefore, please review only those parts that relate to this task, which generally include part of 
TodoContext, whole NotificationContext, Loader, Notification, and the mostly completed but untested TodoInput.

Dominik-Dudek-code:
I would just like to add that the methods for renaming and deleting only work on the locally stored state, because as I mentioned, they are from another task. I will implement them properly in the next part of the current task.

danon321 (changes requested):
Link do demo nie działa :(

danon321 (approved):

PR #1731, opened 2025-02-18
Develop

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_todo-app-loading-todos/)


src/components/Filter/Filter.tsx
```diff
@@ -0,0 +1,100 @@
...
+          selected: isSelected === FilterOptions.Completed,
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={() =>
+          handleFilterSelect(
+            todos,
+            FilterOptions.Completed,
+            onFilterSelect,
+            setIsSelected,
+          )
```
naziamoff, line 93: this piece of code is often repeated, however only one prop differs. 
Also, 2 of the props are taken from the parent component as a prop, it means that it can be simplified. Do we need to expect todos and onFilterSelect in the function?

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+  todos: Todo[];
+};
+
+export const Header: FC<Props> = ({ todos }) => {
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={classNames('todoapp__toggle-all', {
+          active: verifyActiveTodos(todos),
```
naziamoff, line 19: I would say we don't need a function for that since it is only used once

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const verifyActiveTodos = (todos: Todo[]) => {
+  return todos.every(todo => todo.completed);
+};
+
+type Props = {
+  todos: Todo[];
+};
+
+export const Header: FC<Props> = ({ todos }) => {
```
naziamoff, line 13: we can return right away since there's no component logic inside

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Todo } from '../../types/Todo';
+import { TodoItem } from '../TodoItem/TodoItem';
+
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
```
naziamoff, line 11: return statement is not necessary here, we can return right away by wrapping component result in the round braces instead of the curly ones

naziamoff (approved):

iBrianWarner (approved):
Good job!

PR #1732, opened 2025-02-18
add solution

nineuito:
- [DEMO LINK](https://nineuito.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Hi, It is bad practice to write all the code in one component, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

anastasiiavorobiova (approved):

PR #1742, opened 2025-02-28
add react todo app loading todos

ValentynaD:
https://ValentynaD.github.io/react_todo-app-loading-todos/


etojeDenys (approved):
image
looks broken

don't forget to fix that and split your into some components(like Header,Footer,TodoList)

PR #1751, opened 2025-03-05
add solution

ArtemHryb:
[DEMO LINK](https://ArtemHryb.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,18 +1,70 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+
+type SelectStatus = 'All' | 'Active' | 'Completed';
```
volodymyr-soltys97: Create enum for this type and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to separate the logic into different components. For example `Header, Footer, TodoItem, TodoList`

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+        />
+        <span className="visually-hidden" />
+      </label>
+
+      <span data-cy="TodoTitle" className="todo__title">
+        {todo.title}
```
volodymyr-soltys97, line 30: You can use destructuring for `todo`

volodymyr-soltys97 (approved):
Great! 🔥

PR #1759, opened 2025-03-10
add task solution

anastasia-maidanyuk:
[DEMO LINK](https://anastasia-maidanyuk.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

vadiimvooo (approved):
Well done

PR #1760, opened 2025-03-10
add task solution2

YehorHeiko:
[DEMO LINK](https://YehorHeiko.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoItem/TodoItem.tsx
+++ b/src/components/TodoItem/TodoItem.tsx
@@ -1,0 +1,42 @@
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  visibleGoods: Todo;
+};
+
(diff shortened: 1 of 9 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,42 @@
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  visibleGoods: Todo;
```
etojeDenys, line 4:
```suggestion
  todo: Todo;
```

etojeDenys (approved):

PR #1761, opened 2025-03-10
task sol

VolodimirKurhan:
https://VolodimirKurhan.github.io/react_todo-app-loading-todos/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

maxim2310 (changes requested):
Good work, but you need to pass all tests firstly

src/App.tsx
```diff
@@ -1,203 +1,162 @@
...
+import { Todo } from './types/Todo';
+import { getTodos, addTodo, deleteTodo, updateTodo } from './api/todos';
+import TodoList from './components/TodoList';
+import Footer from './components/Footer';
+import Header from './components/Header';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [newTodo, setNewTodo] = useState('');
+  const [loading, setLoading] = useState(false);
```
anastasiiavorobiova, line 16:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
It's better to name boolean states starting from the verb `is` / `has` etc.

src/App.tsx
```diff
@@ -1,203 +1,162 @@
...
+  const [error, setError] = useState<string | null>(null);
+  const [activeFilter, setActiveFilter] = useState('All');
```
anastasiiavorobiova, line 18: Consider using an enum here

src/App.tsx
```diff
@@ -1,203 +1,162 @@
...
+      setError('Unable to update a todo');
+      setTodos(
+        todos.map(t => (t.id === todo.id ? { ...t, isLoading: false } : t)),
+      );
+    }
+  };
+
+  const filteredTodos = () => {
+    switch (activeFilter) {
+      case 'Active':
```
anastasiiavorobiova, line 100: Consider using an enum here

anastasiiavorobiova (approved):
Great work!
Added a few comments to fix

PR #1765, opened 2025-03-12
Develop

Beliyandr:
https://beliyandr.github.io/react_todo-app-loading-todos/


src/App.tsx
```diff
@@ -1,101 +1,226 @@
...
+  const [loading, setLoading] = useState(false);
+  const [errorMsg, setErrorMsg] = useState<string>('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todo, setTodo] = useState<string>('');
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  const [activeFilter, setActiveFilter] = useState<filterName>('All');
+
   if (!USER_ID) {
```
etojeDenys: move this condition after all hooks
image

etojeDenys (changes requested):

src/utils/fetchClient.ts
```diff
@@ -1,6 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 const BASE_URL = 'https://mate.academy/students-api';
 
+// Your userId is  1455
+// const USER_ID = 'https://mate.academy/students-api/todos?userId=1455';
```
volodymyr-soltys97, line 5: Remove comments

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Remove this file 
image

Beliyandr:
I fixed
Will I do decompositions in next task?

etojeDenys (approved):

PR #1767, opened 2025-03-12
solution

MarushchakM:


src/App.tsx
```diff
@@ -1,204 +1,89 @@
...
-              className="filter__link selected"
-              data-cy="FilterLinkAll"
-            >
-              All
-            </a>
+  useEffect(() => {
+    if (allTodos) {
+      setTodos(filterTodo(allTodos, filterData));
+    }
+  }, [filterData, allTodos]);
```
etojeDenys: it's just can be the var
```
```

src/App.tsx
```diff
@@ -1,204 +1,89 @@
...
-              <div className="loader" />
-            </div>
-          </div>
-        </section>
+export const App: React.FC = () => {
+  const [errorMassage, setErrorMassage] = useState('');
+  const [allTodos, setAllTodos] = useState<Todo[]>();
+  const [filterData, setFilterData] = useState('All');
+  const [todos, setTodos] = useState<Todo[]>();
+  const [todosCounter, setTodosCounter] = useState(0);
```
etojeDenys, line 32: this state looks redundant to me

etojeDenys (changes requested):

src/App.tsx
MarushchakM: I deleted line 32. Can you recommend what else I can improve?

vadiimvooo (approved):
Well done

PR #1770, opened 2025-03-14
sollution

serbinm:
https://serbinm.github.io/react_todo-app-loading-todos/


src/App.tsx
```diff
@@ -1,13 +1,103 @@
...
+type Filter = 'all' | 'completed' | 'active';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [originalTodoList, setOriginalTodoList] = useState<Todo[]>([]);
```
etojeDenys, line 20: you don't need this state

src/App.tsx
```diff
@@ -1,13 +1,103 @@
...
+      case 'active':
+        setTodoList(originalTodoList.filter(task => !task.completed));
+        break;
+      case 'completed':
+        setTodoList(originalTodoList.filter(task => task.completed));
+        break;
+      default:
+        setTodoList(originalTodoList);
+    }
+  };
```
etojeDenys, line 100: it can just be the var
```
```

etojeDenys, line 100: and this function will just change the status

etojeDenys (changes requested):

PR #1776, opened 2025-03-19
sol

dvdmsk:
- [DEMO LINK](https://dvdmsk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -35,13 +35,7 @@ export const App: React.FC = () => {
   const filteredTodos = getFilteredTodos(todos, filter);
 
   function getCountActive(): number {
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,204 +1,82 @@
...
+  const filteredTodos = getFilteredTodos(todos, filter);
+
+  function getCountActive(): number {
+    return todos.reduce((acc, el) => {
+      if (!el.completed) {
+        return acc + 1;
+      }
+
+      return acc;
+    }, 0);
```
etojeDenys: you can just filter by completed === true and get the length of that array

etojeDenys (approved):

PR #1785, opened 2025-03-24
add task solution

Sonya-Miss:
[DEMO LINK](https://Sonya-Miss.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,39 +2,40 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID_G } from './api/todos';
 import { Todo } from './types/Todo';
-import { client } from './utils/fetchClient';
+// import { client } from './utils/fetchClient';
+
+enum FilterType {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
 
 export const App: React.FC = () => {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,18 +1,96 @@
...
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { client } from './utils/fetchClient';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState<boolean>(false);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: it's better to create a separate enum for those values and reuse everywhere

src/App.tsx
```diff
@@ -1,18 +1,96 @@
...
+  const loadTodos = () => {
+    if (!USER_ID) {
+      return;
+    }
+
+    setLoading(true);
+    setError('');
+
+    client
+      .get<Todo[]>(`todos?userId=${USER_ID}`)
```
etojeDenys: you can just call the getTodos

etojeDenys (changes requested):

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one place, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

vadiimvooo (approved):
Well done

PR #1797, opened 2025-03-27
add first part of the solution

mmaryblackk:
[DEMO LINK](https://mmaryblackk.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,204 +1,83 @@
+/* eslint-disable @typescript-eslint/no-shadow */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useEffect, useState } from 'react';
+// import { UserWarning } from './UserWarning';
+import { getTodos } from './api/todos';
+import { TodoForm } from './components/TodoForm/TodoForm';
+import { TodoList } from './components/TodoList/TodoList';
+import { TodoFilter } from './components/TodoFilter/TodoFilter';
+import { Error } from './components/Error/Error';
+import { Todo } from './types/Todo';
+import { FilterOption } from './types/Filter';
+import { ErrorType } from './types/Error';
+import { Loader } from './components/Loader/Loader';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  // if (!USER_ID) {
+  //   return <UserWarning />;
+  // }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<ErrorType>(
+    ErrorType.noError,
+  );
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,204 +1,83 @@
...
+  const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<ErrorType>(
+    ErrorType.noError,
+  );
+  const [filterField, setFilterField] = useState(FilterOption.all);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
```
reb0rned, line 30: You can make less overloaded useEffect function, just implement all the logic outside, and then just invoke it in useEffect.
something like

```
useEffect(() => {
  getData()
}, [])
```

src/App.tsx
```diff
@@ -1,204 +1,83 @@
...
+      .catch(() => {
+        setErrorMessage(ErrorType.loading);
+        setTimeout(() => {
```
reb0rned: It is recommended to clear the timeout to avoid potential memory leaks or incorrect behavior if the component gets unmounted before the timeout completes...

reb0rned (changes requested):
👍  good job, small notes below, pls fix them

P.S. never gonna give you up, never gonna let you down, never gonna run around :)

reb0rned (approved):

src/App.tsx
alexander-ignatow: IMO - not critical in this case
Even more - inlined function has some advantages here

src/App.tsx
```diff
@@ -1,204 +1,89 @@
...
+  useEffect(() => {
+    loadData();
+  }, []);
+
+  const filteredTodos = todos.filter(
+    todo =>
+      filterField === FilterOption.all ||
+      (filterField === FilterOption.completed && todo.completed) ||
+      (filterField === FilterOption.active && !todo.completed),
+  );
```
alexander-ignatow, line 51: With the small number of todos it is probably ok, but for the  larger lists - I don't like the filter called on every render regardless of the `filterField` state
For instance - you can skip filtering for filterField === FilterOption.all

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: filterField === FilterOption.completed,
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onFilter(FilterOption.completed)}
+        >
+          Completed
+        </a>
```
alexander-ignatow, line 54: It is not critical - but you have 3 nearly identical <a> elements

alexander-ignatow (approved):
Good job - there are few minor issues but nothing is really critical!

src/components/TodoFilter/TodoFilter.tsx
mmaryblackk: I thought about mapping, tbh, but it is critical to have different data-cy :)

src/App.tsx
mmaryblackk: Already fixed in the next parts :)

PR #1820, opened 2025-04-03
Develop

kondratiukD:
[DEMO LINK](https://kondratiukD.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Hi, the demo link looks different as on the working example.
Need to remove it
image

Need to hide footer when we don't have todos 
image

The first letter must be in uppercase

etojeDenys (approved):

PR #1821, opened 2025-04-04
add task solution

nataliaso23:
[DEMO LINK] (https://nataliaso23.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,18 +1,66 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
```
raulriato, line 2: Disabling eslint is not a good practice

raulriato (approved):
As your code works with no major problems, I'll approve it. 

But be careful with disabling eslint.

PR #1837, opened 2025-04-16
solution

pryge:
[DEMO LINK](https://pryge.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,18 +37,20 @@ export const App: React.FC = () => {
 
       return () => clearTimeout(timer);
     }
+
+    return () => {};
   }, [errorMessage]);
 
   const filteredTodos = todos.filter(todo => {
-    if (filter === 'Active') {
-      return !todo.completed;
+    switch (filter) {
+      case 'Active':
+        return !todo.completed;
+      case 'Completed':
+        return todo.completed;
+      case 'All':
+      default:
+        return true;
     }
-
-    if (filter === 'Completed') {
-      return todo.completed;
-    }
-
-    return true;
   });
 
   if (!USER_ID) {
@@ -66,7 +68,6 @@ export const App: React.FC = () => {
             type="button"
             className="todoapp__toggle-all active"
             data-cy="ToggleAllButton"
-            disabled
           />
 
           {/* Add a todo on form submit */}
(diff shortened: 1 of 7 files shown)


volodymyr-soltys97 (changes requested):
Hi, here is no demo link. Also, you need to provide the demo link to the PR description 
image

It is also bad practice to write all the code in one place, you need to separate the logic into different components, for example `Header, Footer, TodoItem, TodoList`

volodymyr-soltys97 (changes requested):
Hi, here is no demo link. You need to provide the demo link to the PR description 
image

src/App.tsx
```diff
@@ -1,203 +1,86 @@
...
+import { TodoFooter } from './components/Footer';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
-  }
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filterBy, setFilterBy] = useState<Filter>('All');
```
volodymyr-soltys97: Сreate a enum for 'All', 'Active', 'Completed' and use it everywhere

src/App.tsx
```diff
@@ -1,203 +1,86 @@
...
+  );
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterBy) {
+        case 'Active':
+          return !todo.completed;
+
+        case 'Completed':
+          return todo.completed;
```
volodymyr-soltys97: Use enum here

src/components/Footer.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import cn from 'classnames';
+
+import { Filter } from '../types/Filter';
+
+const FILTER_LINKS: Filter[] = ['All', 'Active', 'Completed'];
```
volodymyr-soltys97: Use enum and `Object.values(your created enum)` for creating array with these values

src/types/Filter.ts
```diff
@@ -0,0 +1 @@
+export type Filter = 'All' | 'Active' | 'Completed';
```
volodymyr-soltys97: Enum

volodymyr-soltys97 (changes requested):
Almost done!

etojeDenys (approved):

PR #1849, opened 2025-04-29
Task solution

MateuszCieplak:
[DEMO LINK](https://MateuszCieplak.github.io/react_todo-app-loading-todos/)


Zibi95 (changes requested):
Hey. Demo link doesnt work and fix lint problems

Zibi95 (changes requested):
You went a little bit overboard with this one. You only had to implement the loading of todos. There are some things that I would change, but they are not related to this part. I can give you a hint: it's about the `Header` and input logic in `App`. Fix the lint issues so you will pass the CI tests, and we are good to go to the next step.

danon321 (approved):

PR #1872, opened 2025-05-23
Solution

JPodgrodzki:
[DEMO LINK](https://JPodgrodzki.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,49 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,10 +1,49 @@
...
+  const filteredTodos = todos.filter(todo => {
+    switch (filter) {
+      case 'all':
+        return true;
+      case 'active':
+        return !todo.completed;
+      case 'completed':
+        return todo.completed;
+    }
+  });
```
natalia-klonowska: It's better to avoid hardcoding filter values like 'all', 'active', and 'completed' directly in useState and switch cases. Instead consider defining them as union type or enum. This makes the code easier to maintain and less error-prone especially when these values are used in multiple places or need to be updated.

src/App.tsx
```diff
@@ -1,10 +1,49 @@
...
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
```
natalia-klonowska: Instead of passing arbitrary values into setError define fixed set of possible error types to ensure consistency and maintainability

natalia-klonowska (changes requested):
Adjust the visual appearance during loading to better match design
image

Zibi95 (approved):

PR #1880, opened 2025-06-03
solution

VladaProskura:
- [DEMO LINK](https://vladaproskura.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,7 +14,6 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
   const [filter, setFilter] = useState<Filter>('All');
-  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
   const activeTodosQuantity = todos.filter(todo => !todo.completed).length;
 
   const clearErrorMessage = () => {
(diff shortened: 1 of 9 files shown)


etojeDenys (changes requested):
looks good overall, but don't forget to provide the demo link in the PR description and then re-request the review

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filter, setFilter] = useState<Filter>('All');
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
etojeDenys: oh, didn't notice that, this state is redundant

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+    switch (filter) {
+      case 'All':
+        break;
+      case 'Active':
+        newTodos = todos.filter(todo => !todo.completed);
+        break;
+      case 'Completed':
+        newTodos = todos.filter(todo => todo.completed);
+        break;
+    }
```
etojeDenys: it can be just some var
```
```
and this function will return filteredTodos to you

etojeDenys (changes requested):
small update

src/App.tsx
VladaProskura: I've deleted:  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);

src/App.tsx
VladaProskura: function filterTodos(): Todo[] {
    switch (filter) {
      case 'All':
        return todos;
      case 'Active':
      case 'Completed':
        return todos.filter(todo => todo.completed);
    }
  }

  const filteredTodos = filterTodos();

VitaliyBondarenko1982 (approved):
Looks good.

PR #1882, opened 2025-06-04
Develop

xxxKoRoBoKxxx:
[DEMO LINK](https://xxxkorobokxxx.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,7 @@ import React, { useEffect, useState } from 'react';
 import { getTodos, USER_ID } from './api/todos';
 import { wait } from './utils/fetchClient';
 import { filteringTodos } from './utils/queueTodos';
+import { countItemsLeft } from './utils/countItemsLeft';
 import { Todo } from './types/Todo';
 import { TodoFilter } from './types/TodoFilter';
 
@@ -34,7 +35,8 @@ export const App: React.FC = () => {
     return <UserWarning />;
   }
 
-  const queuedTodos = filteringTodos(allTodos, filter);
+  const queuedTodos: Todo[] = filteringTodos(allTodos, filter);
+  const itemsLeft: number = countItemsLeft(allTodos);
 
   return (
     <div className="todoapp">
@@ -46,7 +48,7 @@ export const App: React.FC = () => {
         <TodoList todos={queuedTodos} />
 
         {allTodos.length > 0 && (
-          <Footer filter={filter} setFilter={setFilter} />
+          <Footer filter={filter} setFilter={setFilter} itemsLeft={itemsLeft} />
         )}
       </div>
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,7 +36,7 @@ export const App: React.FC = () => {
         setError('noTodos');
         wait(3000).then(() => setError(''));
       });
-  }, []);
+  }, [countCompletedTodos]);
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(updateTodosFromServer, []);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,17 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useCallback, useEffect, useRef, useState } from 'react';
-import { UserWarning } from './UserWarning';
-import { getTodos, USER_ID } from './api/todos';
-import { Todo } from './types/Todo';
-import { Todos } from './components/Todos/Todos';
-import classNames from 'classnames';
-import { wait } from './utils/fetchClient';
 
-type TodosFilter = 'All' | 'Active' | 'Completed';
+import { getTodos, USER_ID } from './api/todos';
+import { wait } from './utils/fetchClient';
+import { Todo } from './types/Todo';
+import { TodosFilter } from './types/TodosFilter';
+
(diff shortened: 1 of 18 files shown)


src/App.tsx
```diff
@@ -1,10 +1,61 @@
...
+    if (todosFilter === 'Active') {
+      newTodos = allTodos.filter(todo => !todo.completed);
+    } else if (todosFilter === 'Completed') {
+      newTodos = allTodos.filter(todo => todo.completed);
+    } else {
+      newTodos = allTodos;
+    }
+
+    setQueuedTodos(newTodos);
+  }, [todosFilter]);
```
anastasiiavorobiova:
```suggestion
  const filteredTodos = allTodos.filter(todo => { ...
```
There is no need to use additional state and useeffect here

src/App.tsx
```diff
@@ -34,170 +85,211 @@ export const App: React.FC = () => {
...
+              {/* This is a completed todo */}
+              <div data-cy="Todo" className="todo completed">
+                <label className="todo__status-label">
+                  <input
+                    data-cy="TodoStatus"
+                    type="checkbox"
+                    className="todo__status"
+                    checked
+                  />
+                </label>
```
anastasiiavorobiova:
```suggestion

```

anastasiiavorobiova: If you don't need the part of the layout you can delete it

src/App.tsx
```diff
@@ -34,170 +85,211 @@ export const App: React.FC = () => {
...
-          <nav className="filter" data-cy="Filter">
-            <a
-              href="#/"
-              className="filter__link selected"
-              data-cy="FilterLinkAll"
-            >
-              All
-            </a>
+        {todosFromServer.current.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
anastasiiavorobiova: Consider moving this to the footer component

src/App.tsx
```diff
@@ -34,170 +85,211 @@ export const App: React.FC = () => {
...
+              Clear completed
+            </button>
+          </footer>
+        )}
       </div>
 
       {/* DON'T use conditional rendering to hide the notification */}
       {/* Add the 'hidden' class to hide the message smoothly */}
       <div
         data-cy="ErrorNotification"
```
anastasiiavorobiova: Consider moving this to the error component

src/App.tsx
```diff
@@ -34,170 +85,211 @@ export const App: React.FC = () => {
         </header>
```
anastasiiavorobiova: Consider moving this to the header component

anastasiiavorobiova (changes requested):
Great work!
Added comments to fix

src/App.tsx
xxxKoRoBoKxxx: Hello,
I can change useEffect to some callback, no problem, but design of these 3 buttons relies on applied filter, how can i change statuses of each button without filter state? In this way i can just change filter state and design of each button will be changed automatically, without calling any callbacks.

Thank you, @anastasiiavorobiova

src/App.tsx
xxxKoRoBoKxxx: I kept it for reference, in final version it will be removed

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { Header } from './components/Header';
+import { Todos } from './components/Todos/Todos';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [queuedTodos, setQueuedTodos] = useState<Todo[]>([]);
```
etojeDenys:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+  const [todosFilter, setTodosFilter] = useState<TodosFilter>('All');
+
+  const [error, setError] = useState('');
+
+  const todosFromServer = useRef<Todo[]>([]);
+  const [completedTodosCount, setCompletedTodosCount] = useState<number>(0);
```
etojeDenys: looks redundant to me

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+    getTodos()
+      .then(todosRecieved => {
+        todosFromServer.current = [...todosRecieved];
+        setQueuedTodos([...todosRecieved]);
+        countCompletedTodos();
+      })
+      .catch(() => {
+        setError('Unable to load todos');
+        wait(3000).then(() => setError(''));
+      });
```
etojeDenys:
```suggestion
    getTodos()
      .then(todosRecieved => {
        setTodos([...todosRecieved]);
      })
      .catch(() => {
        setError('Unable to load todos');
        wait(3000).then(() => setError(''));
      });
```
just move this logic to useEffect

src/App.tsx
etojeDenys: I didn't get it, you don't need to have the additional state for filteredTodos, just save it to some var

if you have some questions about it, feel free to ask them in the chat, you can always ping me there
image

src/App.tsx
etojeDenys: it's also just can be some var

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Looks good.

PR #1887, opened 2025-06-07
Develop

keex1k:
[DEMO LINK](https://keex1k.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,107 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { addTodos, getTodos, USER_ID } from './api/todos';
+import { UserTodosList } from './UserTodosList';
+import { Todo } from './types/Todo';
+import { ErrorMessage } from './ErrorMessage';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [newTodo, setNewTodo] = useState<string>('');
+  const [checkedAll, setCheckedAll] = useState<boolean>(false);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<string>('all');
+
+  const clearError = () => {
+    setError('');
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(data => {
+        setTodos(data ?? null);
+      })
+      .catch(() => setError('load'));
+  }, []);
+
+  useEffect(() => {
+    if (error) {
+      const timer = setTimeout(() => {
+        clearError();
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setNewTodo(newTodo);
+
+    const newTodoObject: Todo = {
+      userId: USER_ID,
+      title: newTodo,
+      completed: false,
+      id: todos !== null ? todos.length + 1 : 1,
+    };
+
+    if (newTodo.trim().length === 0) {
+      setError('empty title');
+    } else {
+      if (todos) {
+        setTodos([...todos, newTodoObject]);
+      } else {
+        setTodos([newTodoObject]);
+      }
+
+      addTodos(newTodoObject).catch(() => setError('add'));
+    }
+
+    setNewTodo('');
+  };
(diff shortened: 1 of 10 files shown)


danon321:
Zapodaj jeszcze demo link

Replace <your_account> with your Github username in the [DEMO LINK](https://%3Cyour_account%3E.github.io/react_todo-app-loading-todos/) and add it to the PR description.

danon321 (changes requested):
Zapodaj link do demo

keex1k:
> Zapodaj link do demo

zapodane :)

src/App.tsx
```diff
@@ -1,204 +1,238 @@
...
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(data => {
+        setTodos(data ?? null);
+        setShouldFetch(false);
+      })
+      .catch(() => setError('load'));
+  }, [shouldFetch]);
```
zibi-dev: on each action it is fetching todos two times.

src/App.tsx
```diff
@@ -1,204 +1,238 @@
...
+    Promise.all(
+      todos.map(todo =>
+        updateTodo(todo.id, { completed: newCompletedStatus }).catch(() =>
+          setError('update'),
+        ),
+      ),
+    );
+
+    setShouldFetch(true);
+  };
```
zibi-dev: These actions are not relevant for this part of task

src/App.tsx
```diff
@@ -1,204 +1,238 @@
...
+  useEffect(() => {
+    if (error) {
+      const timer = setTimeout(() => {
+        clearError();
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
```
zibi-dev: You can remove this line. Its doing it anyway

src/App.tsx
```diff
@@ -1,204 +1,238 @@
...
+          <form onSubmit={handleSubmit}>
             <input
               data-cy="NewTodoField"
               type="text"
               className="todoapp__new-todo"
               placeholder="What needs to be done?"
+              value={newTodo}
+              onChange={e => setNewTodo(e.target.value)}
             />
           </form>
```
zibi-dev, line 162: This change of state is rerendering whole component on each key stroke

src/App.tsx
```diff
@@ -1,204 +1,238 @@
...
+                href="#/completed"
+                className={classNames({
+                  filter__link: true,
+                  selected: filter === 'completed',
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
```
zibi-dev: These links look similiar maybe it can be separated component

src/additionalTasks.txt
zibi-dev, line 1: ?

Zibi95 (changes requested):

src/App.tsx
keex1k: but in console, in network I see just once:/ could You tell me how to fix it, because I haven't got any idea

src/App.tsx
keex1k: yeah, I know have I remove this code? or can I leave it

src/App.tsx
keex1k: You means line 45?

src/App.tsx
keex1k: so How I should do that? I should set new state only in submit function? and what I should put into onChange prop?

src/additionalTasks.txt
keex1k: I creaeted it for next part of task;) I didn't want to lose html structure

src/App.tsx
natalia-klonowska: you don't need shouldFetch state and to re-fetch todos. After adding, updating or deleting you can just update the local todos state. This avoids unnecessary network requests

src/App.tsx
natalia-klonowska: yes, you don't need `return undefined;` because function returns undefined by default if you don't return anything

src/App.tsx
natalia-klonowska: It’s better to remove it now as fixing it without tests from the next part of the project would be difficult

natalia-klonowska (changes requested):

src/App.tsx
zibi-dev: Move it into other component with state. Keeping such state in root of components is a killer for performance. You woul have to use memoizations techniques like useMemo or useCallback, but firstly you should focus on component composition before using memoization functions.

zibi-dev: You can either create an `Input` component. Alternatively, if you're not checking the input value on every keystroke, you can remove the input's state and instead use a ref or FormData within the handleSubmit function to retrieve the data.
https://developer.mozilla.org/en-US/docs/Web/API/FormData

Zibi95 (approved):
I have left comments regarding that state for input. Fix it in other part of task. This one is okay

PR #1893, opened 2025-06-09
solution

Vlad-2608:
https://Vlad-2608.github.io/react_todo-app-loading-todos/


etojeDenys (changes requested):
you need to pass all tests, feel free to ask for some help with that in the chat and then re-reques the review, you can always ping me there

anastasiiavorobiova (approved):
Awesome!

PR #1907, opened 2025-06-11
Solution

TetianaHodis1308:
[DEMO LINK](https://TetianaHodis1308.github.io/react_todo-app-loading-todos/)


MykhailoSavchenko (approved):
Все наче працює, більше сказати не можу)

src/App.tsx
```diff
@@ -1,28 +1,86 @@
...
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import cn from 'classnames';
+import { Filters } from './types/Filtres';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isErrorHidden, setIsErrorHidden] = useState(true);
```
ArturErmolenko:
```suggestion
```
why we need this state?

src/App.tsx
```diff
@@ -1,28 +1,86 @@
...
+
+  useEffect(() => {
+    setIsErrorHidden(true);
+
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setSelectedTodos(todosFromServer);
+      })
+      .catch(() => setErrorMessage('Unable to load todos'))
```
ArturErmolenko: use enum for error messages instead of magic string

src/App.tsx
```diff
@@ -1,28 +1,86 @@
...
+      .finally(() => {
+        setLoading(false);
+      });
+  }, []);
+
+  useEffect(() => {
+    if (errorMessage) {
+      setIsErrorHidden(false);
```
ArturErmolenko: we should work with errorMessage state
let's get rid of second state

src/App.tsx
```diff
@@ -1,28 +1,86 @@
...
+
+      case Filters.Active:
+        setSelectedTodos(todos.filter(todo => !todo.completed));
+        break;
+
+      case Filters.All:
+      default:
+        setSelectedTodos(todos);
+        break;
+    }
```
ArturErmolenko: no we shouldn't go this way
we need smth like 
const preparetTodos = getPreparedTodos(todos,filter) ...

src/App.tsx
```diff
@@ -1,28 +1,86 @@
...
+import cn from 'classnames';
+import { Filters } from './types/Filtres';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isErrorHidden, setIsErrorHidden] = useState(true);
+  const [filter, setFilter] = useState<Filters>(Filters.All);
+  const [selectedTodos, setSelectedTodos] = useState<Todo[]>([]);
```
ArturErmolenko:
```suggestion
```
one more useless state, why we need it? can we work with todos state

src/types/Filtres.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filters {
```
ArturErmolenko: what filters? users or movies?) we need better naming

ArturErmolenko (changes requested):
Nice start
But we should get rid of extra states

ArturErmolenko (approved):
Fix it in next part

PR #1924, opened 2025-06-14
solution

deadslowright:
[DEMO LINK] https://deadslowright.github.io/react_todo-app-loading-todos/


anastasiia-zu (approved):
good job, good luck with the second part!

src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
+      return !todo.completed;
+    });
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodo, setLoadingTodo] = useState(true);
```
ArturErmolenko, line 34:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
Boolean state should start with is has and start value should be false

src/App.tsx
```diff
@@ -1,10 +1,57 @@
...
+    filteredTodos = filteredTodos.filter(todo => {
+      if (filter.status === FilterOptions.Completed) {
+        return todo.completed;
+      }
+
+      return !todo.completed;
+    });
+  }
+
+  return filteredTodos;
```
ArturErmolenko, line 29: use switch case
try to get rid of let
you can pass switch case in filter callback

ArturErmolenko (approved):
Good start 👍 
Please fix comments in next part

PR #1945, opened 2025-06-22
add task solution

rkoltunov:
[DEMO LINK](https://rkoltunov.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to remove it 
image

src/App.tsx
```diff
@@ -1,203 +1,88 @@
...
-          <div data-cy="Todo" className="todo">
-            <label className="todo__status-label">
-              <input
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'active') {
```
TarasHoliuk, line 19: You should use an enum value in comparing

TarasHoliuk (approved):
Well done! Approved ✅

PR #1960, opened 2025-07-09
result1

Diana2369:
[DEMO LINK](https://Diana2369.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,51 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -15,189 +56,117 @@ export const App: React.FC = () => {
...
+              className={`todo ${todo.completed ? 'completed' : ''}`}
+              key={todo.id}
+            >
+              <label className="todo__status-label">
+                <input
+                  data-cy="TodoStatus"
+                  type="checkbox"
+                  className="todo__status"
+                  checked={todo.completed}
+                  readOnly
```
volodymyr-soltys97: Remove it
```suggestion
```

volodymyr-soltys97 (changes requested):
Hi, It is bad practice to write all the code in one component, you need to split the logic into several different components, for example `Header, Footer, TodoItem, TodoList`

anastasiiavorobiova (approved):
Great work!

PR #1969, opened 2025-07-13
Add solution 

ghost:


vitalii-bondarenko-ua (changes requested):
First of all, provide, please, demo link to PR description.
Make sure demo link works correctrly before next re-request.

PR #1972, opened 2025-07-16
solution of 1st part

LinaDo1609:


anastasiiavorobiova (changes requested):
Great work!
- If the user goes to the completed filter and there are no completed todos, there is no way to change the filter back
Screenshot 2025-07-16 at 18 36 08

vitalii-bondarenko-ua (approved):
Looks good.

PR #1979, opened 2025-07-20
Solution

andreinik1:


src/App.tsx
```diff
@@ -15,189 +105,294 @@ export const App: React.FC = () => {
...
+            {/* Active link should have the 'selected' class */}
+            <nav className="filter" data-cy="Filter">
+              <a
+                href="#/"
+                className={`filter__link ${query === '' ? 'selected' : ''}`}
+                data-cy="FilterLinkAll"
+                onClick={() => setQuery('')}
+              >
+                All
+              </a>
```
vadiimvooo: Consider to use enum and map to render links

vadiimvooo (changes requested):
Good job, consider to split your code on some smaller components instead of having all one in App.tsx (Header, Footer, TodoList, TodoItem, Error etc)

src/App.tsx
```diff
@@ -1,10 +1,71 @@
...
+import { Error } from './components/Error';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([] as Todo[]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([] as Todo[]);
+  const [query, setQuery] = useState<string>('');
+  const [leftItems, setLeftItems] = useState<number>(0);
```
vitalii-bondarenko-ua, line 17: redundant state. should be enough todos state. consider to remove in next tasks.
```suggestion
```

vitalii-bondarenko-ua (approved):
Good job.

PR #1991, opened 2025-07-27
Solution

Whslv:
[DEMO LINK](https://Whslv.github.io/react_todo-app-loading-todos/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It is bad practice to write all the code in one component, you need to split the logic into several different components, for example `Header, Footer, TodoItem, TodoList`

vitalii-bondarenko-ua (approved):
Looks good

PR #2013, opened 2025-08-14
Add task

iteamfox:
[DEMO LINK](https://iteamfox.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,4 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import * as todoService from './api/todos';
 import { Header } from './components/Header';
@@ -10,31 +8,26 @@ import { TodoList } from './components/TodoList';
 import { ErrorNotification } from './components/ErrorNotification';
 
 function getFilteredTodos(
-  currentTodos: Todo[],
-  currentFilter: 'all' | 'active' | 'completed',
+  todos: Todo[],
+  filter: 'all' | 'active' | 'completed',
 ): Todo[] {
-  const filteredTodos = [...currentTodos];
-
-  switch (currentFilter) {
+  switch (filter) {
     case 'active':
-      return filteredTodos.filter(todo => !todo.completed);
+      return todos.filter(todo => !todo.completed);
     case 'completed':
-      return filteredTodos.filter(todo => todo.completed);
-    case 'all':
-      return filteredTodos;
+      return todos.filter(todo => todo.completed);
     default:
-      return [];
+      return todos;
   }
 }
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
   const [error, setError] = useState('');
   const [isLoading, setIsLoading] = useState(false);
-  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
-  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
   const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
-  const [inputValue, setInputValue] = useState('');
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
 
   useEffect(() => {
     if (!todoService.USER_ID) {
@@ -52,17 +45,11 @@ export const App: React.FC = () => {
   }, []);
 
   const handleAddTodo = async (title: string) => {
-    if (!title.trim()) {
-      setError('Title should not be empty');
-
-      return;
-    }
-
     setIsLoading(true);
 
     const newTempTodo: Todo = {
       id: 0,
-      title: title.trim(),
+      title,
       completed: false,
       userId: todoService.USER_ID,
     };
@@ -71,18 +58,9 @@ export const App: React.FC = () => {
     setUpdatingTodoIds(ids => [...ids, 0]);
 
     todoService
-      .addTodo({
-        userId: todoService.USER_ID,
-        title: title.trim(),
-        completed: false,
-      })
-      .then(newTodo => {
-        setTodos(currentTodos => [...currentTodos, newTodo]);
-        setInputValue('');
-      })
-      .catch(() => {
-        setError('Unable to add a todo');
-      })
+      .addTodo({ userId: todoService.USER_ID, title, completed: false })
+      .then(newTodo => setTodos(prev => [...prev, newTodo]))
+      .catch(() => setError('Unable to add a todo'))
       .finally(() => {
         setIsLoading(false);
         setTempTodo(null);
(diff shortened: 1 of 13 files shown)


danon321 (changes requested):
Nie ma linku do demo :c

src/components/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(todo => !todo.completed).length} items left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={`filter__link ${setCurrentFilter === 'all' ? 'selected' : ''}`}
```
zibi-dev: You have installed cool lib that allows you to do this with rly cool syntax. Its called `calssnames`.

src/components/Footer.tsx
```diff
@@ -0,0 +1,65 @@
...
+type Props = {
+  todos: Todo[];
+  onFilterChange: (filter: 'all' | 'active' | 'completed') => void;
+  setCurrentFilter: 'all' | 'active' | 'completed';
+  onClearCompleted: () => void;
+};
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  setCurrentFilter,
```
zibi-dev: This prop has name that suggests its setState function, but actually it is state itself. Its confusing

src/components/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+  disabled,
+  inputValue,
+  setInputValue,
+  inputRef,
+}) => {
+  useEffect(() => {
+    if (!disabled) {
+      inputRef.current?.focus();
+    }
+  }, [disabled, inputRef]);
```
zibi-dev: You don't have to put refs into deps array. 
https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

src/App.tsx
```diff
@@ -1,204 +1,200 @@
...
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [tempTodo, setTempTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [updatingTodoIds, setUpdatingTodoIds] = useState<number[]>([]);
+  const [inputValue, setInputValue] = useState('');
```
zibi-dev: Keeping this tate here causes the whole app to rerender on each keystroke

src/components/Loader.tsx
```diff
@@ -0,0 +1,23 @@
+import classNames from 'classnames';
+import React from 'react';
+
+type Props = {
+  todoId: number;
+  updatingTodoIds: number[];
+};
+
+export const Loader: React.FC<Props> = ({ todoId, updatingTodoIds }) => {
+  const isUpdating = updatingTodoIds.includes(todoId);
```
zibi-dev: I think that loader should'nt have this info about that. I believe that this type of component should be dumb and rendered conditionally in component that uses that.

src/components/TodoItem.tsx
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
+import React from 'react';
+import { Todo } from '../types/Todo';
+import { Loader } from './Loader';
+
+type Props = {
+  todo: Todo;
+  toggleTodo: (todo: Todo) => void;
+  isLoading: boolean;
+  updatingTodoIds: number[];
```
zibi-dev: This can be merged into one props `isLoading`

Zibi95 (changes requested):
Hey, good job! I’ve left some comments on what could be improved. This part should only focus on loading todos. I believe you already have the same code covered in other PRs that are part of the next steps. So i will request changes there also.

Zibi95 (changes requested):
I see that you moved input state into header and then again back into `App.tsx` What is the problem?

iteamfox:
> I see that you moved input state into header and then again back into `App.tsx` What is the problem?

The final version in snippet (with state in App.tsx) is correct.
It follows React’s principle of lifting state up: put state in the lowest common parent that needs to control it. Here:
Header only renders the input and forwards changes.
App owns the logic, clears it when needed, and keeps a single source of truth.

iteamfox:
> I see that you moved input state into header and then again back into `App.tsx` What is the problem?

There’s no bug right now — the “problem” was simply that keeping inputValue inside Header broke the data flow.

Zibi95:
> > I see that you moved input state into header and then again back into `App.tsx` What is the problem?
> 
> The final version in snippet (with state in App.tsx) is correct. It follows React’s principle of lifting state up: put state in the lowest common parent that needs to control it. Here: Header only renders the input and forwards changes. App owns the logic, clears it when needed, and keeps a single source of truth.

Nc try :smile: 

The “bad” part isn’t lifting state up — it’s lifting it too far up when it doesn’t need to be.

Keeping all state at the very top level of your app can quickly become a mess:


- Too much re-rendering: if App owns every piece of state, then any change can cause large parts of the tree to re-render unnecessarily.
- Poor separation of concerns: components that could be self-contained become “dumb shells” that rely on the parent for everything.
- Scalability issues: as the app grows, App.tsx turns into a “god component” that’s hard to maintain.

If the input is only relevant to Header → keep the state inside Header, and let it manage clearing itself.
If the input’s value or clearing logic is relevant to App or siblings (e.g. App needs to reset it after a search, or another component needs the value) → lift the state up, and yes, you’ll be passing down a “clear” function.

In this case input state is only relevant to the Header component bcs you are submitting and passing value of input into onAdd function and in handleSubmit you can reset the state. 

And there final question do you need to have controlled input in this App?

src/components/Header.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <header className="todoapp__header">
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${
+          todos.length > 0 && todos.every(todo => todo.completed)
+            ? 'active'
+            : ''
+        }`}
+      />
```
zibi-dev: Props should ideally be as simple and primitive as possible (booleans, strings, numbers), rather than passing down large objects or arrays, unless the child truly needs them.

- Easier to optimize:
If you pass a whole todos array into Header just so it can check “is there at least one todo?”, then every time the array reference changes, Header will re-render — even if the boolean result didn’t change.
- Better separation of concerns:
The parent (App) already knows the list of todos. It should also decide whether the child needs to be “active” or not, and just pass that as a boolean.
- Future-proofing:
If you later memoize Header with React.memo, it will only re-render when the primitive prop changes, not when a new array reference is created.

src/App.tsx
```diff
@@ -1,203 +1,171 @@
...
+  if (!todoService.USER_ID) {
     return <UserWarning />;
   }
 
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'active') {
+      return !todo.completed;
+    }
+
+    if (filter === 'completed') {
```
zibi-dev: It would good to have some object or enum for these values.

Zibi95 (changes requested):

src/components/Header.tsx
```diff
@@ -0,0 +1,40 @@
...
+    e.preventDefault();
+    onAdd(value);
+  };
+
+  return (
+    <header className="todoapp__header">
+      {/* this button should have `active` class only if all todos are completed */}
+      <button
+        type="button"
+        className={`todoapp__toggle-all ${todos.every(todo => todo.completed) && `active`}`}
```
natalia-klonowska: todos are passed here only to check if all are completed. It's better to compute this in parent component and pass a simple boolean prop instead because as Sylwester said before: "Props should ideally be as simple and primitive as possible (booleans, strings, numbers), rather than passing down large objects or arrays, unless the child truly needs them."

src/components/Loader.tsx
```diff
@@ -0,0 +1,15 @@
+type Props = {
+  isLoading?: boolean;
+};
+
+export const Loader: React.FC<Props> = ({ isLoading }) => {
+  return (
+    <div
+      data-cy="TodoLoader"
+      className={`modal overlay ${isLoading ? 'is-active' : ''}`}
```
natalia-klonowska: always use classnames when className depends on conditions. Fix all occurrences in your project

src/components/Footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const Footer: React.FC<Props> = ({
+  todos,
+  currentFilter,
+  onFilterChange,
+}) => {
+  return (
+    <footer className="todoapp__footer" data-cy="Footer">
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(todo => !todo.completed).length} items left`}
```
natalia-klonowska: same here with todos passed as props. you use it only to determine number and boolean values. you can calculate those in parent component and pass them down as primitive props instead

src/components/Footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+            {filterValue.charAt(0).toUpperCase() + filterValue.slice(1)}
+          </a>
+        ))}
+      </nav>
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={todos.every(todo => !todo.completed)}
```
natalia-klonowska: same here with todos passed as props. you use it only to determine number and boolean values. you can calculate those in parent component and pass them down as primitive props instead

src/App.tsx
```diff
@@ -1,204 +1,106 @@
...
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
+      })
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const handleAddTodo = async (title: string) => {
+    if (!title.trim()) {
+      setError('Title should not be empty');
```
natalia-klonowska: Instead of passing arbitrary values into setError define fixed set of possible error types to ensure consistency and maintainability

natalia-klonowska (changes requested):

Zibi95 (approved):
Good job!

PR #2032, opened 2025-08-26
add solution

derpedcatto:
[DEMO LINK](https://derpedcatto.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/components/TodoFooter.tsx
+++ b/src/components/TodoFooter.tsx
@@ -3,12 +3,14 @@ import { TodoFilter } from './TodoFilter';
 
 type Props = {
   todosLeft: number;
+  isHasCompletedTodos: boolean;
   filterStatus: FilterStatus;
   onFilterChange: (status: FilterStatus) => void;
 };
 
 export const TodoFooter: React.FC<Props> = ({
   todosLeft,
+  isHasCompletedTodos,
   filterStatus,
   onFilterChange,
 }) => {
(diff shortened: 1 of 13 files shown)


BlackUserSide (approved):

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,36 @@
+import { FilterStatus } from '../types/FilterStatus';
+import { TodoFilter } from './TodoFilter';
+
+type Props = {
+  todosLeft: number;
+  isHasCompletedTodos: boolean;
```
alexander-ignatow, line 6: Mind the grammar (here and in the other places)
```suggestion
  hasCompletedTodos: boolean;
```

src/components/TodoHeader.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames';
+import { NewTodoForm } from './NewTodoForm';
+
+type Props = {
+  isTodosListEmpty: boolean;
+  isAllTodosCompleted: boolean;
```
alexander-ignatow, line 7: mind the grammar

```suggestion
  areAllTodosCompleted: boolean;
```

alexander-ignatow (approved):
Lets proceed to the next stage, however you need to fix the following issues

What is the reason to have App and TodoApp?

As long as App holds all states, the TodoApp is clearly redundant, and should be removed.

So please either remove TodoApp, or move all the states there, to make App dumb.

PR #2039, opened 2025-08-27
Solution

romkas228:
[DEMO LINK](https://romkas228.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,14 +4,17 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
-
-type FilterType = 'all' | 'active' | 'completed';
+import { FilterType } from './types/FilterType';
+import { TodoList } from './components/TodoList';
+import { TodoFooter } from './components/TodoFooter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Loader } from './components/Loader';
+import { TodoHeader } from './components/TodoHeader/TodoHeader';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState<string | null>(null);
-  const [filter, setFilter] = useState<FilterType>('all');
+  const [filter, setFilter] = useState<FilterType>(FilterType.All);
   const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
+  useEffect(() => {
+    setError('');
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError('Unable to load todos'))
+      .finally(() => {
+        setIsLoading(false);
+        setTimeout(() => setError(null), 3000);
```
lvlashnin: Don't forget to clean out the timeout

lvlashnin (approved):
Your demo works as expected, but next time decompose it into components. APPROVED

src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
+      });
+  }, []);
+
+  const filteredTodos = useMemo(
+    () =>
+      todos.filter(todo => {
+        switch (filter) {
+          case 'active':
```
dothearts: Move all status to constants or enum

src/App.tsx
```diff
@@ -1,10 +1,47 @@
...
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError('Unable to load todos'))
+      .finally(() => {
+        setIsLoading(false);
+        setTimeout(() => setError(null), 3000);
+      });
+  }, []);
+
+  const filteredTodos = useMemo(
```
dothearts, line 44: You can split code of this component. 
Please try use clear function, and split code.

DanilWeda (changes requested):
Good job. 
Please check my comments. 
Many thanks!

src/App.tsx
alexander-ignatow: PLease clean up the timeout

Remember - you can return the cleanup function from useEffefct hook

```
     useEffect(() => {
      // your code
      
      const timeoutID = setTimeout(...);
      
      return () => {
         // cleanup goes here
         clearTimeout(timeoutID)
      }
     }, [...])
        setTimeout(() => setError(null), 3000);
```

see more at https://developer.mozilla.org/en-US/docs/Web/API/Window/clearTimeout

alexander-ignatow (changes requested):
Good, but one issue remained unresolved
And please make sure to transfer all fixes to the next version of the app

src/components/TodoFooter/TodoFooter.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <span className="todo-count" data-cy="TodosCounter">
+        {`${todos.filter(todo => !todo.completed).length} items left`}
+      </span>
+
+      {/* Active link should have the 'selected' class */}
+      <nav className="filter" data-cy="Filter">
+        <a
+          href="#/"
+          className={classNames('filter__link', {
+            selected: filter === 'all',
```
dothearts: Why you don't use ENUM in this case?
Please, change it in new iteration

DanilWeda (changes requested):
I don't see a clearTimeout. Also, you don't do code splitting and don't use ENUM in all cases. Please made this changes and request review.

igorkruz (approved):

PR #2058, opened 2025-09-02
solution

isobol-js:


bebelekele (changes requested):
I see, that you didn't pass tests because you have some issues in your code. You may start from creating a logic of showing Footer of Todos App.

An example:

`{todos.length > 0 && (
          <footer className="todoapp__footer" data-cy="Footer">...</footer>)`

bebelekele (approved):

alexander-ignatow (approved):
I wold separate header snd footer components - but lets look it on the next iterations

PR #2073, opened 2025-09-14
Solution

alyonashunevych:
[DEMO LINK](https://alyonashunevych.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,12 +4,16 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>();
   const [errorMessage, setErrorMessage] = useState('');
-  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [filter, setFilter] = useState<Filter>('all');
 
   useEffect(() => {
     getTodos()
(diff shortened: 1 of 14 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+  todos: Todo[];
+};
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  return (
+    <section className="todoapp__main" data-cy="TodoList">
+      {todos.map(todo => (
+        <div
+          data-cy="Todo"
```
vitalii-bondarenko-ua: make separate component for Todo (TodoItem)

src/App.tsx
```diff
@@ -1,10 +1,44 @@
...
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList/TodoList';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/ErrorNotification';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>();
+  const [errorMessage, setErrorMessage] = useState('');
```
vitalii-bondarenko-ua: create enum for all possible messages. provide enum as generic to useState. and use everywhere instead hardcoded strings.

src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <button data-cy="HideErrorButton" type="button" className="delete" />
+      {/* Unable to load todos
+        <br />
+        Title should not be empty
+        <br />
+        Unable to add a todo
+        <br />
+        Unable to delete a todo
+        <br />
+        Unable to update a todo */}
```
vitalii-bondarenko-ua: do not keep commented code

vitalii-bondarenko-ua (changes requested):
Good job.
Please check comments.

src/App.tsx
```diff
@@ -1,10 +1,43 @@
...
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+import { Filter } from './types/Filter';
+import { ErrorNotification } from './components/ErrorNotification';
+import { ErrorMessage } from './types/ErrorMessage';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>();
+  const [errorMessage, setErrorMessage] = useState<ErrorMessage>('');
+  const [filter, setFilter] = useState<Filter>('all');
```
volodymyr-soltys97, line 15: Create a enum for 'all', 'active', 'completed' and use it everywhere

volodymyr-soltys97 (changes requested):
Hi, here is no demo link. Also, you need to add the demo link to the PR description 
image

src/App.tsx
alyonashunevych: I decided to use a union type Filter instead of an enum, since it is simpler for cases like this. Do you think I should still switch to an enum?
`Filter.ts: export type Filter = 'all' | 'active' | 'completed';`

vadiimvooo (approved):
Well done

PR #2091, opened 2025-10-01
v1

tank-igorevich:
DEMO - https://tank-igorevich.github.io/react_todo-app-loading-todos/


src/App.tsx
vadiimvooo, line 1: Consider to split your app to some smaller components like TodoItem, TodoList, Footer, Header, Error etc

vadiimvooo (changes requested):
Well done, just one comment to fix. Please feel free to ask for help in fe_chat.

PR #2107, opened 2025-10-17
load todos by userId with error handling and filters

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,93 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import cn from 'classnames';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import type { Todo } from './types/Todo';
+import { Filter, type FilterStatus } from './components/Filter';
+import { Notification } from './components/Notification';
+import { TodoList } from './components/TodoList';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const shouldShowUserWarning = !USER_ID;
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('all');
+
+  const newTodoInputRef = useRef<HTMLInputElement>(null);
+  const hideErrorTimeoutIdRef = useRef<number | null>(null);
+
+  const clearErrorTimer = () => {
+    if (hideErrorTimeoutIdRef.current) {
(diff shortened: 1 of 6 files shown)


src/components/Filter.tsx
```diff
@@ -0,0 +1,41 @@
+import cn from 'classnames';
+
+export type FilterStatus = 'all' | 'active' | 'completed';
```
781calmstate: better to use enum

src/components/Filter.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+  value: FilterStatus;
+  onChange: (nextStatus: FilterStatus) => void;
+};
+
+const FILTERS: FilterStatus[] = ['all', 'active', 'completed'];
+const getHref = (status: FilterStatus) =>
+  status === 'all' ? '#/' : `#/${status}`;
```
781calmstate: should be enum value instead of 'all'

src/App.tsx
```diff
@@ -1,204 +1,146 @@
...
+        setTodos(data);
+      } catch {
+        showError('Unable to load todos');
+      } finally {
+        setIsLoading(false);
+      }
+    })();
+
+    return () => clearErrorTimer();
+  }, []);
```
781calmstate: it's not a good practice to omit the dependencies of useEffect, can lead to multiple bugs.

if you need to use functions, you create inside of component, inside of useEffect you have two options - memoize functions with useCallbacks, or declare the function inside of useEffect if you don't need the functionality anywhere else

src/App.tsx
```diff
@@ -1,204 +1,146 @@
...
+
+  if (shouldShowUserWarning) {
     return <UserWarning />;
   }
```
781calmstate: it's better practice to do conditional returns after you other component's code.

declare your callbacks, variables and then return

781calmstate (changes requested):

src/App.tsx
```diff
@@ -1,11 +1,93 @@
...
+    setErrorMessage(null);
+  }, []);
+
+  const showError = useCallback((message: string) => {
+    setErrorMessage(message);
+    clearErrorTimer();
+    hideErrorTimeoutIdRef.current = window.setTimeout(() => {
+      setErrorMessage(null);
+    }, 3000);
+  }, []);
```
781calmstate: you are omitting the dependency of `useCallback` which could lead to the same problems as with `useEffect`.

your `clearErrorTimer` is re-declared on every render, in your case it's not a problem because `clearErrorTimer` doesn't use any dependency that can change on render (it uses the ref value, but refs don't force re-renders and you don't need them as dependency, because it's an object that is being mutated when you change it).

but imagine the situation that one day, the `clearErrorTimer` will have some dependencies that can be changed on one of renders, since you didn't provide `clearErrorTimer`  as a dependency for `useCallback` the data in the cached instance of `clearErrorTimer` will be stalled.

it's a very wide topic, you can read about this a little more on your own. but to summarize, you need to specify as dependency everything you use inside that can be changed on render.

781calmstate (approved):

PR #2111, opened 2025-10-23
Solution

PolianskyiDmytro:
[DEMO LINK](https://PolianskyiDmytro.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,6 +11,7 @@ export const App: React.FC = () => {
   const todoInput = useRef<HTMLInputElement>(null);
 
   const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState<string>('');
   const [loading, setLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState<string>('');
 
@@ -39,10 +40,10 @@ export const App: React.FC = () => {
   const handleSubmit = (event: React.FormEvent) => {
     event.preventDefault();
 
-    if (todoInput.current?.value) {
+    if (title.trim().length > 0) {
       setLoading(true);
 
-      addTodo(todoInput.current?.value)
+      addTodo(title.trim())
         .then((response: Todo) => {
           setTodos([...todos, response]);
           if (todoInput.current) {
@@ -66,6 +67,11 @@ export const App: React.FC = () => {
         setErrorMessage('');
       }, 3000);
     }
+  };
+
+  const handleTitleChange = (value: string) => {
+    setTitle(value);
+    setErrorMessage('');
   };
 
   const handleRemoveError = () => {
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,10 +1,114 @@
...
-import React from 'react';
+import React, { useEffect, useRef, useState } from 'react';
+import { addTodo, getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import { TodoElement } from './Todo';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import classNames from 'classnames';
+
+type FilterStatus = 'all' | 'active' | 'completed';
```
h-amster: Please create an enum that includes all possible filter values and use this enum instead of the current UnityType. Also, move the enum to a separate file and import it to components as needed.

src/App.tsx
```diff
@@ -1,10 +1,114 @@
...
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [showedTodos, setShowedTodos] = useState<Todo[]>([]);
+  const [filterStatus, setFilterStatus] = useState<FilterStatus>('all');
+  const [title, setTitle] = useState<string>('');
+  const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+
+  const activeTodos = todos.filter(todo => !todo.completed);
+  const completedTodos = todos.filter(todo => todo.completed);
```
h-amster, line 29: Create a state called visibleTodos and store it using useMemo. Pass the filterStatus in the dependency array. When the filterStatus changes, filter the todos according to the active filter and use visibleTodos to render the filtered todos. Also remove showedTodos state

src/App.tsx
```diff
@@ -1,10 +1,114 @@
...
+
+  const handleActive = () => {
+    setShowedTodos(activeTodos);
+    setFilterStatus('active');
+  };
+
+  const handleCompleted = () => {
+    setShowedTodos(completedTodos);
+    setFilterStatus('completed');
+  };
```
h-amster: Create a single handler function that takes the status as a parameter and sets it. You don’t need three separate functions; just pass the status as an argument.

src/App.tsx
```diff
@@ -1,10 +1,114 @@
...
+
+      addTodo(title.trim())
+        .then((response: Todo) => {
+          setTodos([...todos, response]);
+          if (todoInput.current) {
+            todoInput.current.value = '';
+          }
+        })
+        .catch(error => {
+          setErrorMessage('Unable to add a todo');
```
h-amster: Do not pass the error text directly as a string. Instead, create a constant object or enum where the key is the error type and the value is the error text. Use this object or enum to display errors, since some errors are common and will be reused across different functions.

h-amster (changes requested):
Great work — it’s almost ready! Please check my comments and update your code accordingly. Also, split your App component into separate components: Header, TodoList, Footer, and ErrorNotification.

src/App.tsx
PolianskyiDmytro: Where do I find a type of error?

vadiimvooo (approved):
Well done

PR #2115, opened 2025-10-27
Develop

N3m0o:
https://n3m0o.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,6 +16,7 @@ enum TodoError {
   DELETE = 'Unable to delete a todo',
   UPDATE = 'Unable to update a todo',
 }
+//added some todos
 
 export const App: React.FC = () => {
   const [isLoading, setIsLoading] = useState(true);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,58 +8,49 @@ import { TodoList } from './components/TodoList';
 import { TodoFooter } from './components/TodoFooter';
 import { Loader } from './components/Loader';
 import { ErrorNotification } from './components/ErrorNotifications';
+import { Filter } from './types/Filter';
+import { TodoError } from './types/TodoError';
 
-enum TodoError {
-  LOAD = 'Unable to load todos',
-  EMPTY_TITLE = 'Title should not be empty',
-  ADD = 'Unable to add a todo',
-  DELETE = 'Unable to delete a todo',
-  UPDATE = 'Unable to update a todo',
-}
 //added some todos
 
 export const App: React.FC = () => {
   const [isLoading, setIsLoading] = useState(true);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState<string | null>(null);
-  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
 
   const activeCount = todos.filter(todo => !todo.completed).length;
 
   const filtredTodos = todos.filter(todo => {
-    if (filter === 'active') {
+    if (filter === Filter.Active) {
       return !todo.completed;
     }
 
-    if (filter === 'completed') {
+    if (filter === Filter.Completed) {
       return todo.completed;
     }
 
     return true;
   });
 
-  function loadTodos() {
+  useEffect(() => {
     setIsLoading(true);
     setError(null);
 
     getTodos()
-      .then(setTodos)
-      .catch(() => setError(TodoError.LOAD))
-      .finally(() => setIsLoading(false));
-  }
-
-  useEffect(() => {
-    loadTodos();
+    .then(setTodos)
+    .catch(() => setError(TodoError.LOAD))
+    .finally(() => setIsLoading(false));
   }, []);
 
   useEffect(() => {
-    if (error) {
-      const timer = setTimeout(() => setError(null), 3000);
-
-      return () => clearTimeout(timer);
+    if (!error) {
+      return;
     }
 
-    return undefined;
+    const timer = setTimeout(() => setError(null), 3000);
+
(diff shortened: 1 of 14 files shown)


src/App.tsx
```diff
@@ -33,171 +83,122 @@ export const App: React.FC = () => {
...
+                </button>
+
+                {/* overlay will cover the todo while it is being deleted or updated */}
+                <div data-cy="TodoLoader" className="modal overlay">
+                  <div className="modal-background has-background-white-ter" />
+                  <div className="loader" />
+                </div>
+              </div>
+            ))}
+          </section>
```
DmytroLymar: ## Task checklist.
1. `App.js` code should be split into several components.

src/App.tsx
```diff
@@ -33,171 +83,122 @@ export const App: React.FC = () => {
...
+            {/* This is a completed todo */}
+
+            {filtredTodos.map(todo => (
+              <div
+                key={todo.id}
+                data-cy="Todo"
+                className={cn('todo', {
+                  completed: todo.completed,
+                })}
+              >
```
DmytroLymar: ## Task checklist.
1. `App.js` code should be split into several components.

DmytroLymar (changes requested):
A working solution, but a checklist is required. The `App.js` code should be split into several components.

DmytroLymar (approved):
Good job!

DmytroLymar (approved):
Good job

mkyloa (changes requested):
hey! 👋 
the list does not show any todos, so I can't test the app :(

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import cn from 'classnames';
+/* eslint-disable @typescript-eslint/indent */
+type Props = {
+  activeCount: number;
+  filter: 'all' | 'active' | 'completed';
+  setFilter: React.Dispatch<
+    React.SetStateAction<'all' | 'active' | 'completed'>
+  >;
```
781calmstate: passing the setState function directly usually is a bad practice, more common way to do it is 
`onFilterChange: (value) => void;`

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <span className="todo-count" data-cy="TodosCounter">
+      {activeCount} {activeCount === 1 ? 'item' : 'items'} left
+    </span>
+
+    <nav className="filter" data-cy="Filter">
+      <a
+        href="#/"
+        className={cn('filter__link', { selected: filter === 'all' })}
+        data-cy="FilterLinkAll"
+        onClick={() => setFilter('all')}
```
781calmstate: i think it makes sense to move the filter values into enum, so in imaginable scenario where you would need to change some of the values you would have to change it in single place in the enum instead of every place you used it. Whad do you think?

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import cn from 'classnames';
+/* eslint-disable @typescript-eslint/indent */
+type Props = {
+  activeCount: number;
+  filter: 'all' | 'active' | 'completed';
```
781calmstate: probably would be better to use enum

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+        onClick={() => setFilter('all')}
+      >
+        All
+      </a>
+
+      <a
+        href="#/active"
+        className={cn('filter__link', { selected: filter === 'active' })}
+        data-cy="FilterLinkActive"
+        onClick={() => setFilter('active')}
```
781calmstate: same

src/components/TodoFooter.tsx
```diff
@@ -0,0 +1,60 @@
...
+      >
+        Active
+      </a>
+
+      <a
+        href="#/completed"
+        className={cn('filter__link', { selected: filter === 'completed' })}
+        data-cy="FilterLinkCompleted"
+        onClick={() => setFilter('completed')}
```
781calmstate: same

src/App.tsx
```diff
@@ -1,204 +1,88 @@
...
+}
+//added some todos
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
-    return <UserWarning />;
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
781calmstate: would be easier to have enum here

src/App.tsx
```diff
@@ -1,204 +1,88 @@
...
+
+  function loadTodos() {
+    setIsLoading(true);
+    setError(null);
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => setError(TodoError.LOAD))
+      .finally(() => setIsLoading(false));
   }
```
781calmstate: you don't need this function anywhere else, just declare it within `useEffect` instead?

src/App.tsx
```diff
@@ -1,204 +1,88 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (error) {
+      const timer = setTimeout(() => setError(null), 3000);
+
+      return () => clearTimeout(timer);
+    }
+
+    return undefined;
```
781calmstate: returning from `useEffect`'s callback means utilizing `componentWillUnmount` lifecycle, returning undefined doesn't do anything

781calmstate (changes requested):

alexander-ignatow (approved):
Lets move on to next iterations

PR #2121, opened 2025-10-28
[Todo App Loading Todos React-TS task]

Bohdan-Diatliuk:
[DEMO LINK](https://bohdan-diatliuk.github.io/react_todo-app-loading-todos/)


YuliiaKulish (approved):
👍

src/api/todos.ts
```diff
@@ -1,10 +1,18 @@
 import { Todo } from '../types/Todo';
 import { client } from '../utils/fetchClient';
 
-export const USER_ID = 0;
+export const USER_ID = 3626;
+
+const todoUrl = (id?: number) => `/todos${id ? `/${id}` : ''}`;
```
781calmstate: functions should be named so they indicate the action


```suggestion
const getTodoUrl = (id?: number) => `/todos${id ? `/${id}` : ''}`;
```

src/App.tsx
```diff
@@ -1,204 +1,159 @@
...
+
+  const focusedInput = useRef<HTMLInputElement>(null);
+
+  const allFilters = useMemo(() => {
+    return {
+      [FilterStatus.All]: () => true,
+      [FilterStatus.Active]: (td: Todo) => !td.completed,
+      [FilterStatus.Completed]: (td: Todo) => td.completed,
+    };
+  }, []);
```
781calmstate: this is static value, does it make sense to just move it out of component to avoid using useMemo(additional resources)? :)

src/types/FilterStatus.ts
781calmstate, line 1: only component's file/folder should have Capitalized name

781calmstate (changes requested):

src/components/TodoList.tsx
```diff
@@ -0,0 +1,26 @@
...
+          <input
+            type="checkbox"
+            className="todo__status"
+            checked={todo.completed}
+            readOnly
+          />
+        </label>
+
+        <span className="todo__title">{todo.title}</span>
+      </div>
```
alexander-ignatow: It makes sense to make a todo item a separate component

src/App.tsx
```diff
@@ -1,204 +1,77 @@
...
+    FilterStatus.All,
+  );
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const allFilters = {
+    [FilterStatus.All]: () => true,
+    [FilterStatus.Active]: (td: Todo) => !td.completed,
+    [FilterStatus.Completed]: (td: Todo) => td.completed,
+  };
+
```
alexander-ignatow: Nice idea with methods - but you should move it outside the component

src/App.tsx
```diff
@@ -1,204 +1,77 @@
...
+import Notifications from './components/Notifications';
+import { Todo } from './types/Todo';
+import { Filter } from './components/Filter';
+import { TodoList } from './components/TodoList';
+import { FilterStatus } from './types/filterStatus';
 
 export const App: React.FC = () => {
   if (!USER_ID) {
     return <UserWarning />;
   }
```
alexander-ignatow: That is a big mistake!

Never make a conditional return BEFORE calling ALL hooks

See the rules of hooks 🤓

alexander-ignatow (changes requested):
Declining the PR ONLY to make an extra point on the rules of hooks problem
You can fix only it - all other problems can be fixed in the upcomming iterations

Moroz-Dmytro (approved):
Approve as 3 part is already done and approved

PR #2147, opened 2025-11-09
finished with app.tsx

Kanezoor:
https://kanezoor.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,58 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useMemo } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { useState, useEffect } from 'react';
+import { Todo } from './types/Todo';
+import cn from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState('All');
+
+  const visibleTodos = useMemo(() => {
+    return todos.filter(todo => {
+      if (filterBy === 'Active') {
+        return !todo.completed;
+      }
+
+      if (filterBy === 'Completed') {
+        return todo.completed;
+      }
+
+      return true;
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { useState, useEffect } from 'react';
+import { Todo } from './types/Todo';
+import cn from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState('All');
```
h-amster: create an enum for filterBy. Use it in this state and in all cases where you need to use filter value

src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
+    return todos.filter(todo => {
+      if (filterBy === 'Active') {
+        return !todo.completed;
+      }
+
+      if (filterBy === 'Completed') {
+        return todo.completed;
+      }
+
+      return true;
```
h-amster: replace this construction with switch/case

src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
+  const countActive = useMemo(() => {
+    return todos.filter(todo => !todo.completed).length;
+  }, [todos]);
+
+  useEffect(() => {
+    setErrorMessage('');
+    getTodos()
+      .then(loadedTodos => setTodos(loadedTodos))
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
h-amster: It is better to create object or enum with all error messages and use it instead of strings

src/App.tsx
```diff
@@ -33,171 +81,127 @@ export const App: React.FC = () => {
...
+                    className="todo__status"
+                    checked={todo.completed}
+                  />
+                </label>
+
+                <span data-cy="TodoTitle" className="todo__title">
+                  {todo.title}
+                </span>
+
+                {/* Remove button appears only on hover */}
```
h-amster: Remove all comments before submitting a task

h-amster (changes requested):
Good job! Almost done. Fix your code according to my comments. Also, split your App component into smaller components(Header, TodoList, TodoItem, ErrorNotification, Footer)

maxim2310 (approved):
Well done

PR #2157, opened 2025-11-18
solution

IvanKovalyoff:
https://IvanKovalyoff.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,62 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo, Filter } from './types/Todo';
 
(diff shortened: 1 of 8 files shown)


brespect (changes requested):
1. Pass all tests before submit.
2. I can`t reach your deploy link:
{5AC35CB1-3D18-4DF1-B634-2796A0F1FEEA}

src/App.tsx
```diff
@@ -1,203 +1,201 @@
...
+
+                <button
+                  type="button"
+                  className="todo__remove"
+                  data-cy="TodoDelete"
+                  disabled
+                >
+                  ×
+                </button>
+              </div>
```
2pasha: it's better to move this code to separate component, like `TodoItem`

i mean all code which is about Todo Item

2pasha (changes requested):
almost done!

- please add some tasks from task's readme ([point 3](https://github.com/mate-academy/react_todo-app-loading-todos?tab=readme-ov-file#load-todos-by-userid))
- and please run tests before sending the task, now 1 task failed - please fix it

if you have some questions - please text it in the chat on platform

etojeDenys (changes requested):
all the tests passed, which is great 
however, it would be better to split the app into several components (header, main, and footer)

etojeDenys (approved):
good job

PR #2158, opened 2025-11-18
Solution for first task

BadOfficer:
[DEMO LINK](https://BadOfficer.github.io/react_todo-app-loading-todos/)


vadiimvooo (approved):
Well done. Just don't forget to split your app to smaller components like Header, Footer, Error, Filter etc

PR #2165, opened 2025-12-06
added solution

ArtemYakhno:
(https://ArtemYakhno.github.io/react_todo-app-loading-todos/)


src/types/FilterTodo.ts
```diff
@@ -0,0 +1 @@
+export type FilterTodo = 'all' | 'active' | 'completed';
```
etojeDenys, line 1: it would be better to use an enum here

etojeDenys (approved):
great job

to improve: 
move error block to a separate component

PR #2170, opened 2025-12-18
added solution

yahohulia:
[DEMO LINK](https://yahohulia.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,203 +1,422 @@
...
+  addTodos,
+  deleteTodos,
+  getTodos,
+  updateTodos,
+  USER_ID,
+} from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+
+type FilterType = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to create enum values rather than type

src/App.tsx
```diff
@@ -1,203 +1,422 @@
...
+              <a
+                href="#/"
+                className={classNames('filter__link', {
+                  selected: filterBy === 'all',
+                })}
+                data-cy="FilterLinkAll"
+                onClick={() => setFilterBy('all')}
+              >
+                All
+              </a>
```
etojeDenys: create an object with the buttons and parameters, and map that object to create the buttons

etojeDenys (changes requested):
1. demo does not work. please fix it and re-request the review

2. split app into several components (e.g., header, main, and footer)

brespect (changes requested):
Please, try to pass all tests before submitting and requesting the reviews from mentors

vadiimvooo (approved):
Well done

PR #2172, opened 2025-12-20
solution

Andrey1166:
https://Andrey1166.github.io/react_todo-app-loading-todos/


etojeDenys (changes requested):
almost done
1. make a different component for error

2. create an enum type for error values

3. it would be better to create an enum for the filter parameters instead of a type

etojeDenys (approved):
great job

PR #2185, opened 2026-01-05
add task solution

devsync25:


src/App.tsx
```diff
@@ -1,10 +1,40 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('all');
+
+  // Загрузка данных при старте
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
etojeDenys: create an enum value for errors

src/App.tsx
```diff
@@ -1,10 +1,40 @@
...
+import React, { useEffect, useState, useMemo } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState('all');
```
etojeDenys: create an enum value for the filter parameters

etojeDenys (changes requested):
good job

split the app into several components (for example, a header, a main section and a footer)

src/App.tsx
```diff
@@ -14,190 +48,36 @@ export const App: React.FC = () => {
...
+        <button
+          data-cy="HideErrorButton"
+          type="button"
+          className="delete"
+          onClick={() => {
+            setErrorMessage(null);
+          }}
+        />
+        {errorMessage}
       </div>
```
2pasha, line 81: it's better to move this code to separate component

2pasha (approved):
good job 👏

here is one more improvement ⬇️

PR #2186, opened 2026-01-06
Completed Task

Augustwise:
[DEMO LINK](https://Augustwise.github.io/react_todo-app-loading-todos/)

While working on the task, I found it so interesting that I did a bit more than what was required by the requirements - sorry about that :) 

I’ve already implemented todo deletion (using `deleteTodo`) and task status updates (using `updateTodo`), which simplified my work on the second part of the task.

The `classnames` library is used to display errors.

P.S. For now, I write the code using `then/catch`, but I plan to refactor it to `async/await` in the future, since it’s a more modern and cleaner approach

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,18 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React, { useEffect, useMemo, useRef, useState } from 'react';
-import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
 import * as todosApi from './api/todos';
 import type { Todo } from './types/Todo';
+import { ErrorNotification } from './components/ErrorNotification';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { Filter } from './types/Filter';
+import { ErrorMessage } from './types/ErrorMessage';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,11 +1,105 @@
...
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import * as todosApi from './api/todos';
+import type { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: create enum value for the filter parameters

src/App.tsx
```diff
@@ -1,11 +1,105 @@
...
+  const todoInput = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    setErrorMessage('');
+
+    todosApi
+      .getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
```
etojeDenys: create an enum value for errors

etojeDenys (changes requested):
good job

fix the comments and split the app into several components (for example header, main and footer)

Augustwise:
Created 4 components:


image

brespect (approved):
Awesome 🚀

PR #2187, opened 2026-01-06
add part of task solution

n7lots:
[DEMO LINK](https://n7lots.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,203 +1,61 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
-/* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
-import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import React, { useState, useEffect } from 'react';
+import { TodoappHeader } from './components/TodoappHeader';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+import { getTodos } from './api/todos';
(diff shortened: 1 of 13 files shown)


src/types/Filter.ts
```diff
@@ -0,0 +1 @@
+export type Filter = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to use enum instead of type

src/App.tsx
```diff
@@ -1,203 +1,61 @@
...
+  const [inputValue, setInputValue] = useState('');
+  const [filter, setFilter] = useState<Filter>('all');
+
+  useEffect(() => {
+    setErrorMessage('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```
etojeDenys: create an enum for errors

etojeDenys (changes requested):
good job

1. you should pass all the tests
image

2. that todo should be active
image

3. it looks like a redundant part
image

n7lots:
How to hide the Error Block, i use class hidden. But it appears every time 
image
What's wrong?
image

etojeDenys (changes requested):
if you have any problems, it is better to ask for help in the chat because it is difficult to explain things here

etojeDenys (approved):
great job

PR #2194, opened 2026-01-15
add task solution

IvanR222:
[DEMO LINK](https://IvanR222.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+
+type FilterType = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<FilterType>('all');
```

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+
+  // Load todos on component mount
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```
etojeDenys: create enum values for errors

etojeDenys (changes requested):
great job. just a few final fixes

1. split the app into several components (for example, header, main and footer)

brespect (changes requested):
Good progress, but try to pass all tests before submit. Also, previous comment from @etojeDenys was not resolved

Anton-Kuchmasov (changes requested):
Good job!

You should pass lint check before approving your solution. Try to fix it, if you're stuck - feel free to ask in fe_chat

etojeDenys (approved):
good job

PR #2195, opened 2026-01-16
Solution

Oskar274:
https://Oskar274.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,6 +52,20 @@ export const App: React.FC = () => {
       .catch(() => setError('LOAD_TODOS'));
   }, []);
 
+  useEffect(() => {
+    if (!error) {
+      return;
+    }
+
+    const timerId = setTimeout(() => {
+      setError(null);
+    }, 3000);
+
+    return () => {
+      clearTimeout(timerId);
+    };
+  }, [error]);
+
   if (!USER_ID) {
     return <UserWarning />;
   }
@@ -63,7 +77,9 @@ export const App: React.FC = () => {
       <div className="todoapp__content">
         <Header />
         <TodoList todos={visibleTodos} />
-        <Footer onStatusChange={setStatus} />
+        {todos.length > 0 && (
+          <Footer onStatusChange={setStatus} status={status} todos={todos} />
+        )}
       </div>
 
       {/* DON'T use conditional rendering to hide the notification */}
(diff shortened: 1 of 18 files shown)


2pasha (changes requested):
good job! 

your filter visually doesn't work
image

make sure that you have passed all tests, because right now i can see that some tests are failed

if you have some troubles - feel free to ask your question in fe_chat 💬

src/App.tsx
```diff
@@ -1,204 +1,90 @@
...
-          <nav className="filter" data-cy="Filter">
-            <a
-              href="#/"
-              className="filter__link selected"
-              data-cy="FilterLinkAll"
-            >
-              All
-            </a>
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState('all');
```
etojeDenys, line 42: create an enum value for the filter parameters

src/App.tsx
```diff
@@ -1,204 +1,90 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [query, setQuery] = useState('');
+  const [error, setError] = useState<ErrorType>(null);
```
etojeDenys, line 45: it would be better to use an enum here instead of a type

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+        <a
+          href="#/completed"
+          className={`filter__link ${status === 'completed' ? 'selected' : ''}`}
+          data-cy="FilterLinkCompleted"
+          onClick={() => {
+            onStatusChange('completed');
+          }}
+        >
+          Completed
```
etojeDenys, line 48: you can create an array with the 'links' parameter and then map through that array to create the buttons, to avoid code repetition

etojeDenys (changes requested):
1. right now, 'Completed' is selected, but the styles show that 'All' is selected. also, completed tasks should look different. you can view a working example [here](https://kazymyryan.github.io/react_todo-app-loading-todos/)
image

2. we have three items left, but there are only two todos
image

Oskar274:
I think it was bug with npm deploy because on my local host it works correctly.

src/components/Header/import.tsx
```diff
@@ -0,0 +1 @@
+export * from './Header';
```
Anton-Kuchmasov, line 1: better name this file `index.ts`

Anton-Kuchmasov (approved):
Good job!

PR #2200, opened 2026-01-20
progresss

wzfxc:
Demo link
https://wzfxc.github.io/react_todo-app-loading-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,48 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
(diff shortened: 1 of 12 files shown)


brespect (changes requested):
Hello, good progress! Let`s pass all tests before submit

src/App.tsx
```diff
@@ -1,10 +1,55 @@
...
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { TodoList } from './components/TodoList';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+import { Header } from './components/Header';
+import { Footer } from './components/Footer';
+
+export type Filter = 'all' | 'active' | 'completed';
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <a
+          href="#/completed"
+          className={classNames('filter__link', {
+            selected: selectedFilter === 'completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onSelect('completed')}
+        >
+          Completed
+        </a>
```
etojeDenys: you can create an array with the necessary parameters and map through that array to create links and avoid code repeating

etojeDenys (changes requested):
almost done

move error notification in separet component. also fix demo

2pasha (approved):
LGTM 🔥

PR #2201, opened 2026-01-20
is this right?

tvoypervy:
if i understand all tasks right, this must be solution for this task:)
[DEMO LINK](https://tvoypervy.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,12 +1,56 @@
...
+import { Filter } from './component/Filter';
+import { Notification } from './component/Notification';
+import { NewTodo } from './component/NewTodo';
+import { TodoList } from './component/TodoList';
+
+const FILTER = {
+  all: 'all',
+  active: 'active',
+  completed: 'completed',
+};
```
etojeDenys, line 28: it is better to create an enum value

src/App.tsx
```diff
@@ -1,12 +1,56 @@
...
+  const [filter, setFilter] = useState<string>(FILTER.all);
+  const [isTyping, setIsTyping] = useState<boolean>(false);
+
+  useEffect(() => {
+    setLoading(true);
+    setError(null);
+
+    getTodos()
+      .then(data => setTodos(data))
+      .catch(() => setError('Unable to load todos'))
```
etojeDenys: create an enum value for errors as well

src/component/Filter.tsx
```diff
@@ -0,0 +1,55 @@
...
+          selected: currentFilter === 'completed',
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={event => {
+          event.preventDefault();
+          onFilterChange('completed');
+        }}
+      >
+        Completed
+      </a>
```
etojeDenys, line 52: you can create an array containing the necessary parameters, then map through the array to create links and avoid code repetition

etojeDenys (changes requested):
almost done

1.  move the header to separate components
2. the demo does not work. please fix it

src/component/Filter.tsx
tvoypervy: I didn't make any changes to the filter.tsx file, I'm pretty sure this won't be a problem in the future (and quick fixed this in future, if it must to be:))

vadiimvooo (approved):
Well done

PR #2208, opened 2026-01-24
solution

eg598:


eg598:
[DEMO LINK](https://eg598.github.io/react_todo-app-loading-todos/)

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+
+type TodosError =
+  | 'Unable to load todos'
+  | 'Title should not be empty'
+  | 'Unable to add a todo'
+  | 'Unable to delete a todo'
+  | 'Unable to update a todo'
+  | null;
+
+type Filter = 'all' | 'active' | 'completed';
```
etojeDenys, line 16: it would be better to create enum values rather than types

src/App.tsx
```diff
@@ -34,170 +90,145 @@ export const App: React.FC = () => {
...
+                className={
+                  filter === 'completed'
+                    ? 'filter__link selected'
+                    : 'filter__link'
+                }
+                data-cy="FilterLinkCompleted"
+                onClick={() => handleFilter('completed')}
+              >
+                Completed
+              </a>
```
etojeDenys, line 195: you can create an array with the necessary parameters and then map through that array to create links and avoid code repeating

etojeDenys (changes requested):
good job. however, you should split the app into several components (e.g. header, main, footer and error notification)

PR #2225, opened 2026-01-27
Solution

Milana-Danylivska:
[DEMO LINK](https://Milana-Danylivska.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,28 +1,73 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [filterStatus, setFilterStatus] = useState('all');
+
+  // #region effects
+  useEffect(() => {
+    if (!USER_ID) {
+      return;
+    }
+
+    setErrorMessage('');
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
+  }, []);
+
+  useEffect(() => {
+    if (!errorMessage) {
+      return;
+    }
+
+    const timerId = setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+
+    return () => clearTimeout(timerId);
+  }, [errorMessage]);
+  // #endregion
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter(todo => {
+    if (filterStatus === 'active') {
+      return !todo.completed;
+    }
+
+    if (filterStatus === 'completed') {
+      return todo.completed;
+    }
+
+    return true;
+  });
+
+  // #region render
   return (
     <div className="todoapp">
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
           <button
             type="button"
-            className="todoapp__toggle-all active"
+            className={classNames('todoapp__toggle-all', {
+              active: todos.length > 0 && todos.every(todo => todo.completed),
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,28 +1,73 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [filterStatus, setFilterStatus] = useState('all');
+
+  // #region effects
+  useEffect(() => {
+    if (!USER_ID) {
+      return;
+    }
```
de-vasta, line 50: why not return `<UserWarning />` right here?

src/App.tsx
```diff
@@ -1,28 +1,73 @@
...
       <h1 className="todoapp__title">todos</h1>
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
           <button
             type="button"
-            className="todoapp__toggle-all active"
+            className={classNames('todoapp__toggle-all', {
+              active: todos.length > 0 && todos.every(todo => todo.completed),
```
de-vasta: i'd extract `todos.every(todo => todo.completed` to separate variable

src/App.tsx
```diff
@@ -33,172 +78,116 @@ export const App: React.FC = () => {
...
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Todo is being saved now
+        {todos.length > 0 && (
+          <section className="todoapp__main" data-cy="TodoList">
```
de-vasta, line 101: i'd break it to separate component

src/App.tsx
```diff
@@ -33,172 +78,116 @@ export const App: React.FC = () => {
...
+                </div>
+              </div>
+            ))}
+          </section>
+        )}
+
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span className="todo-count" data-cy="TodosCounter">
+              {todos.filter(todo => !todo.completed).length} items left
```
de-vasta: i'd extract it to variable too

src/App.tsx
```diff
@@ -33,172 +78,116 @@ export const App: React.FC = () => {
...
-
-            <a
-              href="#/completed"
-              className="filter__link"
-              data-cy="FilterLinkCompleted"
+            <nav className="filter" data-cy="Filter">
+              <a
+                href="#/"
+                className={classNames('filter__link', {
+                  selected: filterStatus === 'all',
```
de-vasta: i'm for using enums in cases like this instead of strings

de-vasta (approved):
👍🏼

src/App.tsx
dothearts: +, you can write a function getActiveTodosCount, and use in this place.

src/App.tsx
```diff
@@ -1,28 +1,73 @@
...
+
+    return () => clearTimeout(timerId);
+  }, [errorMessage]);
+  // #endregion
+
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const visibleTodos = todos.filter(todo => {
```
dothearts: use Enum for statuses. 
Also, you can make it a pure function (geVisibleTodos for example) move in separate file, then use in this place const visibleTodos = getVisibleTodos(..)

src/App.tsx
```diff
@@ -1,28 +1,73 @@
...
+  // #region effects
+  useEffect(() => {
+    if (!USER_ID) {
+      return;
+    }
+
+    setErrorMessage('');
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'));
```
dothearts: Also, you can create Enum for errors

DanilWeda (changes requested):
Please check the comments that I left. 
Many thanks for your solution!

src/App.tsx
```diff
@@ -1,10 +1,73 @@
...
+  Completed = 'completed',
+}
+
+enum ErrorMessage {
+  LoadTodos = 'Unable to load todos',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState<string>('');
```
ArturErmolenko, line 21: why string?
you create enum for this purpose - use it
```suggestion
  const [errorMessage, setErrorMessage] = useState<ErrorMessages>(ErrorMessages.Default);
```

ArturErmolenko (approved):
Well done 👍

PR #2229, opened 2026-01-28
load todos with filtering and error handling

SergiyKusiy:
[DEMO LINK](https://SergiyKusiy.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,21 +4,23 @@ import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
 import { getTodos, USER_ID } from './api/todos';
 import { TodoType } from './types/Todo';
-import { Filter as FilterComponent } from './components/Filter';
 import { ErrorMessage } from './types/AppError';
 import cn from 'classnames';
 import { NewTodo } from './components/NewTodo';
 import { TodoList } from './components/TodoList';
 import { FilterType } from './types/Filter';
+import { Footer } from './components/Footer';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<TodoType[]>([]);
-  const [loading, setLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState<ErrorMessage>(ErrorMessage.Default);
   const [filter, setFilter] = useState<FilterType>(FilterType.All);
 
   useEffect(() => {
-    setLoading(true);
+    let timeoutId: number | undefined;
+
+    setIsLoading(true);
 
     getTodos()
       .then(data => {
@@ -27,13 +29,19 @@ export const App: React.FC = () => {
       .catch(() => {
         setError(ErrorMessage.LoadTodos);
 
-        setTimeout(() => {
+        timeoutId = window.setTimeout(() => {
           setError(ErrorMessage.Default);
         }, 3000);
       })
       .finally(() => {
-        setLoading(false);
+        setIsLoading(false);
       });
+
+    return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
+    };
   }, []);
 
   const visibleTodos = todos.filter(todo => {
@@ -75,25 +83,12 @@ export const App: React.FC = () => {
         {/* overlay will cover the todo while it is being deleted or updated */}
 
         {hasTodos && (
-          <footer className="todoapp__footer" data-cy="Footer">
-            <span className="todo-count" data-cy="TodosCounter">
-              {activeCount} items left
-            </span>
-
-            <FilterComponent
-              filter={filter}
-              onFilterChange={handleFilterChange}
-            />
-
-            <button
-              type="button"
-              className="todoapp__clear-completed"
-              data-cy="ClearCompletedButton"
-              disabled={completedCount === 0}
-            >
-              Clear completed
-            </button>
-          </footer>
+          <Footer
+            activeCount={activeCount}
+            completedCount={completedCount}
+            filter={filter}
+            onFilterChange={handleFilterChange}
+          />
         )}
       </div>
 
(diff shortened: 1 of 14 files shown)


DemonWayne (changes requested):
You did a great job with this task, BUT I think you missed part about decomposition.
https://github.com/mate-academy/react_todo-app-loading-todos/blob/master/checklist.md#task-checklist

> App.js code should be split into several components.

src/types/Todo.ts
```diff
@@ -1,4 +1,4 @@
-export interface Todo {
+export interface TodoType {
```
DemonWayne, line 1: I'm not sure if naming a type "type" is a good idea. As an alternative, I suggest renaming the Todo component to `TodoItem`, which would make more sense to me than adding `Type` to the type name, although this change can be optional.

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+      default:
+        return true;
+    }
+  });
+
+  const handleFilterChange = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    newFilter: FilterType,
+  ) => {
+    event.preventDefault();
```
DemonWayne, line 62: I think using preventDefault is better exactly in a child component, but this approach can also be okay

DemonWayne (approved):
I think it's okay now, I provided some possible code improvements and you can implement them if you want

src/components/Filter.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onFilterChange: (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    newFilter: FilterType,
+  ) => void;
+};
+
+export const Filter: React.FC<Props> = ({ filter, onFilterChange }) => {
+  return (
+    <nav className="filter" data-cy="Filter">
+      <a
```
Moroz-Dmytro, line 15: You can use your eunm and map here, not to duplicate the same code 3 times

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+import { Filter as FilterComponent } from './components/Filter';
+import { ErrorMessage } from './types/AppError';
+import cn from 'classnames';
+import { NewTodo } from './components/NewTodo';
+import { TodoList } from './components/TodoList';
+import { FilterType } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<TodoType[]>([]);
+  const [loading, setLoading] = useState(false);
```
Moroz-Dmytro:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+    setLoading(true);
+
+    getTodos()
+      .then(data => {
+        setTodos(data);
+      })
+      .catch(() => {
+        setError(ErrorMessage.LoadTodos);
+
+        setTimeout(() => {
```
Moroz-Dmytro: Need to clear this timeout before component unmount

src/App.tsx
```diff
@@ -14,191 +68,54 @@ export const App: React.FC = () => {
...
-                data-cy="TodoStatus"
-                type="checkbox"
-                className="todo__status"
-              />
-            </label>
-
-            <span data-cy="TodoTitle" className="todo__title">
-              Todo is being saved now
+        {hasTodos && (
+          <footer className="todoapp__footer" data-cy="Footer">
```
Moroz-Dmytro: Create separate component for footer

Moroz-Dmytro (changes requested):
Why do you display spinner at the bottom of the page 🤔

src/App.tsx
```diff
@@ -1,10 +1,72 @@
...
+        return todo.completed;
+      default:
+        return true;
+    }
+  });
+
+  const handleFilterChange = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+    newFilter: FilterType,
+  ) => {
```
ArturErmolenko, line 61: why you need this function?
i don't understand
KISS

src/components/Filter.tsx
```diff
@@ -0,0 +1,47 @@
+import cn from 'classnames';
+import { FilterType } from '../types/Filter';
+
+type Props = {
+  filter: FilterType;
```
ArturErmolenko, line 5: bad naming
```suggestion
  filter: Filter;
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,39 @@
...
+        {activeCount} items left
+      </span>
+
+      <Filter filter={filter} onFilterChange={onFilterChange} />
+
+      <button
+        type="button"
+        className="todoapp__clear-completed"
+        data-cy="ClearCompletedButton"
+        disabled={completedCount === 0}
```
ArturErmolenko, line 33:
```suggestion
        disabled={!completedCount}
```

ArturErmolenko: or Boolean

src/components/Todo.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <label className="todo__status-label">
+        <input
+          data-cy="TodoStatus"
+          type="checkbox"
+          className="todo__status"
+          checked={todo.completed}
+          readOnly
+        />
+      </label>
+
```
ArturErmolenko, line 26: destructing todo

src/App.tsx
```diff
@@ -1,10 +1,72 @@
...
+      .finally(() => {
+        setIsLoading(false);
+      });
+
+    return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
+    };
+  }, []);
```
ArturErmolenko, line 45: why it looks so hard?
why we need this logic in one useEffect?
can we split it?

S- single responsibility

ArturErmolenko (approved):
Good start 👍 
Please fix comments in next part

PR #2243, opened 2026-02-02
task solution

mesu96:


src/App.tsx
```diff
@@ -33,171 +125,121 @@ export const App: React.FC = () => {
...
-              ×
-            </button>
-
-            {/* 'is-active' class puts this modal on top of the todo */}
-            <div data-cy="TodoLoader" className="modal overlay is-active">
-              <div className="modal-background has-background-white-ter" />
-              <div className="loader" />
+        <section
+          className={classNames('todoapp__main', {
+            hidden: isLoading || todos.length === 0,
```
zibi-dev: The requirement says "Hide list and footer if there are no todos" - but hiding during isLoading is incorrect.

src/App.tsx
```diff
@@ -33,171 +125,121 @@ export const App: React.FC = () => {
...
-              href="#/active"
-              className="filter__link"
-              data-cy="FilterLinkActive"
-            >
-              Active
-            </a>
+        {todos.length > 0 && (
+          <footer className="todoapp__footer" data-cy="Footer">
+            <span className="todo-count" data-cy="TodosCounter">
+              {`${todos.filter(t => !t.completed).length} items left`}
```
zibi-dev: Should be "1 item left" (singular) when count is 1.

src/App.tsx
```diff
@@ -33,171 +125,121 @@ export const App: React.FC = () => {
...
+              >
+                Completed
+              </a>
+            </nav>
+
+            <button
+              type="button"
+              className="todoapp__clear-completed"
+              data-cy="ClearCompletedButton"
+              disabled={todos.every(t => !t.completed)}
```
zibi-dev: The active todos count is calculated multiple times:
  - In the footer for "items left"
  - In ClearCompletedButton disabled check

src/App.tsx
```diff
@@ -1,10 +1,102 @@
...
+    event.preventDefault();
+    setFilter('active');
+  };
+
+  const handleFilterCompleted = (
+    event: React.MouseEvent<HTMLAnchorElement>,
+  ) => {
+    event.preventDefault();
+    setFilter('completed');
+  };
```
zibi-dev: Three nearly identical handlers can be simplified:

  const handleFilter = (status: FilterStatus) => (e: React.MouseEvent) => {
    e.preventDefault();
    setFilter(status);
  };

src/App.tsx
```diff
@@ -18,7 +110,7 @@ export const App: React.FC = () => {
           {/* this button should have `active` class only if all todos are completed */}
           <button
             type="button"
-            className="todoapp__toggle-all active"
+            className="todoapp__toggle-all"
```
zibi-dev: this should have active class when ALL todos are completed. Currently it never has the class.

Zibi95 (changes requested):
Add DEMO link so i can test it. So far 
- Good use of classNames library
- TypeScript types are correct

It would be good to split stuff around components so that App doeasnt hold whole logic

mesu96:
[DEMO LINK](https://mesu96.github.io/react_todo-app-loading-todos/)

src/App.tsx
```diff
@@ -1,10 +1,110 @@
...
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+
+type FilterStatus = 'all' | 'active' | 'completed';
+
+const ERROR_HIDE_DELAY = 3000;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [, setIsLoading] = useState(false);
```
zibi-dev: If not used then it can be removed

src/App.tsx
```diff
@@ -1,10 +1,110 @@
...
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+
+    return () => {
+      clearErrorTimeout();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
zibi-dev: That error handling is a bit hard to reason about. We could simplify it by adding one extra useEffect that will handle error clearing etc.

src/App.tsx
```diff
@@ -1,10 +1,110 @@
...
+
+  const completedCount = useMemo(
+    () => todos.filter(t => t.completed).length,
+    [todos],
+  );
+
+  const isAllCompleted = useMemo(
+    () => todos.length > 0 && todos.every(t => t.completed),
+    [todos],
+  );
```
zibi-dev: Its cheap, but still it could be done with one loop 

```suggestion
 const { activeCount, completedCount, isAllCompleted } = useMemo(() => {
    let active = 0;
    for (const t of todos) {
      if (!t.completed) active++;
    }
    return {
      activeCount: active,
      completedCount: todos.length - active,
      isAllCompleted: todos.length > 0 && active === 0,
    };
  }, [todos]);
```

src/App.tsx
```diff
@@ -33,171 +135,123 @@ export const App: React.FC = () => {
...
-            Clear completed
-          </button>
-        </footer>
+              <a
+                href="#/completed"
+                className={classNames('filter__link', {
+                  selected: filter === 'completed',
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={handleFilterCompleted}
```
zibi-dev: Three handler functions are created every render for a simple setFilter call.

Zibi95 (changes requested):
Left comments in code. Also please add some todos for that hardcoded USER_ID so I can test that web app

Zibi95 (approved):

PR #2250, opened 2026-02-07
add task solution

Ezero1:
[DEMO LINK](https://Ezero1.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,177 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
+import classNames from 'classnames';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import {
+  USER_ID,
+  getTodos,
+  createTodo,
+  updateTodo,
+  deleteTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,10 +1,177 @@
...
+  updateTodo,
+  deleteTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: it would be better to create an enum value for the filter parameters

src/App.tsx
```diff
@@ -1,10 +1,177 @@
...
+  const [title, setTitle] = useState('');
+  const [tempTodos, setTempTodos] = useState<number[]>([]);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```
etojeDenys: also, create an enum for errors

src/App.tsx
```diff
@@ -15,189 +182,144 @@ export const App: React.FC = () => {
...
+              <a
+                href="#/"
+                className={classNames('filter__link', {
+                  selected: filter === 'all',
+                })}
+                data-cy="FilterLinkAll"
+                onClick={() => setFilter('all')}
+              >
+                All
+              </a>
```
etojeDenys: you can create an array with the required parameters and then map through that array to create links and avoid code repetition

etojeDenys (changes requested):
good job

to improve:
split the app into several components (for example, header, main, footer and errorNotification)

etojeDenys (approved):
great job!

PR #2261, opened 2026-02-12
React + TS loading todos

Andrew17-2006:
[DEMO LINK](https://Andrew17-2006.github.io/react_todo-app-loading-todos/)

refactored todo app, added useEffect and fixed some bugs

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,30 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
-import { useState } from 'react';
 import { Todo } from './types/Todo';
 import { getTodos } from './api/todos';
+import { Filter } from './types/Filter';
+import { ErrorMessage } from './types/ErrorMessage';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [error, setError] = useState<string>('');
-  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
 
-  const loadtodos = async () => {
+  const loadTodos = async () => {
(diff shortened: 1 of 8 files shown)


brespect (changes requested):
Good progress, but let`s pass all tests before PR submit

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import { useState } from 'react';
+import { Todo } from './types/Todo';
+import { getTodos } from './api/todos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string>('');
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: create enum values for the filter parameters

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+
+  const loadtodos = async () => {
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (err) {
+      setError('Unable to load todos');
```
etojeDenys: also create an enum for errors

src/App.tsx
```diff
@@ -33,171 +80,103 @@ export const App: React.FC = () => {
...
+              </a>
+
+              <a
+                data-cy="FilterLinkCompleted"
+                href="#/completed"
+                className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
```
etojeDenys, line 161: you can create an array with the necessary parameters and then map through it to create a link and avoid code repetition

etojeDenys (changes requested):
1. you should pass all the tests in this task

2. split the app into several components (for example, header, main, footer and errorNotification)

src/App.tsx
Andrew17-2006: yes, but I completed it in the next part of the task, because this repository is the first part of the task, which consists of 3 parts, and here I pretty much completed everything according to the tasks and tests because there are no errors

vitalii-bondarenko-ua (approved):
Looks good

PR #2268, opened 2026-02-18
add solution task

Ilona-Tkachenko:


src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import classNames from 'classnames';
+import { Todo } from './types/Todo';
+
+type Filter = 'all' | 'active' | 'completed';
```

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+  const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<Filter>('all');
+
+  useEffect(() => {
+    setErrorMessage('');
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .catch(() => setErrorMessage('Unable to load todos'))
```
etojeDenys: create an enum for errors

src/App.tsx
```diff
@@ -29,175 +73,125 @@ export const App: React.FC = () => {
...
+                className={
+                  filter === 'completed'
+                    ? 'filter__link selected'
+                    : 'filter__link'
+                }
+                onClick={() => setFilter('completed')}
+                data-cy="FilterLinkCompleted"
+              >
+                Completed
+              </a>
```
etojeDenys: you can create an array with the necessary parameters and then map it to create links and avoid code repetition

etojeDenys (changes requested):
almost done

1. split the app into several components (for example header, main, footer and errorNotification

src/App.tsx
```diff
@@ -1,204 +1,90 @@
...
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
+
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
+
+enum ErrorMessage {
+  LoadTodos = 'Unable to load todos',
```
Anton-Kuchmasov, line 15: don't forget to add other ErrorMessages in further chapters of this task

Anton-Kuchmasov (approved):
Well done!

Check out comment below - it should be helpful

PR #2271, opened 2026-02-19
solution

Wiolip:
[DEMO LINK](https://wiolip.github.io/react_todo-app-loading-todos/)


danon321 (changes requested):
Zapodaj link do demo

src/types/Filters.ts
```diff
@@ -0,0 +1,7 @@
+export const FILTERS = {
+  all: 'all',
+  active: 'active',
+  completed: 'completed',
+} as const;
+
+export type Filter = keyof typeof FILTERS;
```
zibi-dev: Better derive from values, bcs u are using them. In this case key is the same as value so its fine.
```suggestion
export type Filter = (typeof FILTERS)[keyof typeof FILTERS];
```

src/App.tsx
```diff
@@ -1,204 +1,91 @@
...
-        Title should not be empty
-        <br />
-        Unable to add a todo
-        <br />
-        Unable to delete a todo
-        <br />
-        Unable to update a todo
-      </div>
+      <ErrorNotification message={error} onClose={() => setError(null)} />
+      {loading && <div className="loader overlay is-active" />}
```
zibi-dev: that loader is out of the UI.

src/App.tsx
```diff
@@ -1,204 +1,91 @@
...
+import { NewTodoForm } from './components/NewTodoForm';
+import { Todo } from './types/Todo';
+import { TodoList } from './components/TodoList';
+import { Filter, FILTERS } from './types/Filters';
+import { ErrorNotification } from './components/ErrorNotification';
+import { FilterComponent } from './components/FilterComponent';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
zibi-dev: Shouldn;t that be by default set to true?

src/App.tsx
```diff
@@ -1,204 +1,91 @@
...
+  const [filter, setFilter] = useState<Filter>('all');
+  const [error, setError] = useState<string | null>(null);
+
+  const showError = (message: string) => {
+    setError(message);
+
+    setTimeout(() => setError(null), 3000);
```
zibi-dev: timeout is not cleaned up.  If the component unmounts before 3 seconds or if showError is called multiple times, stale timeouts will fire and may cause state updates on unmounted components or hide newer errors prematurely.

src/App.tsx
```diff
@@ -1,204 +1,91 @@
...
   if (!USER_ID) {
     return <UserWarning />;
   }
 
+  const filteredTodos =
+    filter === FILTERS.all
+      ? todos
+      : filter === FILTERS.active
+        ? todos.filter(t => !t.completed)
+        : todos.filter(t => t.completed);
```
zibi-dev: nested ternary is a code smell
```suggestion
  const filteredTodos = useMemo(() => {
    switch (filter) {
      case FILTERS.active: return todos.filter(t => !t.completed);
      case FILTERS.completed: return todos.filter(t => t.completed);
      default: return todos;
    }
  }, [todos, filter]);
```

Zibi95 (changes requested):

Zibi95 (approved):

PR #2296, opened 2026-03-10
The first part is ready

aCTORr8:
[DEMO LINK](https://aCTORr8.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,58 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+
+type FilterStatus = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState<FilterStatus>('all');
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,59 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+
+type FilterStatus = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to use enum here

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filterBy, setFilterBy] = useState<FilterStatus>('all');
+
+  useEffect(() => {
+    setErrorMessage('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```

src/App.tsx
```diff
@@ -33,171 +78,114 @@ export const App: React.FC = () => {
...
+                className={
+                  filterBy === 'active'
+                    ? 'filter__link selected'
+                    : 'filter__link'
+                }
+                data-cy="FilterLinkActive"
+                onClick={() => setFilterBy('active')}
+              >
+                Active
+              </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create a link and avoid code repetition

etojeDenys (changes requested):
good job

to improve:
1. split the app into several components (for example, header, main, footer, errorNotification)

brespect (approved):
Lgtm

PR #2297, opened 2026-03-11
solution

Skay9n:
[DEMO LINK](https://Skay9n.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-len */
 /* eslint-disable react-hooks/rules-of-hooks */
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+type filter = 'all' | 'active' | 'completed' | '';
```

src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState('');
+  const [filter, setFilter] = useState<filter>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```

src/App.tsx
```diff
@@ -32,172 +51,106 @@ export const App: React.FC = () => {
...
+                Active
+              </a>
+              <a
+                href="#/completed"
+                className={`filter__link${filter === 'completed' ? ' selected' : ''}`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
```
etojeDenys: you can create an array with the needed parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job

to improve:

brespect (approved):
Good job!

PR #2298, opened 2026-03-13
Implemented filtering and fetch data from server

friranger1:


src/App.tsx
```diff
@@ -1,10 +1,43 @@
...
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [toDos, setTodos] = useState<Todo[]>([]);
+  const [error, setError] = useState<string | null>(null);
+  const [filter, setFilter] = useState<string>('All');
```
etojeDenys: create enum for filter parameters

src/App.tsx
```diff
@@ -1,10 +1,43 @@
...
+    }, 3000);
+  }
+
+  useEffect(() => {
+    setError(null);
+    getTodos()
+      .then(data => {
+        setTodos(data);
+      })
+      .catch(() => showError('Unable to load todos'));
```
etojeDenys, line 64: create enum for errors as well

src/App.tsx
```diff
@@ -34,137 +67,86 @@ export const App: React.FC = () => {
...
+                setFilter('All');
+              }}
             >
               All
             </a>
 
             <a
               href="#/active"
-              className="filter__link"
+              className={`filter__link ${filter === 'Active' ? 'selected' : ''}`}
```
etojeDenys: it would be better to use the className library here

src/App.tsx
```diff
@@ -34,137 +67,86 @@ export const App: React.FC = () => {
...
               href="#/"
-              className="filter__link selected"
+              className={`filter__link ${filter === 'All' ? 'selected' : ''}`}
               data-cy="FilterLinkAll"
+              onClick={() => {
+                setFilter('All');
+              }}
             >
               All
             </a>
```

etojeDenys (changes requested):
1. make sure you pass all the tests

Image

2. split the app into several components (for example, header, main, footer and errorNotification)

3. provide a demo in the description of the pull request

etojeDenys (changes requested):
demo does not work. please fix it and request the review again

friranger1:
At the moment I don't receive "todo" from the server because for my account I don't have it. I solved through the adding temporary array with 2 "todos". Unfortunately, in this case some tests doesn't passing. Hopfully it'll work

etojeDenys (changes requested):
demo link still does not work. please fix it and provide a demo link in the description of the pull request

PR #2311, opened 2026-03-21
add task solution

IvanovvvIhor:
[DEMO LINK](https://IvanovvvIhor.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,15 +4,18 @@ import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
 import { getTodos, USER_ID } from './api/todos';
 import { Todo } from './types/Todo';
-import classNames from 'classnames';
+import { Header } from './components/Header';
+import { TodoList } from './components/TodoList';
+import { Footer } from './components/Footer';
+import { ErrorNotification } from './components/ErrorNotification';
 
-enum Filter {
+export enum Filter {
   All = 'All',
   Active = 'Active',
   Completed = 'Completed',
 }
 
-enum Errors {
+export enum Errors {
   Load = 'Unable to load todos',
   Title = 'Title should not be empty',
   UnableTodo = 'Unable to add a todo',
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+
+const FILTER_ALL = 'All';
+const FILTER_ACTIVE = 'Active';
+const FILTER_COMPLETED = 'Completed';
```
etojeDenys: good idea, but it would be better to create an enum for the filter parameters

src/App.tsx
```diff
@@ -1,10 +1,63 @@
...
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      try {
+        const data = await getTodos();
+
+        setTodos(data);
+      } catch (error) {
+        // Замість throw error, встановлюємо стан помилки для UI
+        setErrorMessage('Unable to load todos');
```

src/App.tsx
```diff
@@ -34,170 +89,134 @@ export const App: React.FC = () => {
...
+                className={classNames('filter__link', {
+                  selected: filterStatus === FILTER_COMPLETED,
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => {
+                  setFilterStatus(FILTER_COMPLETED);
+                }}
+              >
+                Completed
+              </a>
```

etojeDenys (changes requested):
almost done

split the app into several components (for example: header, main, footer, errorNotification)

vadiimvooo (changes requested):
Almost done. One comment still not fixed:
split the app into several components (for example: header, main, footer, errorNotificatio
Please feel free to ask for help in fe_chat.

brespect (approved):
Lgtm

PR #2321, opened 2026-04-03
Develop

AlexanderBarkar:


src/App.tsx
```diff
@@ -1,10 +1,176 @@
...
+  getTodos,
+  addTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+
+type Filter = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to use an enum here. also, move the types to a separate file

src/App.tsx
```diff
@@ -1,10 +1,176 @@
...
+
+  const loadTodos = async () => {
+    setError('');
+
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch {
+      setError('Unable to load todos');
```
etojeDenys, line 28: create an enum for errors as well

src/App.tsx
```diff
@@ -15,189 +181,148 @@ export const App: React.FC = () => {
 
       <div className="todoapp__content">
         <header className="todoapp__header">
-          {/* this button should have `active` class only if all todos are completed */}
+          {/* ✅ ВЕРНУЛИ ToggleAllButton */}
```
etojeDenys: remove redundant comments

src/App.tsx
```diff
@@ -15,189 +181,148 @@ export const App: React.FC = () => {
...
-            Clear completed
-          </button>
-        </footer>
+              <a
+                data-cy="FilterLinkCompleted"
+                className="filter__link"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
```

etojeDenys (changes requested):
1. split the app into several components (for example: header, main, footer, and errorNotification)

vitalii-bondarenko-ua (approved):
Looks good

PR #2326, opened 2026-04-14
Part 1 - implement load todos

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_todo-app-loading-todos/)


src/components/ErrorNotification/ErrorNotification.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useEffect, useRef } from 'react';
+import cn from 'classnames';
+
+type Props = {
+  message: string;
+  hidden: boolean;
+};
+
+const ErrorNotificationBase: React.FC<Props> = ({ message, hidden }) => {
+  const [isHidden, setIsHidden] = React.useState(hidden);
```
polosanya: It is a bad practice to set props as a state directly. Better use `hidden` directly and update it in parent component or pass the callback to change it as props too

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,71 @@
...
+            selected: appliedFilter === 'Completed',
+          })}
+          data-cy="FilterLinkCompleted"
+          onClick={() => onFilterChange('Completed')}
+        >
+          Completed
+        </a>
+      </nav>
+
+      {/* this button should be disabled if there are no completed todos */}
```
polosanya, line 59: Shouldn't it be implemented?

src/App.tsx
```diff
@@ -1,10 +1,68 @@
...
+
+const ERROR_MESSAGES = {
+  failedLoadingTodos: 'Unable to load todos',
+  failedAddingTodo: 'Unable to add a todo',
+  failedDeletingTodo: 'Unable to delete a todo',
+  failedUpdatingTodo: 'Unable to update a todo',
+  emptyTitle: 'Title should not be empty',
+};
+
+export type Filter = 'All' | 'Active' | 'Completed';
```
polosanya: Better use Enum instead

src/App.tsx
```diff
@@ -1,10 +1,68 @@
...
+      })
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let filteredTodos: Todo[];
+
+    switch (appliedFilter) {
+      case 'Active':
+        filteredTodos = todos.filter(todo => !todo.completed);
```
polosanya:
```suggestion
      case 'Active':
        return todos.filter(todo => !todo.completed);
```
`filteredTodos` variable is redundant

polosanya (changes requested):

src/components/Footer/Footer.tsx
Dianaeva: I assume, that it should not be implemented in this part because all tests are passed, there is no mention about it in description. Also it's implemented in the second part.

polosanya (approved):

PR #2328, opened 2026-04-15
react_todo_app_load_todos

TarasYelahin:
https://TarasYelahin.github.io/react_todo-app-loading-todos/


brespect (changes requested):
Seems like your App is broken:
Image

TarasYelahin:
fixed

src/App.tsx
```diff
@@ -1,10 +1,119 @@
...
+  const [loading, setLoading] = useState<boolean>(false);
+  const [notification, setNotification] = useState<{
+    message: string | null;
+    visible: boolean;
+  }>({
+    message: null,
+    visible: false,
+  });
+
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: it would be better to create an enum value for the filter parameters

src/App.tsx
```diff
@@ -1,10 +1,119 @@
...
+import { useState, useRef, useEffect, useMemo, useCallback } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState<boolean>(false);
```
etojeDenys: it is not necessary to specify simple types like that
```suggestion
  const [loading, setLoading] = useState(false);
```

src/App.tsx
```diff
@@ -1,10 +1,119 @@
...
+
+  const allCompleted = todos.length > 0 && todos.every(item => item.completed);
+
+  function handleAddTodo(e: React.FormEvent) {
+    e.preventDefault();
+    hideNotification();
+    const title = newTodoText.trim();
+
+    if (!title) {
+      showNotification('title should not be empty');
```
etojeDenys: create an enum for notifications as well

src/App.tsx
```diff
@@ -15,189 +124,161 @@ export const App: React.FC = () => {
...
+                  event.preventDefault();
+                  setFilter('active');
+                }}
+                className={classNames('filter__link', {
+                  selected: filter === 'active',
+                })}
+                data-cy="FilterLinkActive"
+              >
+                Active
+              </a>
```
etojeDenys: you can create an array with the needed parameters and map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job
to improve:

1. split the app into several components (for example, header, footer,  main section, and errorNotification)
2. fix the problems highlighted in the comments

src/App.tsx
TarasYelahin: implemented suggested changes in
3ff204b37eb8a1a9fafd547dd09490dd9800eab2

brespect (approved):
Well done!

PR #2345, opened 2026-05-03
add task solution

freelinex:


brespect (approved):
Good job, but in next part you should separate your App.tsx into smaller components

PR #2357, opened 2026-05-11
start

flyman13:


2pasha (changes requested):
you should finish your task and pass all tests

src/App.tsx
```diff
@@ -1,203 +1,330 @@
...
+import {
+  USER_ID,
+  getTodos,
+  createTodo,
+  deleteTodo,
+  updateTodo,
+} from './api/todos';
+import { Todo } from './types/Todo';
+
+type Filter = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to create an enum than a type

src/App.tsx
```diff
@@ -1,203 +1,330 @@
...
+    }
+    const input = newTodoRef.current;
+    if (!input) {
+      return;
+    }
+    const raw = input.value;
+    const title = raw.trim();
+
+    if (!title) {
+      showError('Title should not be empty');
```

src/App.tsx
```diff
@@ -1,203 +1,330 @@
...
+              </a>
+              <a
+                href="#/completed"
+                className={`filter__link ${filter === 'completed' ? 'selected' : ''}`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
+            </nav>
```

etojeDenys (changes requested):
good job, to improve:

1. split the app into several components (for example, header, main, footer and errorNotification)

2. make sure the demo works correctly

vadiimvooo (approved):
Well done

PR #2361, opened 2026-05-14
success

KKRuslan:
[DEMO LINK](https://KKRuslan.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,203 +1,302 @@
...
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import * as todoService from './api/todos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  if (!USER_ID) {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<'all' | 'active' | 'completed'>('all');
```
etojeDenys: it would be better to create an enum for the filter parameters

src/App.tsx
```diff
@@ -1,203 +1,302 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  useEffect(() => {
+    setIsLoading(true);
+    todoService
+      .getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setErrorMessage('Unable to load todos');
```

src/App.tsx
```diff
@@ -1,203 +1,302 @@
...
+                  className={
+                    status === 'completed'
+                      ? 'filter__link selected'
+                      : 'filter__link'
+                  }
+                  data-cy="FilterLinkCompleted"
+                  onClick={() => setStatus('completed')}
+                >
+                  Completed
+                </a>
```

etojeDenys (changes requested):
good job, to improve:
1. split the app into several components, for example: header, main, footer, and errorNotification

etojeDenys (approved):
good job

PR #2365, opened 2026-05-16
add task solution

Ivadriil:
[DEMO LINK](https://Ivadriil.github.io/react_todo-app-loading-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,180 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import * as postService from './api/todos';
+import cn from 'classnames';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [erorDelet, setErorDelet] = useState(false);
+  const [todosError, setTodosError] = useState(false);
+  const [addTodoError, setAddTodoError] = useState<boolean | string>(false);
+  const [ErorUppdate, setErorUppdate] = useState(false);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+  const [editTitle, setEditTitle] = useState('');
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [erroAll, setErroAll] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [category, setCategory] = useState('all');
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+
+  useEffect(() => {
+    if (!erroAll) {
+      return;
+    }
+
+    const timer = setTimeout(() => {
+      setErroAll(false);
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [erroAll]);
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setTodosError(true);
+        setErroAll(true);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,10 +1,180 @@
...
+  const [todosError, setTodosError] = useState(false);
+  const [addTodoError, setAddTodoError] = useState<boolean | string>(false);
+  const [ErorUppdate, setErorUppdate] = useState(false);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+  const [editTitle, setEditTitle] = useState('');
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [erroAll, setErroAll] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const [category, setCategory] = useState('all');
```
etojeDenys: it would be better to use an enum here

src/App.tsx
```diff
@@ -35,169 +212,175 @@ export const App: React.FC = () => {
...
+                href="#/completed"
+                className={cn('filter__link', {
+                  selected: category === 'completed',
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setCategory('completed')}
+              >
+                Completed
+              </a>
+            </nav>
```

etojeDenys (changes requested):
almost done

1. split the app into several components, for example: header, main, footer and error notification

src/componets/Main.tsx
```diff
@@ -0,0 +1,102 @@
...
+              onChange={() => handleChangeComplete(todo.id)}
+              aria-label="Todo title"
+            />
+          </label>
+          {editingTodoId === todo.id ? (
+            <form
+              onSubmit={event => {
+                event.preventDefault();
+                handleEditSubmit(todo.id);
+              }}
```
2pasha, line 56: move to separate fn

src/componets/Errors.tsx
```diff
@@ -0,0 +1,61 @@
...
+  ErorUppdate: boolean;
+};
+
+export const Errors: React.FC<Props> = ({
+  erroAll,
+  setErroAll,
+  todosError,
+  addTodoError,
+  erorDelet,
+  ErorUppdate,
```
2pasha: no need to create states for every type of errors, it's better to pass error message here directly

*create separate enum for all your error messages and use it

2pasha (changes requested):
good job! 👏

take a look on comments below ⬇️

src/componets/Errors.tsx
```diff
@@ -0,0 +1,51 @@
...
+      })}
+    >
+      <button
+        data-cy="HideErrorButton"
+        type="button"
+        className="delete"
+        onClick={() => setErroAll(TypeErroros.Normal)}
+      />
+      {/* show only one message at a time */}
+      {erroAll === TypeErroros.NotFindTodosErrors && <>Unable to load todos</>}
```
2pasha: i mean in enum you can save exactly error message

```ts
NotFindTodosErrors = 'Unable to load todos',
...
```

and use this values direct in this component and you do not need to check is `erroAll` has some specific value

src/App.tsx
```diff
@@ -1,10 +1,200 @@
...
+import { TypeErroros } from './types/Errors';
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodoId, setLoadingTodoId] = useState<number | null>(null);
+  const [editTitle, setEditTitle] = useState('');
+  const [editingTodoId, setEditingTodoId] = useState<number | null>(null);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+  const activeTodosCount = todos.filter(todo => !todo.completed).length;
+  const [erroAll, setErroAll] = useState<TypeErroros>(TypeErroros.Normal);
```
2pasha: i suppose it will be better to call it just `error` and `setError`

2pasha (changes requested):
good job! 👏

take a look on comments below ⬇️

etojeDenys (approved):
good job

PR #2369, opened 2026-05-18
add task solution

nikassafronovs:


brespect (changes requested):
Good progress, but you need to fix all linter checks

vitalii-bondarenko-ua (approved):
Looks good

PR #2377, opened 2026-05-19
Solution

Misharkus:
[DEMO LINK](https://Misharkus.github.io/react_todo-app-loading-todos/)


sergeisvidlo4 (approved):
Luke will probably not accept the task. I checked it quickly and I don’t see autofocus, and it’s also not working in the project itself, while the task requirements say: "text field should be focused by default;".

Also, adding todos is not working for me, but that might be caused by the extension.

dothearts (approved):

PR #2383, opened 2026-05-21
solution

KaramushkoAndrii:


src/App.tsx
```diff
@@ -1,17 +1,51 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [filter, setFilter] = useState<FilterType>('All');
+
+  useEffect(() => {
+    getTodos()
+      .then(data => setTodos(data))
+      .catch(() => setErrorMessage('Unable to load todos'))
```
etojeDenys: create enum values for errors as well

src/App.tsx
```diff
@@ -1,17 +1,51 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { UserWarning } from './UserWarning';
 import { USER_ID } from './api/todos';
+import { getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+
+type FilterType = 'All' | 'Active' | 'Completed';
```
etojeDenys: it would be better to use enum here

src/App.tsx
```diff
@@ -33,171 +67,128 @@ export const App: React.FC = () => {
...
+
+              <a
+                href="#/completed"
+                className={`filter__link ${filter === 'Completed' ? 'selected' : ''}`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('Completed')}
+              >
+                Completed
+              </a>
+            </nav>
```

etojeDenys (changes requested):
good job, to improve:
1. split the app into several components (for example: header, main, footer and errorNotification)

2pasha (approved):
lgtm 🔥

PR #2384, opened 2026-05-21
add solution

MaxShaggy:
https://MaxShaggy.github.io/react_todo-app-loading-todos/


src/App.tsx
```diff
@@ -1,10 +1,66 @@
...
+
+    const timer = setTimeout(() => {
+      setErrorMessage('');
+    }, 3000);
+
+    return () => clearTimeout(timer);
+  }, [errorMessage]);
+
+
+  const visibleTodos = useMemo(() => {
```
dothearts, line 48: You can create a function, and move it in separate file, for example, helpers.ts 
And in this place, use only getVisibleTodos function (pure)

src/App.tsx
```diff
@@ -1,10 +1,66 @@
...
+
+  const loadTodos = () => {
+    setErrorMessage('');
+
+    getTodos()
+    .then(loadedTodos => {
+      setTodos(loadedTodos);
+    })
+    .catch(() => {
+      setErrorMessage('Unable to load todos');
```
dothearts, line 27: You can move all errors in enum/constants

dothearts (approved):
I left several comments. 
Please fix it in the next iteration. 

Many thanks!

PR #2393, opened 2026-05-27
solution

lilp3ka:
https://lilp3ka.github.io/react_todo-app-loading-todos/


andriy-stetsula (changes requested):
The task is directed to part 1 of the work, and in this task all 3 are completed, so after part 1, it needs to be transferred to part 2, and from 2 to 3

PR #2412, opened 2026-06-20
Solution

azesmmisha:
https://azesmmisha.github.io/react_todo-app-loading-todos/


vadiimvooo (approved):
Well done. Consider to split app on smaller components like TodoItem, TodoList, Footer, Error etc.

PR #2420, opened 2026-07-03
TODO part 1

masteryada378:
[DEMO LINK](https://masteryada378.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,29 +1,74 @@
...
+
+  useEffect(() => {
+    if (!USER_ID) {
+      return;
+    }
+
+    setErrorMessage('');
+    getTodos()
+      .then(setTodos)
+      .catch(() => showError('Unable to load todos'));
```

src/components/Filter.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <a
+        href="#/completed"
+        className={classNames('filter__link', {
+          selected: filter === FilterType.Completed,
+        })}
+        data-cy="FilterLinkCompleted"
+        onClick={() => onFilterChange(FilterType.Completed)}
+      >
+        Completed
+      </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
code looks good, but the demo does not work. please fix it and request the review again

src/App.tsx
masteryada378: done

src/components/Filter.tsx
masteryada378: done

etojeDenys (approved):
lgtm

PR #2426, opened 2026-07-10
change 1

PavloMolytovnyk:


vadiimvooo (approved):
Well done. Consider to split app to smaller components in the next task (TodoItem, TodoList, Footer, Error and so on)

PR #2428, opened 2026-07-12
Add Solution

CodableBase:


src/api/todos.ts
```diff
@@ -1,7 +1,27 @@
...
+  }
+
+  try {
+    const savedUser = window.localStorage.getItem('user');
+
+    if (!savedUser) {
+      return 0;
+    }
+
+    const parsedUser = JSON.parse(savedUser) as { id?: number };
```
Denys-Kravchuk9988: It's better to extract `{ id?: number }` into a separate type and use it here

src/App.tsx
```diff
@@ -33,171 +94,120 @@ export const App: React.FC = () => {
...
-              data-cy="FilterLinkCompleted"
+            <nav className="filter" data-cy="Filter">
+              <a
+                href="#/"
+                className={classNames('filter__link', {
+                  selected: filter === FILTERS.all,
+                })}
+                data-cy="FilterLinkAll"
+                onClick={event => {
+                  event.preventDefault();
```
Denys-Kravchuk9988: I would recommend to extract in to a separate callback and use it here. The logic is repeatable, only parameter is changed. `(e) => handleFilter(e, FILTERS.all)` would be better. The following click handlers also

Denys-Kravchuk9988 (changes requested):
Checked this link: https://codablebase.github.io/react_todo-app-loading-todos

The demo link doesn't work. Try to re-deploy. If you have any questions - feel free to ask in chat

roman-mirzoian (changes requested):
Clicking this link does not open the project itself; instead, a message appears stating that you need to use the correct ID. Please check that you have deployed and updated the project correctly, and provide the correct link to your project's demo:
https://codablebase.github.io/react_todo-app-loading-todos
Image

src/App.tsx
2pasha, line 1: separate your App to components

src/App.tsx
```diff
@@ -1,10 +1,79 @@
...
+    const loadTodos = async () => {
+      setIsErrorVisible(false);
+      setErrorMessage('');
+
+      try {
+        const loadedTodos = await getTodos();
+
+        setTodos(loadedTodos);
+      } catch {
+        setErrorMessage('Unable to load todos');
```
2pasha: move all your error messages to `enum`

src/App.tsx
```diff
@@ -33,171 +102,115 @@ export const App: React.FC = () => {
...
+              <a
+                href="#/completed"
+                className={classNames('filter__link', {
+                  selected: filter === FILTERS.completed,
+                })}
+                data-cy="FilterLinkCompleted"
+                onClick={handleFilterChange(FILTERS.completed)}
+              >
+                Completed
+              </a>
```
2pasha: you may add all your filter values to kind of array and map through it to avoid code duplication

2pasha (changes requested):
previous comment still actual 

Image

take a look on comments below ⬇️

src/components/Footer.tsx
```diff
@@ -0,0 +1,53 @@
...
+            key={filterValue}
+            href={`#/${filterValue === Filter.All ? '' : filterValue}`}
+            className={classNames('filter__link', {
+              selected: filter === filterValue,
+            })}
+            data-cy={`FilterLink${filterValue.charAt(0).toUpperCase() + filterValue.slice(1)}`}
+            onClick={e => {
+              e.preventDefault();
+              setFilter(filterValue);
+            }}
```
2pasha: move to separate fn

2pasha (changes requested):
i suppose you have forgot to deploy your app 

Image

because issue is still actual 

Image

2pasha (approved):
lgtm 🔥

PR #2433, opened 2026-07-16
React-todo-app-loading-todos 1.0

FRANCISCOPELEGRINI:


raulriato (approved):
The only issue with your code is that the `error message` is appearing while the `todos` are still being loaded. This should not happen. But since it's not a major problema, I'll just recommend you fix it for the next steps of this projetct.

PR #2442, opened 2026-07-22
Develop

IronHedgehog:
[DEMO LINK](https://IronHedgehog.github.io/react_todo-app-loading-todos/)


Denys-Kravchuk9988 (changes requested):
Demo link doesn't work. Try to re-deploy the project

roman-mirzoian (changes requested):
Great start!
Things to improve:
- We need to add a placeholder to display a temporary “TODO” and show it during the loading process until the item is successfully created, so that the user can see that something is happening even with a slow internet connection
- The same applies to deletion: currently, the “TODO” disappears before the server returns a confirmation that the deletion was successful

https://gyazo.com/79ff3475ab167ae2fce6e4f2392c9593

src/App.tsx
```diff
@@ -1,204 +1,244 @@
...
+
+      setIsLoading(true);
+      setErrorMessage('');
+
+      try {
+        const fetchedTodos = await getTodos();
+
+        setTodos(fetchedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to load todos');
```
2pasha: move all your error messages to `enum`

2pasha (changes requested):
good job! 👏

take a look on comment below ⬇️

2pasha (approved):
lgtm 🔥

PR #2448, opened 2026-07-24
Solution

kseniiavasylenko:


src/App.tsx
```diff
@@ -1,204 +1,232 @@
...
+
+  // > Fetching
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        displayError(ErrorMessages.FAILED_LOAD);
+      });
+  }, []);
```
roman-mirzoian, line 48: Don't leave comments in the code that repeat what can be inferred from the code itself

roman-mirzoian (approved):
Good job!
Don't forget to include a link to the demo of your result and take note of the comment below: you shouldn't leave comments in the code that describe exactly what the code does, for example,  `// > Add Todo` when the handler below is named `handleAddTodo`. Comments are needed when the code is difficult to understand.

https://kseniiavasylenko.github.io/react_todo-app-loading-todos/

PR #2449, opened 2026-07-28
added 1-st part

Skoulskiy:
[DEMO LINK](https://skoulskiy.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+        const fetchedTodos = await getTodos();
+
+        setTodos(fetchedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to load todos');
+      }
+    };
+
+    fetchTodosData();
+  }, []);
```
Zoolgrand: Please create custom hook like useTodo and extract all states, effects, handlers and other non related to renders logic in this hook

src/App.tsx
```diff
@@ -33,171 +89,115 @@ export const App: React.FC = () => {
...
+                  data-cy="TodoDelete"
+                >
+                  ×
+                </button>
+
+                <div data-cy="TodoLoader" className="modal overlay">
+                  <div className="modal-background has-background-white-ter" />
+                  <div className="loader" />
+                </div>
+              </div>
```
Zoolgrand: Could you please extract this into separate components

src/App.tsx
```diff
@@ -33,171 +89,115 @@ export const App: React.FC = () => {
...
-            type="button"
-            className="todoapp__clear-completed"
-            data-cy="ClearCompletedButton"
-          >
-            Clear completed
-          </button>
-        </footer>
+              Clear completed
+            </button>
+          </footer>
```
Zoolgrand: We can also extract this footer into separate component

src/App.tsx
```diff
@@ -1,10 +1,64 @@
...
+    setErrorMessage('');
+
+    const fetchTodosData = async () => {
+      try {
+        const fetchedTodos = await getTodos();
+
+        setTodos(fetchedTodos);
+      } catch (error) {
+        setErrorMessage('Unable to load todos');
+      }
```
Zoolgrand: error variable is unsed here, you can re,ove it or log with console.error

Zoolgrand (changes requested):
The first part of the task is almost finished; however, App.tsx requires refactoring. Also, there are no real todos for user 4401 right now, so we always have an empty list, but this will be required in the next step of the task.

roman-mirzoian (approved):
Well done 👍

Zoolgrand (approved):
Well done

PR #2452, opened 2026-07-31
first update

student1232354:
[DEMO LINK](https://student1232354.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,14 +1,127 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+
+  const [deleteError, setdeleteError] = useState(false);
+  const [Error, setError] = useState(false);
+  const [Rror, setRror] = useState(false);
+  const [hidden, sethidden] = useState(false);
+  const [notTitle, setNotTitle] = useState(false);
+  const [Update, setUpdate] = useState(false);
+  const [firstFilter, setFirstFilter] = useState('All');
```
etojeDenys: create an enum for the filter parameters

src/App.tsx
```diff
@@ -1,14 +1,127 @@
...
+import { Todolist } from './forArray/Todolist';
+import { Todofilter } from './filter/Todofilter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [title, setTitle] = useState('');
+
+  const [deleteError, setdeleteError] = useState(false);
+  const [Error, setError] = useState(false);
+  const [Rror, setRror] = useState(false);
```
etojeDenys: instead of having several states for different errors, you can use only one state and write the relevant error message in it

src/filter/Todofilter.tsx
```diff
@@ -0,0 +1,74 @@
...
+          className={
+            firstFilter === 'Completed'
+              ? 'filter__link selected'
+              : 'filter__link'
+          }
+          data-cy="FilterLinkCompleted"
+          onClick={() => setFirstFilter('Completed')}
+        >
+          Completed
+        </a>
```
etojeDenys: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job,  but demo link does not work. please fix it and re-request the review

etojeDenys (changes requested):
demo link is working now, but you should fix the problem mentioned in the comments

vitalii-bondarenko-ua (approved):
Looks good

PR #2461, opened 2026-08-05
result

Zakys1d:
[DEMO LINK](https://Zakys1d.github.io/react_todo-app-loading-todos/)


src/App.tsx
```diff
@@ -1,28 +1,78 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+
+import React, { useEffect, useMemo, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { USER_ID, getTodos } from './api/todos';
+import { Todo } from './types/Todo';
+
+type Filter = 'all' | 'active' | 'completed';
```
etojeDenys, line 10: it would be better to create an enum than a type

src/App.tsx
```diff
@@ -1,28 +1,78 @@
...
+    if (!USER_ID) {
+      return;
+    }
+
+    setError('');
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        setError('Unable to load todos');
```
etojeDenys, line 27: create an enum for errors as well

src/App.tsx
```diff
@@ -33,171 +83,110 @@ export const App: React.FC = () => {
...
+                href="#/completed"
+                className={`filter__link ${
+                  filter === 'completed' ? 'selected' : ''
+                }`}
+                data-cy="FilterLinkCompleted"
+                onClick={() => setFilter('completed')}
+              >
+                Completed
+              </a>
+            </nav>
```
etojeDenys, line 162: you can create an array with the required parameters and then map through it to create links and avoid code repetition

etojeDenys (changes requested):
good job, to improve:

2. pay attention to the comments

roman-mirzoian (changes requested):
Please take a look at the previous review, as it contains comments that need to be fixed:
https://github.com/mate-academy/react_todo-app-loading-todos/pull/2461#pullrequestreview-4867338443

PR #2463, opened 2026-08-08
ready

MarianaKlochnyk:
https://MarianaKlochnyk.github.io/react_todo-app-loading-todos/


src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
+  };
+
+  const loadTodos = () => {
+    setError('');
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        showError('Unable to load todos');
```

src/App.tsx
```diff
@@ -1,10 +1,58 @@
...
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { UserWarning } from './UserWarning';
-import { USER_ID } from './api/todos';
+import { getTodos, USER_ID } from './api/todos';
+import { Todo } from './types/Todo';
+import classNames from 'classnames';
+
+type Filter = 'all' | 'active' | 'completed';
```
etojeDenys: it would be better to create an enum than a type

src/App.tsx
```diff
@@ -15,189 +63,141 @@ export const App: React.FC = () => {
...
+                <a
+                  href="#/completed"
+                  className={classNames('filter__link', {
+                    selected: filter === 'completed',
+                  })}
+                  data-cy="FilterLinkCompleted"
+                  onClick={() => setFilter('completed')}
+                >
+                  Completed
+                </a>
```

etojeDenys (changes requested):
almost done, to improve:

2. demo link does not work. please fix it

Anton-Kuchmasov (approved):
Well done!

PR #2465, opened 2026-08-12
Develop

johnnyvasconcelos:


DanielCamposDevX (changes requested):
React Hook useEffect has a missing dependency: 'todos'. Either include it or remove the dependency array on line 45 of app.tsx. No demo link was provided, remember to read the task guide

